PR #3, opened 2019-12-18
Add C build tools repo as dependency

dcristoloveanu:
Add C build tools repo as dependency & enable requirements traceability leg

### DIFF
--- a/build/devops_gated.yml
+++ b/build/devops_gated.yml
@@ -253,7 +253,6 @@ jobs:
     - Cmd
     - msbuild
     - visualstudio
-  condition: false
 
   steps:
   - task: BatchScript@1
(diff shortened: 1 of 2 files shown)


build/devops_gated.yml
```diff
@@ -253,7 +253,6 @@ jobs:
     - Cmd
     - msbuild
     - visualstudio
-  condition: false
```
mattdurak: >condition: false [](start = 2, length = 16)

Is there a reason to enable only this job?

dcristoloveanu: I would enable all, but we don't have a CMakeLists yet. I enabled this one to validate that the c build tools building of the traceability tool works.

PR #13, opened 2020-02-10
have a global timer

anporumb:


tests/reals/real_timer_renames.h
```diff
@@ -1,8 +1,9 @@
...
-#define timer_start           real_timer_start
-#define timer_get_elapsed     real_timer_get_elapsed
-#define timer_get_elapsed_ms  real_timer_get_elapsed_ms
-#define timer_destroy         real_timer_destroy
+#define timer_create            real_timer_create
+#define timer_start             real_timer_start
+#define timer_get_elapsed       real_timer_get_elapsed
+#define timer_get_elapsed_ms    real_timer_get_elapsed_ms
+#define timer_destroy           real_timer_destroy
+#define g_timer_get_elapsed_ms  real_g_timer_get_elapsed_ms
```
dcristoloveanu: >g_timer_get_elapsed_ms [](start = 8, length = 22)

Small ask: timer_global_get_elapsed_ms pretty please. Simply because g_ in front makes my brain immediately think of a global *variable*.

tests/timer_win32_ut/timer_mocked.c
```diff
@@ -1,6 +1,6 @@
 // Copyright (c) Microsoft. All rights reserved.
 
-#include "Windows.h"
+#include "windows.h"
```
dcristoloveanu: >"wi [](start = 9, length = 3)

Cheers

adapters/timer.c
```diff
@@ -77,3 +77,21 @@ void timer_destroy(TIMER_HANDLE timer)
         free(timer);
     }
 }
+
+static LARGE_INTEGER g_freq;
+static volatile LONG g_timer_state = 0; /*0 - not "created", 1 - "created", "2" - creating*/
```
dcristoloveanu: >t "created", 1 - "created", "2" - creating*/ [](start = 48, length = 44)

I'd do an enum for these states, easier to figure in the code what the states are rather than looking at numbers :-).

tests/reals/real_timer_renames.h
anporumb: renamed

adapters/timer.c
anporumb: I initially had an enum, then I decided I liked the approach with "0,1,2" better. They are all confined to 3 lines of code.

PR #16, opened 2020-02-27
THANDLEs

anporumb:


devdoc/thandle_requirements.md
```diff
@@ -0,0 +1,135 @@
+# thandle_requirements
+================
+
+## Overview
+
+`thandle` is a module that provides a list of macros that help with ref counted handles.
+
+Given a type `LL` previously introduced by `MU_DEFINE_STRUCT`, `thandle`'s macros will encapsulate the type, provide a HANDLE (that is, opaque type) to it which other APIs can use. The handle type is `THANDLE(LL)`.
```
dcristoloveanu: >LL [](start = 14, length = 2)

T

dcristoloveanu: >LL [](start = 209, length = 2)

I guess T also

devdoc/thandle_requirements.md
```diff
@@ -0,0 +1,135 @@
...
+```c
+MOCKABLE_FUNCTION(, void, THANDLE_ASSIGN(T), THANDLE(T) *, t1, THANDLE(T), t2 );
+```
+
+`THANDLE_ASSIGN` does t1=t2. Both `t1` and `t2` are existing constructed handles.
+
+
+**SRS_THANDLE_02_006: [** If `t1` is `NULL` then `THANDLE_ASSIGN` shall return. **]**
+
+**SRS_THANDLE_02_007: [** If `*t1` is `NULL` and `t2` is `NULL` then `THANDLE_ASSIGN` shall return. **]**
```
dcristoloveanu: >*t1` [](start = 30, length = 4)

So t1 is sort of an in/out....

devdoc/thandle_requirements.md
```diff
@@ -0,0 +1,135 @@
...
+
+**SRS_THANDLE_02_008: [** If `*t1` is `NULL` and `t2` is not `NULL` then `THANDLE_ASSIGN` shall increment the reference count of `t2` and store `t2` in `*t1`. **]**
+
+**SRS_THANDLE_02_009: [** If `*t1` is not `NULL` and `t2` is `NULL` then `THANDLE_ASSIGN` shall decrement the reference count of `*t1` and store `NULL` in `*t1`. **]**
```
dcristoloveanu: >If `*t1` is not `NULL` and `t2` is `NULL` then `THANDLE_ASSIGN` shall decrement the reference count of `*t1` and store  [](start = 26, length = 119)

I find this very interesting. And nice. And at the same time I am a bit scared of it as these assignment will hide another layer of inc ref/dec ref. Let's see how it goes.

devdoc/thandle_requirements.md
```diff
@@ -0,0 +1,135 @@
...
+
+**SRS_THANDLE_02_010: [** If `*t1` is not `NULL` and `t2` is not `NULL` then `THANDLE_ASSIGN` shall increment the reference count of `t2`, shall decrement the reference count of `*t1` and store `t2` in `*t1`. **]**
+
+###  THANDLE_INITIALIZE(T)
+```c
+ MOCKABLE_FUNCTION(, void, THANDLE_INITIALIZE(T), THANDLE(T) *, t1, THANDLE(T), t2 );
+ ```
+
+`THANDLE_ASSIGN` does t1=t2. `t1` is NOT a constructed handle. Example: it is a field in a struct that has just been `malloc`'d.
```
dcristoloveanu: >DLE_ASSIGN` does t1=t2. [](start = 5, length = 23)

copy pasta detected. This should be THANDLE_INITIALIZE

devdoc/thandle_requirements.md
```diff
@@ -0,0 +1,135 @@
...
+
+**SRS_THANDLE_02_008: [** If `*t1` is `NULL` and `t2` is not `NULL` then `THANDLE_ASSIGN` shall increment the reference count of `t2` and store `t2` in `*t1`. **]**
+
+**SRS_THANDLE_02_009: [** If `*t1` is not `NULL` and `t2` is `NULL` then `THANDLE_ASSIGN` shall decrement the reference count of `*t1` and store `NULL` in `*t1`. **]**
+
+**SRS_THANDLE_02_010: [** If `*t1` is not `NULL` and `t2` is not `NULL` then `THANDLE_ASSIGN` shall increment the reference count of `t2`, shall decrement the reference count of `*t1` and store `t2` in `*t1`. **]**
+
+###  THANDLE_INITIALIZE(T)
+```c
+ MOCKABLE_FUNCTION(, void, THANDLE_INITIALIZE(T), THANDLE(T) *, t1, THANDLE(T), t2 );
```
dcristoloveanu: >t1 [](start = 64, length = 2)

This might deserve a bit better naming since there is clearly a comment in ln. 97 warning that this is not the same as the lvalue in THANDLE_ASSIGN.

Might I suggest naming the THANDLE_ASSIGN args lvalue and rvalue.
And a suggestion for t1 here would be wrapped_type, while t2 could be out_handle.

You consider it...

devdoc/thandle_requirements.md
```diff
@@ -0,0 +1,135 @@
...
+**SRS_THANDLE_02_012: [** `THANDLE_INITIALIZE` shall increment the reference count of `t2` and store it in `*t1`. **]**
+
+## THANDLE_TYPE_DEFINE(T)
+```
+#define THANDLE_TYPE_DEFINE(T)
+```
+
+`THANDLE_TYPE_DEFINE` introduces the implementation for the functions in `THANDLE_TYPE_DECLARE` (`THANDLE_DEC_REF`, `THANDLE_INC_REF`, `THANDLE_ASSIGN`, `THANDLE_INITIALIZE`) and two new memory management functions `THANDLE_MALLOC(T)` and `THANDLE_FREE(T)`.
+
+### THANDLE_MALLOC(T)
```
dcristoloveanu: Some of our structs have flexible arrays and ref count has grown a malloc with extra size.
I think it would be very good to have this in here also.

devdoc/thandle_requirements.md
```diff
@@ -0,0 +1,135 @@
...
+```c
+static T* THANDLE_MALLOC(T)(void(*dispose)(T*))
+```
+
+`THANDLE_MALLOC` return a pointer to `T`. `dispose` is a function that the `THANDLE_DEC_REF` calls when the reference count reaches 0 in order to free the resources allocated by the user in `T`. `dispose` can be `NULL` in which case there are no user resources to be de-allocated.
+
+**SRS_THANDLE_02_013: [** `THANDLE_MALLOC` shall allocate memory. **]**
```
dcristoloveanu: >llocate memory [](start = 50, length = 14)

all of it :-)? Consider speccing sizeof(T).

dcristoloveanu: Ah I see, it probably needs sizeof(T) + the size of the dispose pointer.

---
In reply to: [385370942](https://github.com/Azure/azure-c-util/pull/16#discussion_r385370942) [](ancestors = 385370942)

devdoc/thandle_requirements.md
```diff
@@ -0,0 +1,135 @@
...
+
+**SRS_THANDLE_02_015: [** If `malloc` fails then `THANDLE_MALLOC` shall fail and return `NULL`. **]**
+
+
+### `THANDLE_FREE(T)`
+```c
+    THANDLE_FREE(T)(T* t)
+```
+
+`THANDLE_FREE` free the allocated memory by `THANDLE_MALLOC`. It is called from `THANDLE_DEC_REF` (when reference count reaches 0) after `dispose`, or it can be called from the user code to free the memory.
```
dcristoloveanu: >free [](start = 15, length = 4)

frees

inc/azure_c_util/thandle.h
```diff
@@ -0,0 +1,189 @@
...
+    MU_DEFINE_STRUCT(THANDLE_WRAPPER_TYPE_NAME(T), T, data, THANDLE_EXTRA_FIELDS(T));                                                                                   \
+    static T* THANDLE_MALLOC(T)(void(*dispose)(T*))                                                                                                                     \
+    {                                                                                                                                                                   \
+        T* result;                                                                                                                                                      \
+        /*Codes_SRS_THANDLE_02_013: [ THANDLE_MALLOC shall allocate memory. ]*/                                                                                         \
+        THANDLE_WRAPPER_TYPE_NAME(T)* handle_impl = (THANDLE_WRAPPER_TYPE_NAME(T)*)malloc(sizeof(THANDLE_WRAPPER_TYPE_NAME(T)));                                        \
+        if (handle_impl == NULL)                                                                                                                                        \
```
dcristoloveanu: These macro that we write start to look like templates :-)

inc/azure_c_util/thandle.h
```diff
@@ -0,0 +1,189 @@
...
+            {                                                                                                                                                           \
+                if (t2 == NULL)                                                                                                                                         \
+                {                                                                                                                                                       \
+                    /*Codes_SRS_THANDLE_02_009: [ If *t1 is not NULL and t2 is NULL then THANDLE_ASSIGN shall decrement the reference count of *t1 and store NULL in *t1. ]*/ \
+                    THANDLE_DEC_REF(T)(*t1);                                                                                                                            \
+                    * (T const**)t1 = t2;                                                                                                                               \
```
dcristoloveanu: >(T const**) [](start = 22, length = 11)

I was looking for this cast :-)

tests/thandle_ut/thandle_ut.c
```diff
@@ -0,0 +1,407 @@
...
+    umock_c_reset_all_calls();
+}
+
+TEST_FUNCTION_CLEANUP(cleans)
+{
+    TEST_MUTEX_RELEASE(g_testByTest);
+}
+
+/*Tests_SRS_THANDLE_02_013: [ THANDLE_MALLOC shall allocate memory. ]*/
+/*Tests_SRS_THANDLE_02_014: [ THANDLE_MALLOC shall initialize the reference count to 1, store dispose and return a T* . ]*/
```
dcristoloveanu: Could we have that region-like comment with /* THANDLE_MALLOC */. I just found that useful a couple of times in my latest changes and it's not a lot to add them to ask for :-).

inc/azure_c_util/containing_record.h
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#ifndef CONTAINING_RECORD_H
+#define CONTAINING_RECORD_H
+
```
dcristoloveanu: Nice!

dcristoloveanu:
The only more interesting ask here it to have the malloc with extra size to accommodate flexible arrays in the underlying type.
If you can make it as part of the same PR that would be awesome, so we're then done with this.

devdoc/thandle_requirements.md
anporumb: changed

anporumb: old remains are old. changed.

---
In reply to: [385363919](https://github.com/Azure/azure-c-util/pull/16#discussion_r385363919) [](ancestors = 385363919)

anporumb: Yeah, and done with it

---
In reply to: [385364246](https://github.com/Azure/azure-c-util/pull/16#discussion_r385364246) [](ancestors = 385364246)

anporumb:
I had the exact same thought.

Mind if I do it with another checkin? Should be easy enough and I want to separate them.

---
In reply to: [592180478](https://github.com/Azure/azure-c-util/pull/16#issuecomment-592180478) [](ancestors = 592180478)

anporumb:
added a task to have HANDLE creation with trailing space. [https://msazure.visualstudio.com/One/_workitems/edit/6360591](https://msazure.visualstudio.com/One/_workitems/edit/6360591)

---
In reply to: [592187882](https://github.com/Azure/azure-c-util/pull/16#issuecomment-592187882) [](ancestors = 592187882,592180478)

devdoc/thandle_requirements.md
anporumb: As the type implies, T* (note the missing const)

---
In reply to: [385366520](https://github.com/Azure/azure-c-util/pull/16#discussion_r385366520) [](ancestors = 385366520)

devdoc/thandle_requirements.md
anporumb: Frankly, initially I did not even have inc_REF, because "you never want to inc-ref just because". But then I thought "so what happens when the hadle is passed as context?"

So I needed INC_REF. :(

And yes, I agree, let's see how it goes.

---
In reply to: [385367955](https://github.com/Azure/azure-c-util/pull/16#discussion_r385367955) [](ancestors = 385367955)

devdoc/thandle_requirements.md
anporumb: true that. fixed.,

---
In reply to: [385368448](https://github.com/Azure/azure-c-util/pull/16#discussion_r385368448) [](ancestors = 385368448)

devdoc/thandle_requirements.md
anporumb: I heard you in a previous comment and I will do it - when this goes in. Should be pretty quick now :)

---
In reply to: [385370511](https://github.com/Azure/azure-c-util/pull/16#discussion_r385370511) [](ancestors = 385370511)

devdoc/thandle_requirements.md
anporumb: myeah, I could add "at least sizeof(T)" but that doesn't help much :(. I propose to leave it as is.

---
In reply to: [385371334](https://github.com/Azure/azure-c-util/pull/16#discussion_r385371334) [](ancestors = 385371334,385370942)

devdoc/thandle_requirements.md
anporumb: fixed.

---
In reply to: [385371446](https://github.com/Azure/azure-c-util/pull/16#discussion_r385371446) [](ancestors = 385371446)

inc/azure_c_util/thandle.h
anporumb: Well - we HAVE to cheat a bit. But "keep it low". Cannot be done other way - or let the handles be assignable - which is somewhat ofd a corner stone of the PR.

---
In reply to: [385373842](https://github.com/Azure/azure-c-util/pull/16#discussion_r385373842) [](ancestors = 385373842)

inc/azure_c_util/thandle.h
anporumb: "If the compiler doesn't do it!!! We have to do it ourselves!!!"

---
In reply to: [385373033](https://github.com/Azure/azure-c-util/pull/16#discussion_r385373033) [](ancestors = 385373033)

devdoc/thandle_requirements.md
anporumb: renamed these to be lvalue and rvalue.

---
In reply to: [385369765](https://github.com/Azure/azure-c-util/pull/16#discussion_r385369765) [](ancestors = 385369765)

inc/azure_c_util/containing_record.h
anporumb: Thanks!

---
In reply to: [385375187](https://github.com/Azure/azure-c-util/pull/16#discussion_r385375187) [](ancestors = 385375187)

tests/thandle_ut/thandle_ut.c
anporumb: I do not like "regions" because I believe they are redundant with the rest of the text in the file.

1. redundant with the TEST_FUNCTIOIN(API_NAME_does_something)
2. redundant with the /*TESTS_SRS_BALBALA_... where it explains what it does

I will add them anyway.


---
In reply to: [385375049](https://github.com/Azure/azure-c-util/pull/16#discussion_r385375049) [](ancestors = 385375049)

PR #20, opened 2020-03-13
thandle needs an accessor to t

anporumb:


inc/azure_c_util/thandle.h
```diff
@@ -12,7 +12,10 @@
 #include "refcount_os.h"
 
 /*the incomplete unassignable type*/
-#define THANDLE(T) const T* const
+#define THANDLE(T) MU_C2(CONST_P2_CONST_,T)
```
dcristoloveanu: >T) MU_C2(CONST_P2_CONST_,T) [](start = 16, length = 27)

Ah, this should fix the issue I mentioned with passing the array of handles as out arg.

dcristoloveanu: Nice.

---
In reply to: [392014336](https://github.com/Azure/azure-c-util/pull/20#discussion_r392014336) [](ancestors = 392014336)

tests/thandle_ut/thandle_ut.c
```diff
@@ -488,4 +488,41 @@ TEST_FUNCTION(THANDLE_FREE_with_t_NULL_returns)
     ///cleanup
 }
```
dcristoloveanu: Can you actually add a test for:

THANDLE(T)** x;
*x = malloc(sizeof(THANDLE(T)) * 42);

I suspect it should compile now.

anporumb: Wrote a test

---
In reply to: [392014690](https://github.com/Azure/azure-c-util/pull/20#discussion_r392014690) [](ancestors = 392014690)

PR #25, opened 2020-04-06
THANDLE needs to have a malloc/free function of its own.

anporumb:


inc/azure_c_util/thandle.h
```diff
@@ -8,9 +8,25 @@
...
+        /*do nothing, the macros here will call whatever THANDLE_MALLOC_FUNCTION/THANDLE_FREE_FUNCTION expands to*/
+    #endif
+#else
+    #ifdef THANDLE_FREE_FUNCTION
+        #error THANDLE_MALLOC_FUNCTION and THANDLE_FREE_FUNCTION must be both defined or both not defined
+    #else
+        /*include the "stock" implementation*/
+        #include "azure_c_util/thandle_stdlib.h"
+    #endif
+#endif
```
dcristoloveanu, line 28: I think this can be written with if defined to be less verbose.

anporumb: So there are 3 cases possible out of this construct:
1) do nothing
2) spew an error
3) include thandle_stdlib.

1) can go away and not be mentioned explicitly.
2) and 3) need to stay.

Code looks... more convoluted, but less lines of code.

 This is it:
```c
#if (defined THANDLE_MALLOC_FUNCTION) ^ (defined THANDLE_FREE_FUNCTION)
#error THANDLE_MALLOC_FUNCTION and THANDLE_FREE_FUNCTION must be both defined or both not defined
#else
#if !defined THANDLE_MALLOC_FUNCTION
    #include "azure_c_util/thandle_stdlib.h"
#endif
#endif
```

I don't know how to write it any nicer. I would like to keep it verbose so it is easier to understand. The way this is written only matters to humans. It will expand to the same thing one way or the other.

dcristoloveanu: I guess another option would be:

#if defined M and defined F
// use them
#elif defined M or defined F
#error
#else
use defaults
#endif

But this is really a minor thing, so choose one and go. :-)

/Rudotel

anporumb: Chose mine, went: )

PR #27, opened 2020-04-18
add THANDLE_COPY

anporumb:


devdoc/thandle_requirements.md
```diff
@@ -163,4 +163,24 @@ Given a previously constructed `THANDLE(T)`, `THANDLE_GET_T(T)` reeturns a point
 
 **SRS_THANDLE_02_023: [** If `t` is `NULL` then `THANDLE_GET_T(T)` shall return `NULL`. **]**
 
-**SRS_THANDLE_02_024: [** `THANDLE_GET_T(T)` shall return the same pointer as `THANDLE_MALLOC`/`THANDLE_MALLOC_WITH_EXTRA_SIZE` returned at the handle creation time. **]**
\ No newline at end of file
+**SRS_THANDLE_02_024: [** `THANDLE_GET_T(T)` shall return the same pointer as `THANDLE_MALLOC`/`THANDLE_MALLOC_WITH_EXTRA_SIZE` returned at the handle creation time. **]**
+
+###     THANDLE_COPY_MACRO(T)
```
dcristoloveanu, line 168: >THANDLE_COPY [](start = 8, length = 12)

poetry: I actually find copy a bit misleading, because copy would make me think of copying the contents of an existing handle.
I would name this something like: create_from_content...
/poetry.

dcristoloveanu: Renames are cheap.

---
In reply to: [410621415](https://github.com/Azure/azure-c-util/pull/27#discussion_r410621415) [](ancestors = 410621415)

tests/thandle_ut/thandle_ut.c
```diff
@@ -71,6 +72,50 @@ extern "C" {
...
+
+    return result;
+}
+
+static void dispose_A_S(A_S* a_s)
+{
+    free(a_s->s);
+}
+
+#define THANDLE_MALLOC_FUNCTION gballoc_malloc
```
dcristoloveanu, line 106: An interesting thing about #define and #undef: Recently I tried enabling precompiled headers. And ... #define/#undef - like the ones in umock or here are actually not that nice.
I don't have a better solution now, but maybe we can both think of something better... for the next PR.

devdoc/thandle_requirements.md
anporumb: poetrizied.

tests/thandle_ut/thandle_ut.c
anporumb: yup, cannot have precompiled headers that change shape/form :(

PR #30, opened 2020-04-29
Disable w4200 globally

dcristoloveanu:
Disable compiler warning 4200 for MSC


CMakeLists.txt
```diff
@@ -55,6 +55,10 @@ if(${use_etw} STREQUAL "TRACELOGGING")
     add_definitions(-DUSE_TRACELOGGING)
 endif() #(${use_etw} STREQUAL "TRACELOGGING")
 
+# warning C4200: nonstandard extension used: zero-sized array in struct/union : looks very standard in C99 and it is called flexible array. Documentation-wise is a flexible array, but called "unsized" in Microsoft's docs
+# https://msdn.microsoft.com/en-us/library/b6fae073.aspx
+add_definitions(/wd4200)
```
anporumb, line 60: >add_definitions [](start = 0, length = 15)

You should only add this definition to the C/CXX compiler (only). For C# it might have different semantics / maybe none... 

Splitting C/CXX from CSharp is another option too.

anporumb, line 60: Is there any way to add this to an even lower level, such as... build-tools?

PR #31, opened 2020-04-29
Add InterlockedHL_WaitForNotValue with code and tests

mattdurak:
Adds InterlockedHL_WaitForNotValue, which is InterlockedHL_WaitForValue except it waits for the value to equal anything else.


devdoc/interlocked_hl_requirements.md
```diff
@@ -70,6 +71,27 @@ MOCKABLE_FUNCTION_WITH_RETURNS(, INTERLOCKED_HL_RESULT, InterlockedHL_WaitForVal
...
+
+`InterlockedHL_WaitForNotValue` waits for the value at a given address to not be equal to a target value.
+
+**SRS_INTERLOCKED_HL_42_001: [** If `address` is `NULL`, `InterlockedHL_WaitForNotValue` shall fail and return `INTERLOCKED_HL_ERROR`. **]**
+
+**SRS_INTERLOCKED_HL_42_002: [** If the value at `address` is not equal to `value`, `InterlockedHL_WaitForNotValue` shall return `INTERLOCKED_HL_OK`. **]**
+
+**SRS_INTERLOCKED_HL_42_003: [** If the value at `address` is equal to `value`, `InterlockedHL_WaitForNotValue` shall wait until the value at `address` changes in order to compare it again to `value` by using `WaitOnAddress`. **]**
+
+**SRS_INTERLOCKED_HL_42_004: [** When waiting for the value at address to change, the `milliseconds` argument value shall be used as timeout. **]**
```
dcristoloveanu, line 88: >When waiting for the value at address to change, the `milliseconds` argument value shall be used as timeout. [](start = 33, length = 108)

This can actually wait for longer than milliseconds. I see that that was done in the original - apparently written by me based on the 01 Id.
Maybe we should probably have another task somewhere to fix this to be accurate. Seems like we would need timer calls for it.

mattdurak: Created a task to fix that, good call

---
In reply to: [417474866](https://github.com/Azure/azure-c-util/pull/31#discussion_r417474866) [](ancestors = 417474866)

PR #35, opened 2020-05-07
have state manager

anporumb:


adapters/interlocked_hl_win32.c
```diff
@@ -116,6 +116,47 @@ IMPLEMENT_MOCKABLE_FUNCTION(, INTERLOCKED_HL_RESULT, InterlockedHL_WaitForValue,
...
 }
 
+IMPLEMENT_MOCKABLE_FUNCTION(, INTERLOCKED_HL_RESULT, InterlockedHL_WaitForValue64, LONG64 volatile*, address, LONG64, value, DWORD, milliseconds)
+{
+    INTERLOCKED_HL_RESULT result;
+
+    /*Codes_SRS_INTERLOCKED_HL_02_021: [ If address is NULL then InterlockedHL_WaitForValue64 shall fail and return INTERLOCKED_HL_ERROR. ]*/
+    if (address == NULL)
+    {
+        result = INTERLOCKED_HL_ERROR;
```
dcristoloveanu, line 126: >result = INTERLOCKED_HL_ERROR; [](start = 8, length = 30)

Let's add a LogError.

adapters/interlocked_hl_win32.c
```diff
@@ -116,6 +116,47 @@ IMPLEMENT_MOCKABLE_FUNCTION(, INTERLOCKED_HL_RESULT, InterlockedHL_WaitForValue,
...
+            }
+
+            /* Codes_SRS_INTERLOCKED_HL_02_026: [ If the value at address does not match, InterlockedHL_WaitForValue64 shall issue another call to WaitOnAddress. ]*/
+
+            /* Codes_SRS_INTERLOCKED_HL_02_023: [ If the value at address is not equal to value, InterlockedHL_WaitForValue64 shall wait until the value at address changes in order to compare it again to value by using WaitOnAddress. ]*/
+            /* Codes_SRS_INTERLOCKED_HL_02_024: [ When waiting for the value at address to change, the milliseconds argument value shall be used as timeout. ]*/
+            if (!WaitOnAddress(address, &current_value, sizeof(current_value), milliseconds))
+            {
+                /* Codes_SRS_INTERLOCKED_HL_02_027: [ If WaitOnAddress fails, InterlockedHL_WaitForValue64 shall fail and return INTERLOCKED_HL_ERROR. ]*/
+                result = INTERLOCKED_HL_ERROR;
```
dcristoloveanu, line 150: >result = INTERLOCKED_HL_ERROR; [](start = 16, length = 30)

LogError?

CMakeLists.txt
```diff
@@ -97,6 +97,8 @@ function(set_platform_files c_util_dir)
         set(UNIQUEID_C_FILE ${c_util_dir}/adapters/uniqueid_win32.c PARENT_SCOPE)
         set(STRING_UTILS_C_FILE ${c_util_dir}/adapters/string_utils.c PARENT_SCOPE)
         set(SRW_LOCK_C_FILE ${c_util_dir}/adapters/srw_lock.c PARENT_SCOPE)
+        set(SM_C_FILE ${c_util_dir}/src/sm.c PARENT_SCOPE)
```
dcristoloveanu, line 100: >sm [](start = 40, length = 2)

I know it's poetry, but let's name this state_manager. I like the manager part in it :-). It will be easier for someone else (Parth, Cesar, future generations) looking at our code to figure out what this unit does.

CMakeLists.txt
```diff
@@ -106,6 +108,8 @@ function(set_platform_files c_util_dir)
         set(XLOGGING_C_FILE ${c_util_dir}/src/xlogging.c PARENT_SCOPE)
         set(LOGGING_C_FILE ${c_util_dir}/src/consolelogger.c PARENT_SCOPE)
         set(LOGGING_H_FILE ${c_util_dir}/inc/azure_c_util/consolelogger.h PARENT_SCOPE)
+        set(SM_C_FILE) #there's no SM for linux
```
dcristoloveanu, line 111: >#there's no SM for linux [](start = 23, length = 24)

There will be soon, as soon as Parth puts in that interlocked_if. :-)

devdoc/sm_requirements.md
```diff
@@ -0,0 +1,228 @@
+# State manager requirements
```
dcristoloveanu, line 1: I have waited a long time to see this done, so I are very excited. My starting this effort is from 1y ago!

devdoc/sm_requirements.md
```diff
@@ -0,0 +1,228 @@
...
+
+5. there's no "parking", "queueing", "sleeping", "waiting", "buffering", "saving" or otherwise postponing or delaying the execution of an API: it either starts to execute (for barries execution includes waiting for all the previous APIs to finish executing) or it fails. Retrying is a user-land mechanism.
+
+## Design
+
+Historic context: before `sm` the world had explicit states, such as 
+- "OPEN" (where all APIs would be allowed to execute)
+- "OPENING" (marks the transiton from "CREATED" to "OPEN")
+- "CLOSING" (where APIs would be drained - it devolves into "CREATED" once that was done)
+- all sort of other substates of "OPEN" which would not allow other APIs to execute (and once work was done, the substte would revert back to "OPEN").
```
dcristoloveanu, line 34: >substte [](start = 113, length = 7)

substate

devdoc/sm_requirements.md
```diff
@@ -0,0 +1,228 @@
...
+- providing progress in all cases for APIs that require their own sub-state undisturbed (that is, APIs that cannot execute in parallel with any other APIs), so no more a problem of "enough luck".
+
+To achieve the above goals, `sm` assigns to each API call a sequence number (`n`). Sequence numbers are consecutive. The logic is simple: an API is granted execution when there's no barrier (it acts as invariant for the increment of `n`). Otherwise, the API is prohibited from execution.
+
+Barriers follow their own ever increasing numbers called `b_now`. `b_now` start at 0. A barrier is taken when `b_now`'s least significant bit is 1. Example `b_now` == 0 means there's no ongoing barrier. `b_now` == 3 means there's an ongoing barrier.
+
+Multiple barrier can attempt to set `b_now`'s least significant bit to 1. Only 1 of them will win the competition onver `InterlockedOr64` so only 1 will grant execution. The other barries will lose arbitration and will not be granted execution rights. When user calls `sm_barrier_end` to indicate that the barrier should be lifted, `b_now` is `InterlockedIncremented64`.
```
dcristoloveanu, line 50: >onver [](start = 114, length = 5)

over

devdoc/sm_requirements.md
```diff
@@ -0,0 +1,228 @@
...
+```c
+MOCKABLE_FUNCTION(, SM_HANDLE, sm_create, const char*, name);
+```
+`sm_create` creates a new State manager handle. `name` is used with logging and bears no functional significance.
+
+**SRS_SM_02_001: [** If `name` is `NULL` then `sm_create` shall behave as if `name` was "NO_NAME". **]**
+
+**SRS_SM_02_002: [** `sm_create` shall allocate memory for the instance. **]**
+
+**SRS_SM_02_003: [** `sm_create` shall set `b_now` to -1, `n` to 0, and `e` to 0 succeed and return a non-`NULL` value. **]**
```
dcristoloveanu, line 98: > 0 suc [](start = 78, length = 6)

0, succeed

src/sm.c
```diff
@@ -0,0 +1,331 @@
...
+#include "azure_c_util/gballoc.h"
+#include "azure_c_util/interlocked_hl.h"
+
+#include "azure_c_util/sm.h"
+
+typedef struct SM_HANDLE_DATA_TAG
+{
+    volatile LONG64 b_now; /*where's the barrier at, starts at 0. 1 is when barrier is active. 2,4,6... barrier inactive. 3,5,7... barrier active*/
+    volatile LONG64 n; /*number of API calls to barriers and non-barriers alike*/
+    volatile LONG64 e; /*how many of the APIs have executed. Both abrriers and non-barriers set this*/
```
dcristoloveanu, line 20: >abrriers [](start = 66, length = 8)

barriers

src/sm.c
```diff
@@ -0,0 +1,331 @@
...
+{
+    if (sm == NULL)
+    {
+        /*Codes_SRS_SM_02_010: [ If sm is NULL then sm_open_end shall return. ]*/
+        LogError("invalid argument SM_HANDLE sm=%p", sm);
+    }
+    else
+    {
+        /*Codes_SRS_SM_02_011: [ sm_open_end shall increment e, b_now to 0 and return. ]*/
+        InterlockedIncrement64(&sm->e);
```
dcristoloveanu, line 123: >InterlockedIncrement64 [](start = 8, length = 22)

(void)

src/sm.c
```diff
@@ -0,0 +1,331 @@
...
+        InterlockedExchange64(&sm->b_now, 0);
```
dcristoloveanu, line 124: >InterlockedExchange64 [](start = 8, length = 21)

(void)

src/sm.c
```diff
@@ -0,0 +1,331 @@
...
+        LONG64 b_now = InterlockedCompareExchange64(&sm->b_now, 0, -1);
+        if (b_now != -1)
+        {
+            /*Codes_SRS_SM_02_012: [ If sm_open_end doesn't follow a call to sm_open_begin then sm_open_end shall return. ]*/
+            LogError("cannot execute, name=%s, b_now=%" PRId64 "", sm->name, b_now);
+            result = SM_EXEC_REFUSED;
+        }
+        else
+        {
+            InterlockedIncrement64(&sm->n);
```
dcristoloveanu, line 106: >InterlockedIncrement64 [](start = 12, length = 22)

(void)

tests/sm_int/main.c
```diff
@@ -0,0 +1,11 @@
+// Copyright(C) Microsoft Corporation.All rights reserved.
```
dcristoloveanu, line 1: Likely MIT as it is in the open

dcristoloveanu: Here and everywhere

---
In reply to: [421813759](https://github.com/Azure/azure-c-util/pull/35#discussion_r421813759) [](ancestors = 421813759)

src/sm.c
```diff
@@ -0,0 +1,331 @@
...
+        /*cannot really check that it is paired with a _close_begin without moving b_now through user land, not going to do that. That solution woould push the b_now of the "close" in user land. And user has to return with the same b_now here and check that it is the same*/
+        /*Codes_SRS_SM_02_019: [ sm_close_end shall switch b_now to -1, n to 0 and e to 0. ]*/
+        (void)InterlockedExchange64(&sm->b_now, -1);
+        (void)InterlockedExchange64(&sm->n, 0);
+        (void)InterlockedExchange64(&sm->e, 0);
+        (void)InterlockedExchange(&sm->e_done, 0);
+    }
+}
+
+SM_RESULT sm_begin(SM_HANDLE sm)
```
dcristoloveanu, line 230: Probably not actionable in this review:

When I wrote my first attempt I went down the same route: having the user of the state manager call into a pair of functions to mark where the operation (like a write) starts and end.
That obviously still requires the user to write something like:

if (sm_begin() != OK)
{
   ... log error ...
}
else
{
    .. do some work ...
    sm_end() <- this gets executed if a later error happened, otherwise there is a similar call in the callback
}

I tried to think of a better way to tie together the begin/end so that they do not accidentally get mismatched but I could not really come up with anything interesting. Maybe you have better ideas.

dcristoloveanu: Sort of RAII like maybe with some macros.

---
In reply to: [421816537](https://github.com/Azure/azure-c-util/pull/35#discussion_r421816537) [](ancestors = 421816537)

src/sm.c
```diff
@@ -0,0 +1,331 @@
...
+        result = SM_ERROR;
+    }
+    else
+    {
+        LONG64 b_now_1 = InterlockedAdd64(&sm->b_now, 0);
+
+        /*Codes_SRS_SM_02_022: [ If there's a barrier set then sm_begin shall return SM_EXEC_REFUSED. ]*/
+        if ((b_now_1 & 1)==1)
+        {
+            result = SM_EXEC_REFUSED;
```
dcristoloveanu, line 246: >result = SM_EXEC_REFUSED; [](start = 12, length = 25)

I like the lack of logging here :-).

src/sm.c
```diff
@@ -0,0 +1,331 @@
...
+            LONG64 b_now_2 = InterlockedAdd64(&sm->b_now, 0);
+
+            /*Codes_SRS_SM_02_036: [ If the barrier changed after incrementing n then sm_begin shall increment e, signal a potential drain, and return SM_EXEC_REFUSED. ]*/
+            if (b_now_2 != b_now_1)
+            {
+                LONG64 e = InterlockedIncrement64(&sm->e);
+                LONG64 n = InterlockedAdd64(&sm->n, 0);
+                if (n - e == 1)
+                {
+                    InterlockedHL_SetAndWake((void*)&sm->e_done, 1);
```
dcristoloveanu, line 262: >InterlockedHL_SetAndWake [](start = 20, length = 24)

(void), or maybe it should check the return and logerror at least.

src/sm.c
```diff
@@ -0,0 +1,331 @@
...
+                }
+                result = SM_EXEC_REFUSED;;
```
dcristoloveanu, line 264: >;; [](start = 40, length = 2)

Only one ; necessary :-)

src/sm.c
```diff
@@ -0,0 +1,331 @@
...
+    else
+    {
+        /*Codes_SRS_SM_02_025: [ sm_end shall increment the number of executed APIs (e). ]*/
+        LONG64 e = InterlockedIncrement64(&sm->e);
+
+        /*Codes_SRS_SM_02_026: [ If n-e is 1 then sm_end shall wake up the waiting barrier. ]*/
+        LONG64 n = InterlockedAdd64(&sm->n, 0);
+        if (n - e == 1)
+        {
+            InterlockedHL_SetAndWake((void*)&sm->e_done, 1);
```
dcristoloveanu, line 295: >InterlockedHL_SetAndWake [](start = 12, length = 24)

(void) or check for error

src/sm.c
```diff
@@ -0,0 +1,331 @@
...
+}
+
+void sm_end(SM_HANDLE sm)
+{
+    SM_RESULT result;
+    /*Codes_SRS_SM_02_024: [ If sm is NULL then sm_end shall return. ]*/
+    if (sm == NULL)
+    {
+        LogError("invalid argument SM_HANDLE sm=%p", sm);
+        result = MU_FAILURE;
```
dcristoloveanu, line 284: >MU_FAILURE [](start = 17, length = 10)

Missing macro_utils in the include list

src/sm.c
```diff
@@ -0,0 +1,331 @@
...
+{
+    /*Codes_SRS_SM_02_032: [ If sm is NULL then sm_barrier_end shall return. ]*/
+    if (sm == NULL)
+    {
+        LogError("invalid argument SM_HANDLE sm=%p", sm);
+    }
+    else
+    {
+        /*Codes_SRS_SM_02_033: [ sm_barrier_end shall increment the number of executed operations (e), increment b_now and return. ]*/
+        InterlockedIncrement64(&sm->e);
```
dcristoloveanu, line 327: >InterlockedIncrement64 [](start = 8, length = 22)

(void)

dcristoloveanu: Here and below.

---
In reply to: [421818062](https://github.com/Azure/azure-c-util/pull/35#discussion_r421818062) [](ancestors = 421818062)

src/sm.c
```diff
@@ -0,0 +1,331 @@
...
+    }
+    else
+    {
+        /*Codes_SRS_SM_02_008: [ If b_now is not -1 then sm_open_begin shall fail and return SM_EXEC_REFUSED. ]*/
+        /*Codes_SRS_SM_02_009: [ sm_open_begin shall set b_now to 0, succeed and return SM_EXEC_GRANTED. ]*/
+        LONG64 b_now = InterlockedCompareExchange64(&sm->b_now, 0, -1);
+        if (b_now != -1)
+        {
+            /*Codes_SRS_SM_02_012: [ If sm_open_end doesn't follow a call to sm_open_begin then sm_open_end shall return. ]*/
+            LogError("cannot execute, name=%s, b_now=%" PRId64 "", sm->name, b_now);
```
dcristoloveanu, line 101: >cannot execut [](start = 22, length = 13)

This can say that we cannot execute the begin open

src/sm.c
```diff
@@ -0,0 +1,331 @@
...
+        result = SM_ERROR;
+    }
+    else
+    {
+        /*Codes_SRS_SM_02_008: [ If b_now is not -1 then sm_open_begin shall fail and return SM_EXEC_REFUSED. ]*/
+        /*Codes_SRS_SM_02_009: [ sm_open_begin shall set b_now to 0, succeed and return SM_EXEC_GRANTED. ]*/
+        LONG64 b_now = InterlockedCompareExchange64(&sm->b_now, 0, -1);
+        if (b_now != -1)
+        {
+            /*Codes_SRS_SM_02_012: [ If sm_open_end doesn't follow a call to sm_open_begin then sm_open_end shall return. ]*/
```
dcristoloveanu, line 100: >then sm_open_end shall return. [](start = 91, length = 30)

This spec is misplaced, it pertains to sm_open_end.
I noticed it looking at sm_open_end to see if we check in anyway that sm_open_end changes the state of the module only if it is after sm_open_begin.

tests/sm_int/sm_int.c
```diff
@@ -0,0 +1,264 @@
...
+
+}THREADS_COMMON;
+
+#define THREAD_TYPE_VALUES \
+    THREAD_BARRIER, \
+    THREAD_NON_BARRIER
+
+MU_DEFINE_ENUM(THREAD_TYPE, THREAD_TYPE_VALUE)
+
+static DWORD WINAPI barrier_thread(
```
dcristoloveanu, line 65: Can we please use ThreadAPI_ APIs?
They provide the same functionally and we are starting the Linux build work so it is easier if we use the abstractions rather than OS specific APIs.

tests/sm_int/sm_int.c
```diff
@@ -0,0 +1,264 @@
...
+
+TEST_FUNCTION_INITIALIZE(init)
+{
+
+}
+
+TEST_FUNCTION_CLEANUP(cleanup)
+{
+
+}
```
dcristoloveanu, line 171: Can be removed.

tests/sm_int/sm_int.c
```diff
@@ -0,0 +1,264 @@
...
+                ASSERT_IS_TRUE(
+                    (WAIT_OBJECT_0 <= waitResult) &&
+                    (waitResult <= WAIT_OBJECT_0 + n_barrier_threads - 1)
+                );
+            }
+
+            /*verify the all numbers written by barriers are greater than all previous numbers*/
+            verify(data);
+
+            printf("took %f ms, non_barrier_grants=%" PRId32 ", non_barrier_refusals=%" PRId64 " barrier_grants=%" PRId32 ", barrier_refusals=%" PRId64 "\n", timer_global_get_elapsed_ms() - data->startTimems,
```
dcristoloveanu, line 244: >printf [](start = 12, length = 6)

LogInfo please?

tests/sm_int/sm_int.c
```diff
@@ -0,0 +1,264 @@
...
+
+    double startTimems;
+
+}THREADS_COMMON;
+
+#define THREAD_TYPE_VALUES \
+    THREAD_BARRIER, \
+    THREAD_NON_BARRIER
+
+MU_DEFINE_ENUM(THREAD_TYPE, THREAD_TYPE_VALUE)
```
dcristoloveanu, line 63: >MU_DEFINE_ENUM [](start = 0, length = 14)

Needs macro_utils include.

src/sm.c
```diff
@@ -0,0 +1,331 @@
...
+        /*Codes_SRS_SM_02_016: [ sm_close_begin shall wait for all previous operations to end. ]*/
+        /*Codes_SRS_SM_02_029: [ sm_barrier_begin shall wait for the completion of all the previous operations. ]*/
+        while (get_n_minus_e(sm) != 1)
+        {
+            if (InterlockedHL_WaitForValue(&sm->e_done, 1, INFINITE) != INTERLOCKED_HL_OK)
+            {
+                /*Codes_SRS_SM_02_031: [ If there are any failures then sm_barrier_begin shall fail and return SM_ERROR. ]*/
+                /*Codes_SRS_SM_02_034: [ If there are any failures then sm_close_begin shall fail and return SM_ERROR. ]*/
+                LogError("Catastrophic failure: InterlockedHL_WaitForValue(&sm->e_done, 1, INFINITE), name=%s, n=%" PRId64 "", sm->name, n);
+                InterlockedIncrement64(&sm->e); /*this is pretty fatal here - the wait failed... */
```
dcristoloveanu, line 176: >InterlockedIncrement64 [](start = 16, length = 22)

(void)

src/sm.c
```diff
@@ -0,0 +1,331 @@
...
+
+    /*Codes_SRS_SM_02_014: [ sm_close_begin shall set lowest bit of b_now to 1. ]*/
+    /*Codes_SRS_SM_02_028: [ If b_now has least significand bit set to 1 then sm_barrier_begin shall fail and SM_EXEC_REFUSED. ]*/
+    /*Codes_SRS_SM_02_020: [ If there was no sm_open_begin/sm_open_end called previously, sm_close_begin shall fail and SM_EXEC_REFUSED. ]*/
+    if ((b_now & 1) == 0)
+    {
+        InterlockedExchange(&sm->e_done, 0);
+
+        LONG64 n = InterlockedIncrement64(&sm->n);
+        (void)n;
```
dcristoloveanu, line 160: >   (void)n; [](start = 5, length = 11)

Not needed.

src/sm.c
```diff
@@ -0,0 +1,331 @@
...
+{
+    SM_RESULT result;
+    LONG64 b_now = InterlockedOr64(&sm->b_now, 1);
+
+    /*Codes_SRS_SM_02_014: [ sm_close_begin shall set lowest bit of b_now to 1. ]*/
+    /*Codes_SRS_SM_02_028: [ If b_now has least significand bit set to 1 then sm_barrier_begin shall fail and SM_EXEC_REFUSED. ]*/
+    /*Codes_SRS_SM_02_020: [ If there was no sm_open_begin/sm_open_end called previously, sm_close_begin shall fail and SM_EXEC_REFUSED. ]*/
+    if ((b_now & 1) == 0)
+    {
+        InterlockedExchange(&sm->e_done, 0);
```
dcristoloveanu, line 157: >InterlockedExchange [](start = 8, length = 19)

(void)

src/sm.c
```diff
@@ -0,0 +1,331 @@
...
+        /*Codes_SRS_SM_02_030: [ sm_barrier_begin shall succeed and return SM_EXEC_GRANTED. ]*/
+        /*Codes_SRS_SM_02_017: [ sm_close_begin shall succeed and return SM_EXEC_GRANTED. ]*/
+        result = SM_EXEC_GRANTED;
+
+        /*drain previous executions*/
+        /*Codes_SRS_SM_02_016: [ sm_close_begin shall wait for all previous operations to end. ]*/
+        /*Codes_SRS_SM_02_029: [ sm_barrier_begin shall wait for the completion of all the previous operations. ]*/
+        while (get_n_minus_e(sm) != 1)
+        {
+            if (InterlockedHL_WaitForValue(&sm->e_done, 1, INFINITE) != INTERLOCKED_HL_OK)
```
dcristoloveanu, line 171: Why not wait for value on n-1?

src/sm.c
```diff
@@ -0,0 +1,331 @@
...
+    else
+    {
+        /*Codes_SRS_SM_02_015: [ If setting the lowest bit b_now to 1 fails then sm_close_begin shall return SM_EXEC_REFUSED. ]*/
+        result = SM_EXEC_REFUSED;
+    }
+    return result;
+}
+
+
+SM_RESULT sm_close_begin(SM_HANDLE sm)
```
dcristoloveanu, line 195: >sm_close_begin [](start = 10, length = 14)

I know the spec says that user is in charge of pairing things. But!
We have a very limited number of actions that need pairing.
open, generic barrier (which includes close).

As the code is written now I could call begin_close and then sm_end, the code would execute but leave the state manager maybe in a bad state.
I would say that of course the caller code should not call sm_barrier_end. But it seems that state manager could detect these and at least log.
As a matter of fact state manager does detect calling begin_close before begin_open, but not between begin_open and end_open.

It seems to me that we have a pretty standard state machine, with several states:
NOT_OPENED, OPENING (between open_begin and open+end), OPENED, CLOSING (between close_begin and close_end).

And it looks to me that this state machine has some advantages:
- attempting to open something that opens already will fail (for example a user of the module attempting to call 2 times open)
- attempting to close something that is already closing will also fail
- attempting to close while in opening will also fail

Being that state manager does know of this state (it does have APIs with open/close) I believe it should validate the calls coming from the user. If we do not validate the calls for the user then the user module has to absolutely make sure that no call to begin_open happens between begin_open and end_open.
To achieve that the user module would have to have a state machine with NOT_OPENED, OPENING, OPEN, CLOSING.
We can push this check higher and higher on the stack until it finally will reach a user whose code is not under our control (SF, Hillary, etc.). And when that happens we need to handle these cases at that layer.

Then my question would be: if state manager knows about open/close, why not have the validation done in state manager?
Being a limited number of states I am sure they can be modeled with very few bits :-) and we have an int64 at our fingertips.
One such bit I believe is used today in b_now for storing that state manager is not open (-1) as opposed to another value (check made in open_begin).

I feel a bit bad to bring it up now after the code is written, but I think this will make things better for us and I really want to use this with minimal overhead for myself :-).
Maybe next time we should discuss it before the code is there... On the other hand I understand the need for a prototype first which probably let to the whole thing being written.

src/sm.c
```diff
@@ -0,0 +1,331 @@
...
+        /*Codes_SRS_SM_02_017: [ sm_close_begin shall succeed and return SM_EXEC_GRANTED. ]*/
+        result = SM_EXEC_GRANTED;
+
+        /*drain previous executions*/
+        /*Codes_SRS_SM_02_016: [ sm_close_begin shall wait for all previous operations to end. ]*/
+        /*Codes_SRS_SM_02_029: [ sm_barrier_begin shall wait for the completion of all the previous operations. ]*/
+        while (get_n_minus_e(sm) != 1)
+        {
+            if (InterlockedHL_WaitForValue(&sm->e_done, 1, INFINITE) != INTERLOCKED_HL_OK)
+            {
```
dcristoloveanu, line 172: So this part here is waiting for "pending requests to get to 0". It can be named "the difference between n and e is 1" but essentially that what it does. Honestly the problem we had with the pending request count was that there was no check before incrementing it if someone wants to take the barrier.
This is basically a smarted SRL lcock (because "SRW locks are neither fair nor FIFO." - from MSDN)

So maybe modelling this as a count + a lock bit is actually not that bad - for sure I find it more readable when we need to debug ...
I think it's worth talking about this as we plan to use this in all our code - tomorrow morning maybe?

dcristoloveanu:
I highly commend doing this, I'd like to talk about 2 things:
- modeling the states so that we do validate bad requests from the user of the module
- explore using count of requests + barrier bit. This is simply because I think it makes it more debugable and could simplify the code also.

Whenever you have time!

src/sm.c
anporumb: As always, the anguish produced by a comment is directly proportional to the value of the comment.

Shall rethink.

PR #43, opened 2020-05-20
Interlocked tests

parth21999:
- Added integration tests for the `interlocked` module
- Added unit tests for the Windows implementation of the `interlocked` module
- Added the Windows implementation of the `interlocked` module


tests/interlocked_win32_ut/interlocked_win32_ut.c
```diff
@@ -0,0 +1,708 @@
...
+}
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
+#define ENABLE_MOCKS
+#include "umock_c/umock_c_prod.h"
+MOCKABLE_FUNCTION(, LONG, mock_InterlockedAdd, volatile LONG*, addend, LONG, value);
```
mattdurak, line 47: >LONG [](start = 20, length = 4)

should include "windows.h"

adapters/interlocked_win32.c
```diff
@@ -1,4 +1,318 @@
...
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-#include "azure_c_util/interlocked.h"
\ No newline at end of file
+#include "azure_c_util/interlocked.h"
+#include "windows.h"
+
+IMPLEMENT_MOCKABLE_FUNCTION(, int32_t, interlocked_add, volatile int32_t*, addend, int32_t, value) 
+{
+	/*Codes_SRS_INTERLOCKED_43_001 [ interlocked_add shall atomically add *addend with value and store the result in *addend.]*/
```
mattdurak, line 9: >	 [](start = 0, length = 1)

please fix your VS settings to use 4 spaces instead of tabs

adapters/interlocked_win32.c
```diff
@@ -1,4 +1,318 @@
...
+}
+
+IMPLEMENT_MOCKABLE_FUNCTION(, int8_t, interlocked_exchange_8, volatile int8_t*, target, int8_t, value)
+{
+	/*Codes_SRS_INTERLOCKED_43_017 [ interlocked_exchange_8 shall set the 8-bit variable pointed to by target to value as an atomic operation.]*/
+	/*Codes_SRS_INTERLOCKED_43_049: [ interlocked_exchange_8 shall return the initial value pointed to by target. ]*/
+	/*Codes_SRS_INTERLOCKED_WIN32_43_033: [ interlocked_exchange_8 shall call InterlockedExchange8 from windows.h. ]*/
+	/*Codes_SRS_INTERLOCKED_WIN32_43_034: [ interlocked_exchange_8 shall return the initial value pointed to by target. ]*/
+
+	return (int8_t)InterlockedExchange8((volatile char*)target, (char)value);
```
mattdurak, line 177: >char [](start = 47, length = 4)

should be `CHAR` for these two

adapters/interlocked_win32.c
```diff
@@ -1,4 +1,318 @@
...
+}
+
+IMPLEMENT_MOCKABLE_FUNCTION(, int8_t, interlocked_and_8, volatile int8_t*, destination, int8_t, value)
+{
+	/*Codes_SRS_INTERLOCKED_43_005 [ interlocked_and_8 shall perform an atomic bitwise AND operation on the 8-bit integer values *destination and value and store the result in *destination]*/
+	/*Codes_SRS_INTERLOCKED_43_036: [ interlocked_and_8 shall return the initial value of *destination. ]*/
+	/*Codes_SRS_INTERLOCKED_WIN32_43_009: [ interlocked_and_8 shall call InterlockedAnd8 from windows.h. ]*/
+	/*Codes_SRS_INTERLOCKED_WIN32_43_010: [ interlocked_and_8 shall return the initial value of *destination. ]*/
+
+	return (int8_t)InterlockedAnd8((volatile char*)destination, (char)value);
```
mattdurak, line 54: >char [](start = 42, length = 4)

`CHAR`

adapters/interlocked_win32.c
```diff
@@ -1,4 +1,318 @@
...
+}
+
+IMPLEMENT_MOCKABLE_FUNCTION(, int8_t, interlocked_or_8, volatile int8_t*, destination, int8_t, value)
+{
+	/*Codes_SRS_INTERLOCKED_43_027 [ interlocked_or_8 shall perform an atomic bitwise OR operation on the 8-bit integers *destination and value and store the result in destination.]*/
+	/*Codes_SRS_INTERLOCKED_43_058: [ interlocked_or_8 shall return the initial value of *destination. ]*/
+	/*Codes_SRS_INTERLOCKED_WIN32_43_053: [ interlocked_or_8 shall call InterlockedOr8 from windows.h. ]*/
+	/*Codes_SRS_INTERLOCKED_WIN32_43_054: [ interlocked_or_8 shall return the initial value of *destination. ]*/
+
+	return (int8_t)InterlockedOr8((volatile char*)destination, (char)value);
```
mattdurak, line 277: >char [](start = 41, length = 4)

`CHAR`

tests/interlocked_int/interlocked_int.c
```diff
@@ -0,0 +1,855 @@
...
+#include <stdint.h>
+#endif
+
+
+#include "testrunnerswitcher.h"
+
+static TEST_MUTEX_HANDLE g_testByTest;
+
+
+
```
mattdurak, line 20: nitpick: You can collapse these empty lines into 1, same with removing the empty lines 31/32, 37, 47. (not a big deal)

tests/interlocked_int/interlocked_int.c
```diff
@@ -0,0 +1,855 @@
...
+TEST_FUNCTION_CLEANUP(d)
+{
+    TEST_MUTEX_RELEASE(g_testByTest);
+}
+
+/*
+Tests_SRS_INTERLOCKED_43_001 [ interlocked_add shall atomically add *addend with value and store the result in *addend.]
+Tests_SRS_INTERLOCKED_43_032: [interlocked_add shall return the result of the addition.]
+*/
```
mattdurak, line 58: nit: requirements are usually written as:

```
/*Tests_*/
/*Tests_*/
```

Instead of 

```
/*
Tests_
Tests_
*/
```

Just a consistency thing

tests/interlocked_int/interlocked_int.c
```diff
@@ -0,0 +1,855 @@
...
+{
+    ///arrange
+    volatile uint32_t destination = 0xF0F0F0F0;
+    uint32_t value = 0x0F0F0F0F;
+
+    ///act
+    uint32_t return_val = (uint32_t)interlocked_and((volatile int32_t*)&destination, (int32_t)value);
+
+    ///assert
+    ASSERT_ARE_EQUAL(uint32_t, 0, destination);
```
mattdurak, line 89: >0 [](start = 31, length = 1)

I'd consider changing the values so this results in a non-zero value. 0 seems like a default value, so the test might be a little better to check for something else (e.g. what if the implementation was just setting destination to 0 every time?).

mattdurak: This is for here and everywhere

---
In reply to: [427689648](https://github.com/Azure/azure-c-util/pull/43#discussion_r427689648) [](ancestors = 427689648)

tests/interlocked_win32_ut/mock_interlocked.c
```diff
@@ -0,0 +1,101 @@
...
+#define InterlockedXor64 mock_InterlockedXor64
+#undef InterlockedXor8 
+#define InterlockedXor8 mock_InterlockedXor8
+#undef ENABLE_MOCKS
+
+LONG mock_InterlockedAdd(volatile LONG* addend, LONG value);
+LONG mock_InterlockedAnd(volatile LONG* destination, LONG value);
+SHORT mock_InterlockedAnd16(volatile SHORT* destination, SHORT value);
+LONG64 mock_InterlockedAnd64(volatile LONG64* destination, LONG64 value);
+char mock_InterlockedAnd8(volatile char* destination, char value);
```
mattdurak, line 74: >char [](start = 0, length = 4)

these should also be `CHAR`

tests/interlocked_win32_ut/interlocked_win32_ut.c
```diff
@@ -0,0 +1,708 @@
...
+    comperand_result[0] = INT64_MAX;
+    comperand_result[1] = INT64_MIN;
+
+    int64_t exchange_high = 2;
+    int64_t exchange_low = 3;
+
+    STRICT_EXPECTED_CALL(mock_InterlockedCompareExchange128((volatile LONG64*)destination, (LONG64)exchange_high, (LONG64)exchange_low, (LONG64*)comperand_result))
+        .SetReturn(0);
+    ///act
+    bool return_val = interlocked_compare_exchange_128(destination, exchange_high, exchange_low, comperand_result);
```
mattdurak, line 275: >bool [](start = 4, length = 4)

include stdbool.h

tests/interlocked_win32_ut/interlocked_win32_ut.c
```diff
@@ -0,0 +1,708 @@
...
+    destination[0] = INT64_MAX;
+    destination[1] = INT64_MIN;
+    comperand_result[0] = INT64_MAX;
+    comperand_result[1] = INT64_MIN;
+
+    int64_t exchange_high = 2;
+    int64_t exchange_low = 3;
+
+    STRICT_EXPECTED_CALL(mock_InterlockedCompareExchange128((volatile LONG64*)destination, (LONG64)exchange_high, (LONG64)exchange_low, (LONG64*)comperand_result))
+        .SetReturn(1);
```
mattdurak, line 243: >1 [](start = 19, length = 1)

`TRUE` (all-caps TRUE)

tests/interlocked_win32_ut/interlocked_win32_ut.c
```diff
@@ -0,0 +1,708 @@
...
+    destination[0] = INT64_MAX;
+    destination[1] = INT64_MIN;
+    comperand_result[0] = INT64_MAX;
+    comperand_result[1] = INT64_MIN;
+
+    int64_t exchange_high = 2;
+    int64_t exchange_low = 3;
+
+    STRICT_EXPECTED_CALL(mock_InterlockedCompareExchange128((volatile LONG64*)destination, (LONG64)exchange_high, (LONG64)exchange_low, (LONG64*)comperand_result))
+        .SetReturn(0);
```
mattdurak, line 273: >0 [](start = 19, length = 1)

`FALSE`

tests/CMakeLists.txt
```diff
@@ -30,12 +30,14 @@ if(${run_unittests})
         add_subdirectory(srw_lock_ut)
         add_subdirectory(timer_win32_ut)
         add_subdirectory(interlocked_hl_win32_ut)
+	add_subdirectory(interlocked_win32_ut)
```
mattdurak, line 33: >	 [](start = 0, length = 1)

tab -> spaces

adapters/interlocked_win32.c
```diff
@@ -1,4 +1,318 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-#include "azure_c_util/interlocked.h"
\ No newline at end of file
+#include "azure_c_util/interlocked.h"
+#include "windows.h"
```
mattdurak, line 5: This include should go above. We normally include all standard headers, then system headers like windows.h, then project headers.

Also, you should include stdint.h for the intX_t types.

adapters/interlocked_win32.c
```diff
@@ -1,4 +1,318 @@
...
+{
+	/*Codes_SRS_INTERLOCKED_43_006 [ interlocked_compare_exchange shall compare the 32-bit integers pointed to by destination and comperand. If they are equal, *destination is set to exchange. These operations are performed atomically.*]*/
+	/*Codes_SRS_INTERLOCKED_43_037: [ interlocked_compare_exchange shall return the initial value of *destination.]*/
+	/*Codes_SRS_INTERLOCKED_WIN32_43_011: [ interlocked_compare_exchange shall call InterlockedCompareExchange from windows.h. ]*/
+	/*Codes_SRS_INTERLOCKED_WIN32_43_012: [ interlocked_compare_exchange shall return the initial value of *destination. ]*/
+
+	return (int32_t)InterlockedCompareExchange((volatile LONG*)destination, (LONG)exchange, (LONG)comperand);
+}
+
+IMPLEMENT_MOCKABLE_FUNCTION(, bool, interlocked_compare_exchange_128, volatile int64_t*, destination, int64_t, exchange_high, int64_t, exchange_low, int64_t*, comperand_result)
```
mattdurak, line 67: >interlocked_compare_exchange_128 [](start = 36, length = 32)

So the build gates do run on x86 apparently. That means all of the ICE_128 functions will need to be conditionally removed.

I don't know if there's a standard way to do it, but for Windows at least you could wrap these in `#ifdef _WIN64`.

mattdurak: That's one of your build failures at least

---
In reply to: [427697193](https://github.com/Azure/azure-c-util/pull/43#discussion_r427697193) [](ancestors = 427697193)

adapters/interlocked_win32.c
```diff
@@ -1,4 +1,318 @@
...
+#include "windows.h"
+
+IMPLEMENT_MOCKABLE_FUNCTION(, int32_t, interlocked_add, volatile int32_t*, addend, int32_t, value) 
+{
+	/*Codes_SRS_INTERLOCKED_43_001 [ interlocked_add shall atomically add *addend with value and store the result in *addend.]*/
+	/*Codes_SRS_INTERLOCKED_43_032: [ interlocked_add shall return the result of the addition.]*/
+	/*Codes_SRS_INTERLOCKED_WIN32_43_001: [ interlocked_add shall call InterlockedAdd from windows.h. ]*/
+	/*Codes_SRS_INTERLOCKED_WIN32_43_002: [ interlocked_add shall return the result of the addition. ]*/
+
+	return (int32_t)InterlockedAdd((volatile LONG*)addend, (LONG)value);
```
anporumb, line 14: >(int32_t) [](start = 8, length = 9)

why is the cast necessary? Unless there's a strong reason to have the cast, remove. Here and everywhere.

adapters/interlocked_win32.c
```diff
@@ -1,4 +1,318 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-#include "azure_c_util/interlocked.h"
\ No newline at end of file
+#include "azure_c_util/interlocked.h"
```
anporumb, line 4: >include [](start = 1, length = 7)

missing #include <stdint.h>

tests/CMakeLists.txt
anporumb, line 33: >	 [](start = 0, length = 1)

please replace TABs with SPACEs. Only space is the final frontier!

anporumb: a matter of "here and everywhere"

---
In reply to: [427720271](https://github.com/Azure/azure-c-util/pull/43#discussion_r427720271) [](ancestors = 427720271)

tests/interlocked_int/interlocked_int.c
```diff
@@ -0,0 +1,855 @@
...
+{
+    ///arrange
+    volatile int32_t addend = INT32_MAX;
+    int32_t value = INT32_MIN;
+
+    ///act
+    int32_t return_val = interlocked_add(&addend, value);
+
+    ///assert
+    ASSERT_ARE_EQUAL(int32_t, -1, addend);
```
anporumb, line 70: >addend [](start = 34, length = 6)

Frankly, seeing as "added" is in interlocked variable, it should be read with an interlocked operation.

here and everywhere.

tests/interlocked_int/interlocked_int.c
```diff
@@ -0,0 +1,855 @@
...
+/*
+Tests_SRS_INTERLOCKED_43_007 [ interlocked_compare_exchange_128 shall compare *destination with *comperand_result. If they are equal, destination[0] is set to exchange_low and destination[1] is set to exchange_high. These operations are performed atomically.]
+Tests_SRS_INTERLOCKED_43_039: [ interlocked_compare_exchange_128 shall store the initial value of *destination in *comperand_result regardless of the result of the comparison.` ]
+Tests_SRS_INTERLOCKED_43_038: [ interlocked_compare_exchange_128 shall return true if *comperand_result equals the original value of *destination.]
+*/
+
+TEST_FUNCTION(interlocked_compare_exchange_128_equal_case)
+{
+    ///arrange
+    volatile int64_t* destination = (volatile int64_t*)_aligned_malloc(2 * sizeof(int64_t), 16);
```
anporumb, line 197: >_aligned_malloc [](start = 55, length = 15)

malloc simple is fine.

here's the doc for it: 

```
The storage space pointed to by the return value is guaranteed to be suitably aligned for storage of any type of object that has an alignment requirement less than or equal to that of the fundamental alignment. (In Visual C++, the fundamental alignment is the alignment that's required for a double, or 8 bytes. In code that targets 64-bit platforms, it’s 16 bytes.)
```

from [https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/malloc?view=vs-2019](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/malloc?view=vs-2019)

anporumb: here and everywhere

---
In reply to: [427721662](https://github.com/Azure/azure-c-util/pull/43#discussion_r427721662) [](ancestors = 427721662)

anporumb, line 197: >_aligned_malloc [](start = 55, length = 15)

why not volatile int64_t destination[2]?

anporumb: sort of here and everywhere

---
In reply to: [427722625](https://github.com/Azure/azure-c-util/pull/43#discussion_r427722625) [](ancestors = 427722625)

tests/interlocked_int/interlocked_int.c
```diff
@@ -0,0 +1,855 @@
...
+{
+    ///arrange
+    int value1 = 1;
+    int value2 = 2;
+    volatile void* destination = &value1;
+    void* comperand = &value1;
+    void* exchange = &value2;
+
+    ///act
+    void* return_val = interlocked_compare_exchange_pointer(&destination, exchange, comperand);
```
anporumb, line 348: Interesting how the compile decided not to barf here.

As I read destination, it is a pointer to something volatile. Itself, the pointer, is not volatile.

However, &destination is pointer to a pointer to something volatile. But it should be a pointer to a volatile pointer, (exactly because it is changed by the function call).

Interesting... 

Would you mind taking a look?

anporumb: seems like a case for here and everywhere with  pointers.... 

---
In reply to: [427736331](https://github.com/Azure/azure-c-util/pull/43#discussion_r427736331) [](ancestors = 427736331)

tests/interlocked_int/interlocked_int.c
```diff
@@ -0,0 +1,855 @@
...
+    ASSERT_ARE_EQUAL(int32_t, INT32_MIN, addend);
+    ASSERT_ARE_EQUAL(int32_t, INT32_MIN, return_val);
+
+}
+
+/*
+Tests_SRS_INTERLOCKED_43_012[ interlocked_decrement_16 shall atomically decrement (decrease by one) the 16-bit variable *addend.]
+Tests_SRS_INTERLOCKED_43_044: [ interlocked_decrement_16 shall return the decremented value. ]
+*/
+TEST_FUNCTION(interlocked_decrement_16_upper_bound)
```
anporumb, line 413: >interlocked_decrement_16_upper_bound [](start = 14, length = 36)

carefully chosing the bounds to avoid a underflow/overflow. but those will happen for real.

I believe it is better to have those cases exposed intentionally. That is a INTn_MAX that is incrementd and INTn_MIN that is decremented. Because sweeping under the rug maybe is a not a good idea. Should be explicit about the spiky cases.

tests/interlocked_int/interlocked_int.c
```diff
@@ -0,0 +1,855 @@
...
+{
+    ///arrange
+    volatile uint32_t destination = 0xF0F0F0F0;
+    uint32_t value = 0x0F0F0F0F;
+
+    ///act
+    uint32_t return_val = (uint32_t)interlocked_or((volatile int32_t*)&destination, (int32_t)value);
+
+    ///assert
+    ASSERT_ARE_EQUAL(uint32_t, -1, destination);
```
anporumb, line 718: >-1 [](start = 31, length = 2)

-1 and uint32_t (unsigned) don't play nice together. Another value is proper here.

anporumb: Seems like a case of "here and everywhere" when comparing unsigned values and -1.

---
In reply to: [427738341](https://github.com/Azure/azure-c-util/pull/43#discussion_r427738341) [](ancestors = 427738341)

tests/interlocked_win32_ut/mock_interlocked.c
```diff
@@ -0,0 +1,101 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#include "windows.h"
+
+#undef InterlockedAdd
```
anporumb, line 6: symbols should be undef'd only if they are #defined previously. Otherwise grizzly compiler will complain about it.

Interesting this did not happen here... are these all macros?

tests/interlocked_win32_ut/mock_interlocked.c
```diff
@@ -0,0 +1,101 @@
...
+#define InterlockedOr8 mock_InterlockedOr8
+#undef InterlockedXor 
+#define InterlockedXor mock_InterlockedXor
+#undef InterlockedXor16
+#define InterlockedXor16 mock_InterlockedXor16
+#undef InterlockedXor64 
+#define InterlockedXor64 mock_InterlockedXor64
+#undef InterlockedXor8 
+#define InterlockedXor8 mock_InterlockedXor8
+#undef ENABLE_MOCKS
```
anporumb, line 68: wondering what prompted this.... where was it #defined previously...

tests/interlocked_win32_ut/mock_interlocked.c
```diff
@@ -0,0 +1,101 @@
...
+
+LONG mock_InterlockedAdd(volatile LONG* addend, LONG value);
```
anporumb, line 70: why are these declarations needed?

tests/interlocked_win32_ut/interlocked_win32_ut.c
```diff
@@ -0,0 +1,708 @@
...
+#else
+#include <stdlib.h>
+#include <stddef.h>
+#endif
+
+#include "azure_macro_utils/macro_utils.h"
+
+#include "testrunnerswitcher.h"
+
+void* my_gballoc_malloc(size_t size)
```
anporumb, line 16: >* [](start = 4, length = 1)

static

tests/interlocked_win32_ut/interlocked_win32_ut.c
```diff
@@ -0,0 +1,708 @@
...
+{
+    return malloc(size);
+}
+
+void my_gballoc_free(void* ptr)
```
anporumb, line 21: >my_gballoc_free [](start = 5, length = 15)

static

tests/interlocked_win32_ut/interlocked_win32_ut.c
```diff
@@ -0,0 +1,708 @@
...
+
+BEGIN_TEST_SUITE(interlocked_win32_unittests)
+
+TEST_SUITE_INITIALIZE(suite_init)
+{
+    g_testByTest = TEST_MUTEX_CREATE();
+    ASSERT_IS_NOT_NULL(g_testByTest);
+    ASSERT_ARE_EQUAL(int, 0, umock_c_init(on_umock_c_error));
+    ASSERT_ARE_EQUAL(int, 0, umocktypes_windows_register_types());
+    REGISTER_UMOCK_ALIAS_TYPE(SHORT, int16_t);
```
anporumb, line 92: >REGISTER_UMOCK_ALIAS_TYPE(SHORT, int16_t); [](start = 4, length = 42)

hmmm... looks like umocktypes_windows_register_types should do that for you. Would you mind filing a bug against umock_c about this type?

tests/interlocked_win32_ut/interlocked_win32_ut.c
```diff
@@ -0,0 +1,708 @@
...
+}
+
+/*Tests_SRS_INTERLOCKED_WIN32_43_001: [interlocked_add shall call InterlockedAdd from windows.h.]*/
+/*Tests_SRS_INTERLOCKED_WIN32_43_002 : [interlocked_add shall return the result of the addition.]*/
+TEST_FUNCTION(interlocked_add_calls_InterlockedAdd)
+{
+    ///arrange
+    volatile int32_t addend = INT32_MAX;
+    int32_t value = INT32_MIN;
+    STRICT_EXPECTED_CALL(mock_InterlockedAdd((volatile LONG*)&addend, (LONG)value))
```
anporumb, line 125: >(volatile LONG*) [](start = 45, length = 16)

I am only going to ask this (the second time). why is the cast needed?

My view is simple: either the cast does something to the value, in which case the code will be wrong, as it would not match the definiton domains for LONG/int32_t 

OR

the cast does nothing, in which case why is it needed?

tests/interlocked_win32_ut/interlocked_win32_ut.c
```diff
@@ -0,0 +1,708 @@
...
+        .SetReturn(-1);
```
anporumb, line 126: >SetReturn [](start = 9, length = 9)

Or use a REGISTER_GLOBAL_MOCK_HOOK and in that case, the result is set by the target (real_... ) function. Frankly, I would do the real_ thingie. Saves me some typing. And maybe accuracy.

anporumb:
> #endif

isn't this header missing functions?

where  is interlocked_add_64? Just asking for a friend.

---
Refers to: inc/azure_c_util/interlocked.h:58 in 6a604e4. [](commit_id = 6a604e42b324f9dd0f03955cee77e1281451a409, deletion_comment = False)

adapters/interlocked_win32.c
dcristoloveanu, line 5: We usually put windows before other things.

The order we try to stick to is:

- C standard headers (stdint.h, etc.)
- OS headers (windows, etc.)
- lib headers (azure_c_macro_utils, azure_c_utils, etc.)
Others

adapters/interlocked_win32.c
```diff
@@ -1,4 +1,318 @@
...
+#include "windows.h"
+
+IMPLEMENT_MOCKABLE_FUNCTION(, int32_t, interlocked_add, volatile int32_t*, addend, int32_t, value)
```
dcristoloveanu, line 7: >int32_t [](start = 30, length = 7)

Missing include of stdint.h

tests/interlocked_win32_ut/interlocked_win32_ut.c
```diff
@@ -0,0 +1,708 @@
...
+}
+
+/*Tests_SRS_INTERLOCKED_WIN32_43_007: [interlocked_and_64 shall call InterlockedAnd64 from windows.h.]*/
+/*Tests_SRS_INTERLOCKED_WIN32_43_008 : [interlocked_and_64 shall return the initial value of * destination.]*/
+TEST_FUNCTION(interlocked_and_64_calls_InterlockedAnd64)
+{
+    ///arrange
+    volatile uint64_t destination = 0xF0F0F0F0F0F0F0F0;
+    uint64_t value = 0x0F0F0F0F0F0F0F0F;
+    STRICT_EXPECTED_CALL(mock_InterlockedAnd64((volatile LONG64*)&destination, (LONG64)value))
```
anporumb, line 180: >mock_InterlockedAnd64 [](start = 25, length = 21)

mock_InterlockedAnd64 seems to return some LONG64. Which is signed.

0xf0f0... is that meant to be the representation of some negative value?

adapters/interlocked_win32.c
dcristoloveanu, line 9: >	 [](start = 0, length = 1)

Not a big deal, but can we use spaces for indentation? Makes it consistent across the team. But again, not that big of a deal.

tests/interlocked_win32_ut/interlocked_win32_ut.c
anporumb: this is spefically about .SetReturn(...)

---
In reply to: [427743537](https://github.com/Azure/azure-c-util/pull/43#discussion_r427743537) [](ancestors = 427743537)

adapters/interlocked_win32.c
dcristoloveanu, line 14: >(int32_t) [](start = 8, length = 9)

Unnecessary cast I believe

dcristoloveanu, line 14: >(LONG) [](start = 56, length = 6)

Unnecessary cast suspected.

tests/interlocked_win32_ut/interlocked_win32_ut.c
```diff
@@ -0,0 +1,708 @@
...
+    ASSERT_ARE_EQUAL(char_ptr, umock_c_get_expected_calls(), umock_c_get_actual_calls());
+    ASSERT_ARE_EQUAL(int32_t, INT32_MAX, return_val);
+}
+
+/*Tests_SRS_INTERLOCKED_WIN32_43_013: [interlocked_compare_exchange_128 shall call InterlockedCompareExchange128 from windows.h.]*/
+/*Tests_SRS_INTERLOCKED_WIN32_43_014 : [interlocked_compare_exchange_128 shall return true if* comperand_result equals the original value of* destination.]*/
+TEST_FUNCTION(interlocked_compare_exchange_128_calls_InterlockedCompareExchange128_true_case)
+{
+    ///arrange
+    volatile int64_t* destination = (volatile int64_t*)_aligned_malloc(2 * sizeof(int64_t), 16);
```
anporumb, line 231: >destination [](start = 22, length = 11)

destination[2] is equally good.

anporumb, line 231: >_aligned_malloc [](start = 55, length = 15)

can return NULL, by the way...

tests/interlocked_win32_ut/interlocked_win32_ut.c
```diff
@@ -0,0 +1,708 @@
...
+    int64_t exchange_high = 2;
+    int64_t exchange_low = 3;
+
+    STRICT_EXPECTED_CALL(mock_InterlockedCompareExchange128((volatile LONG64*)destination, (LONG64)exchange_high, (LONG64)exchange_low, (LONG64*)comperand_result))
+        .SetReturn(1);
+    ///act
+    bool return_val = interlocked_compare_exchange_128(destination, exchange_high, exchange_low, comperand_result);
+
+    ///assert
+    ASSERT_ARE_EQUAL(char_ptr, umock_c_get_expected_calls(), umock_c_get_actual_calls());
```
anporumb, line 248: >ASSERT_ARE_EQUAL [](start = 4, length = 16)

ASSERT seems to be lacking comperand_result[0] == 3; [1]==2. Please add.

tests/interlocked_win32_ut/interlocked_win32_ut.c
```diff
@@ -0,0 +1,708 @@
...
+    int64_t exchange_high = 2;
+    int64_t exchange_low = 3;
+
+    STRICT_EXPECTED_CALL(mock_InterlockedCompareExchange128((volatile LONG64*)destination, (LONG64)exchange_high, (LONG64)exchange_low, (LONG64*)comperand_result))
+        .SetReturn(0);
+    ///act
+    bool return_val = interlocked_compare_exchange_128(destination, exchange_high, exchange_low, comperand_result);
+
+    ///assert
+    ASSERT_ARE_EQUAL(char_ptr, umock_c_get_expected_calls(), umock_c_get_actual_calls());
```
anporumb, line 278: >ASSERT_ARE_EQUAL [](start = 4, length = 16)

ASSERT seems to be lacking comperand_result[0], comperand_result[1]

tests/interlocked_win32_ut/interlocked_win32_ut.c
```diff
@@ -0,0 +1,708 @@
...
+    volatile int32_t target = INT32_MIN;
+    int32_t value = INT32_MAX;
+    STRICT_EXPECTED_CALL(mock_InterlockedExchange((volatile LONG*)&target, (LONG)value))
+        .SetReturn(INT32_MIN);
+
+    ///act
+    int32_t return_val = interlocked_exchange(&target, value);
+
+    ///assert
+    ASSERT_ARE_EQUAL(char_ptr, umock_c_get_expected_calls(), umock_c_get_actual_calls());
```
anporumb, line 407: >ASSERT_ARE_EQUAL [](start = 4, length = 16)

should assert target to be something.

anporumb: sort of a here and everywhere matter

---
In reply to: [427744935](https://github.com/Azure/azure-c-util/pull/43#discussion_r427744935) [](ancestors = 427744935)

tests/interlocked_win32_ut/interlocked_win32_ut.c
```diff
@@ -0,0 +1,708 @@
...
+}
+
+/*Tests_SRS_INTERLOCKED_WIN32_43_039: [interlocked_exchange_pointer shall call InterlockedExchangePointer from windows.h.]*/
+/*Tests_SRS_INTERLOCKED_WIN32_43_040 : [interlocked_exchange_pointer shall return the initial address pointed to by the target parameter]*/
+TEST_FUNCTION(interlocked_exchange_pointer_calls_InterlockedExchagePointer)
+{
+    ///arrange
+    int value1 = 1;
+    int value2 = 2;
+    void* volatile target = &value1;
```
anporumb, line 508: >void* volatile target [](start = 4, length = 21)

Interesting, this pointer looks as I expect it to look. 

As opposed to a pointer from integration tests, which looked the opposite.

One of them **must** be wrong!

tests/interlocked_win32_ut/interlocked_win32_ut.c
```diff
@@ -0,0 +1,708 @@
...
+/*Tests_SRS_INTERLOCKED_WIN32_43_051: [interlocked_or_64 shall call InterlockedOr64 from windows.h.]*/
+/*Tests_SRS_INTERLOCKED_WIN32_43_052 : [interlocked_or_64 shall return the initial value of * destination.]*/
+TEST_FUNCTION(interlocked_or_64_calls_InterlockedOr64)
+{
+    ///arrange
+    volatile uint64_t destination = 0xF0F0F0F0F0F0F0F0;
+    uint64_t value = 0x0F0F0F0F0F0F0F0F;
+    STRICT_EXPECTED_CALL(mock_InterlockedOr64((volatile LONG64*)&destination, (LONG64)value))
+        .SetReturn(0xF0F0F0F0F0F0F0F0);
+    ///act
```
anporumb, line 615: > ///act [](start = 3, length = 7)

I would leave an empty line above each ///act 

./consider

tests/interlocked_win32_ut/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
...
+
+set(${theseTestsName}_c_files
+mock_interlocked.c
+)
+
+set(${theseTestsName}_h_files
+../../inc/azure_c_util/interlocked.h
+)
+
+build_c_tests(${theseTestsName} ON "tests/azure_c_util_tests" ADDITIONAL_LIBS synchronization azure_c_util_reals)
```
anporumb, line 20: >azure_c_util_reals [](start = 94, length = 18)

I did not notice any function being used from azure_c_util_reals.

What happens when you remove this? Maybe replace it with logging lib...

anporumb, line 20: >build_c_tests [](start = 0, length = 13)

need to take a second look, so "waiting".

tests/CMakeLists.txt
dcristoloveanu, line 33: >	 [](start = 0, length = 1)

Nit: this seems a bit misaligned

tests/interlocked_int/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
+#Copyright (c) Microsoft. All rights reserved.
```
dcristoloveanu, line 1: >yright (c) Mi [](start = 4, length = 13)

Missing the MIT thingie

tests/interlocked_int/interlocked_int.c
```diff
@@ -0,0 +1,855 @@
...
+BEGIN_TEST_SUITE(interlocked_int)
+
+
+TEST_SUITE_INITIALIZE(a)
+{
+    g_testByTest = TEST_MUTEX_CREATE();
+    ASSERT_IS_NOT_NULL(g_testByTest);
+
+    
+}
```
dcristoloveanu, line 33: Nitpick: some extra lines detected: this could make me think something is missing or has been removed when looking at the code. So to be neat would probably be nice to remove the extra lines.

tests/interlocked_int/interlocked_int.c
```diff
@@ -0,0 +1,855 @@
...
+    int32_t exchange = INT32_MIN;
+
+    ///act
+    int32_t return_val = interlocked_compare_exchange(&destination, exchange, comperand);
+
+    ///assert
+    ASSERT_ARE_EQUAL(int32_t, INT32_MIN, destination);
+    ASSERT_ARE_EQUAL(int32_t, INT32_MAX, return_val);
+}
+/*
```
dcristoloveanu, line 169: There seems to be a bit of inconsistency in the way the tests are laid out. For example there is no line between these 2, but the subsequent ones have a new line between the tests.
I can read all, but maybe you want to have a quick pass to make them neat and consistent.

tests/interlocked_int/interlocked_int.c
dcristoloveanu, line 197: >_aligned_malloc [](start = 55, length = 15)

I don't think we need _aligned_malloc. SImple malloc should be perfectly fine.

tests/interlocked_int/interlocked_int.c
```diff
@@ -0,0 +1,855 @@
...
+    _aligned_free((void*)destination);
+    _aligned_free(comperand_result);
+}
+
+
+/*
+Tests_SRS_INTERLOCKED_43_039: [ interlocked_compare_exchange_128 shall store the initial value of *destination in *comperand_result regardless of the result of the comparison.` ]
+Tests_SRS_INTERLOCKED_43_063: [ interlocked_compare_exchange_128 shall return false if *comperand_result does not equal the original value of *destination. ]
+*/
+TEST_FUNCTION(interlocked_compare_exchange_128_not_equal_case)
```
dcristoloveanu, line 228: >_case [](start = 56, length = 5)

Note on test names: Some time ago I read "The Art of unit testing" also known as the samurai book by Roy Osherove. Maybe you heard of it. If not it might be an interesting read. One of the things that stuck with me from that book were several rules/patterns for naming tests and how to write good tests.
One of them is that the test names should express the requirement. The reason to be verbose is to be able to immediately see in the test name what is there to be tested by that test function.

Thus I tend to name my tests like: interlocked_compare_exchange_128_does_not_exchange_when_not_equal

It's not that different from what you had, but as tests become more complex it is I think important to pay attention to what the name of the test is. Usually having the verb in it forces you to have a (slightly) better test name.

tests/interlocked_win32_ut/interlocked_win32_ut.c
dcristoloveanu, line 16: Somewhere here there should be a windows.gh include.

tests/interlocked_win32_ut/mock_interlocked.c
```diff
@@ -0,0 +1,101 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#include "windows.h"
+
+#undef InterlockedAdd 
+#define InterlockedAdd mock_InterlockedAdd
```
dcristoloveanu, line 7: >terlockedAdd mock_InterlockedAdd [](start = 10, length = 32)

I think these should be hooked with umock (there is a register hook macro)

tests/interlocked_int/interlocked_int.c
```diff
@@ -0,0 +1,855 @@
...
+    ///act
+    uint32_t return_val = (uint32_t)interlocked_and((volatile int32_t*)&destination, (int32_t)value);
+
+    ///assert
+    ASSERT_ARE_EQUAL(uint32_t, 0, destination);
+    ASSERT_ARE_EQUAL(uint32_t, 0xF0F0F0F0, return_val);
+}
+
+/*
+Tests_SRS_INTERLOCKED_43_003 [ interlocked_and_16 shall perform an atomic bitwise AND operation on the 16-bit integer values *destination and value and store the result in *destination.]
```
dcristoloveanu, line 94: >  [](start = 28, length = 1)

Missing :

tests/interlocked_int/interlocked_int.c
```diff
@@ -0,0 +1,855 @@
...
+    ///act
+    int32_t return_val = interlocked_add(&addend, value);
+
+    ///assert
+    ASSERT_ARE_EQUAL(int32_t, -1, addend);
+    ASSERT_ARE_EQUAL(int32_t, -1, return_val);
+}
+
+/*
+Tests_SRS_INTERLOCKED_43_002[interlocked_and shall perform an atomic bitwise AND operation on the 32 - bit integer values * destination and value and store the result in * destination.]
```
dcristoloveanu, line 75: >2[int [](start = 27, length = 5)

Missing :

tests/interlocked_int/interlocked_int.c
```diff
@@ -0,0 +1,855 @@
...
+
+}
+
+TEST_FUNCTION_CLEANUP(d)
+{
+    TEST_MUTEX_RELEASE(g_testByTest);
+}
+
+/*
+Tests_SRS_INTERLOCKED_43_001 [ interlocked_add shall atomically add *addend with value and store the result in *addend.]
```
dcristoloveanu, line 56: >01 [ i [](start = 26, length = 6)

Missing :

tests/interlocked_int/interlocked_int.c
```diff
@@ -0,0 +1,855 @@
...
+    uint16_t return_val = (uint16_t)interlocked_and_16((volatile int16_t*)&destination, (int16_t)value);
+
+
+    ///assert
+    ASSERT_ARE_EQUAL(uint16_t, 0, destination);
+    ASSERT_ARE_EQUAL(uint16_t, 0xF0F0, return_val);
+}
+
+/*
+Tests_SRS_INTERLOCKED_43_004 [ interlocked_and_64 shall perform an atomic bitwise AND operation on the 64-bit integer values *destination and value and store the result in *destination.]
```
dcristoloveanu, line 114: >4 [  [](start = 27, length = 4)

Seems to be everywhere that the : got removed

tests/interlocked_int/interlocked_int.c
```diff
@@ -0,0 +1,855 @@
...
+{
+    ///arrange
+    volatile uint64_t destination = 0xF0F0F0F0F0F0F0F0;
+    uint64_t value = 0x0F0F0F0F0F0F0F0F;
+
+    ///act
+    uint64_t return_val = (uint64_t)interlocked_or_64((volatile int64_t*)&destination, (int64_t)value);
+
+    ///assert
+    ASSERT_ARE_EQUAL(uint64_t, -1, destination);
```
dcristoloveanu, line 757: >uint64_t, -1 [](start = 21, length = 12)

We should not use -1 as expected, we should use an unsigned value

dcristoloveanu: Here and above in 738

---
In reply to: [427765880](https://github.com/Azure/azure-c-util/pull/43#discussion_r427765880) [](ancestors = 427765880)

dcristoloveanu: And in the other places too

---
In reply to: [427765967](https://github.com/Azure/azure-c-util/pull/43#discussion_r427765967) [](ancestors = 427765967,427765880)

tests/interlocked_int/interlocked_int.c
```diff
@@ -0,0 +1,855 @@
...
+}
+
+/*
+Tests_SRS_INTERLOCKED_43_023 [ interlocked_increment_64 shall atomically increment (increase by one) the 64-bit variable *addend.]
+Tests_SRS_INTERLOCKED_43_054: [ interlocked_increment_64 shall return the incremented value. ]
+*/
+TEST_FUNCTION(interlocked_increment_64_upper_bound)
+{
+    ///arrange
+    volatile int64_t addend = INT64_MAX - 1;
```
dcristoloveanu, line 676: > - 1 [](start = 39, length = 4)

This should also work for INT64_MAX, right?

dcristoloveanu:
Looks pretty good to me, I'd like to have a look after the comments are applied!

adapters/interlocked_win32.c
parth21999: In the docs it seems like they are using the native C types: [https://docs.microsoft.com/en-us/windows/win32/api/winnt/nf-winnt-interlockedand8](https://docs.microsoft.com/en-us/windows/win32/api/winnt/nf-winnt-interlockedand8)

---
In reply to: [427687405](https://github.com/Azure/azure-c-util/pull/43#discussion_r427687405) [](ancestors = 427687405)

mattdurak: You're right, InterlockedExchange8 says `CHAR` but InterlockedAnd8 says `char`.

---
In reply to: [428107038](https://github.com/Azure/azure-c-util/pull/43#discussion_r428107038) [](ancestors = 428107038,427687405)

tests/interlocked_int/interlocked_int.c
mattdurak: Yeah, Parth, this is the source of your other build failure I believe. I wrote out a few more examples in a quick console application, which may be helpful to understand which type is correct:

[https://gist.github.com/mattdurak/3f54cde77767185d3f1070c6edad7b59](https://gist.github.com/mattdurak/3f54cde77767185d3f1070c6edad7b59)


---
In reply to: [427736421](https://github.com/Azure/azure-c-util/pull/43#discussion_r427736421) [](ancestors = 427736421,427736331)

tests/interlocked_win32_ut/mock_interlocked.c
mattdurak: I don't think this would compile without these...the .c file included below will reference the functions so it needs the declaration (we do the same in the other "mock" versions of source files where we are mocking things from windows.h)

---
In reply to: [427739869](https://github.com/Azure/azure-c-util/pull/43#discussion_r427739869) [](ancestors = 427739869)

tests/interlocked_int/interlocked_int.c
parth21999: I tried creating the array on the stack and got this error: Unhandled exception thrown: read access violation. **destination** was 0xFFFFFFFFFFFFFFFF. occurred

---
In reply to: [427722665](https://github.com/Azure/azure-c-util/pull/43#discussion_r427722665) [](ancestors = 427722665,427722625)

tests/interlocked_int/interlocked_int.c
parth21999: Matt suggested using unsigned integers because when I tried using signed integers, I wasn't able to put n bits in an n-bit signed integer. So I declare them as unsigned and then typecast to signed when passing to the function. I can change the type in the ASSERT to make it clear that I am comparing two signed values.

---
In reply to: [427738515](https://github.com/Azure/azure-c-util/pull/43#discussion_r427738515) [](ancestors = 427738515,427738341)

mattdurak: Parth, I think you just want to change it to:

```
ASSERT_ARE_EQUAL(uint32_t, 0xFFFFFFFF, destination);
```

Right now you are using a signed value and saying the type is unsigned.

---
In reply to: [428355088](https://github.com/Azure/azure-c-util/pull/43#discussion_r428355088) [](ancestors = 428355088,427738515,427738341)

parth21999:
> #endif

Do you mean this one? [https://docs.microsoft.com/en-us/windows/win32/api/winnt/nf-winnt-_inlineinterlockedadd64](https://docs.microsoft.com/en-us/windows/win32/api/winnt/nf-winnt-_inlineinterlockedadd64)

---
In reply to: [631237098](https://github.com/Azure/azure-c-util/pull/43#issuecomment-631237098) [](ancestors = 631237098)

---
Refers to: inc/azure_c_util/interlocked.h:58 in 6a604e4. [](commit_id = 6a604e42b324f9dd0f03955cee77e1281451a409, deletion_comment = False)

mattdurak:
> #endif

Weird, MSDN doc's link to that Inline named one, but the actaul win32 header shows InterlockedAdd64 does in fact exist:

winnt.h line 2916
#define InterlockedAdd64 _InlineInterlockedAdd64

Looks like an MSDN bug, might be nice to have all of the InterlockedAddN functions in our PAL

---
In reply to: [632160751](https://github.com/Azure/azure-c-util/pull/43#issuecomment-632160751) [](ancestors = 632160751,631237098)

---
Refers to: inc/azure_c_util/interlocked.h:58 in 6a604e4. [](commit_id = 6a604e42b324f9dd0f03955cee77e1281451a409, deletion_comment = False)

tests/interlocked_win32_ut/interlocked_win32_ut.c
parth21999: If I don't use the cast, I get this error: Warning C4057 'function': 'volatile LONG *' differs in indirection to slightly different base types from 'volatile int32_t *'.

---
In reply to: [427740898](https://github.com/Azure/azure-c-util/pull/43#discussion_r427740898) [](ancestors = 427740898)

tests/interlocked_win32_ut/interlocked_win32_ut.c
parth21999: Yes, that is a negative value.

---
In reply to: [427743632](https://github.com/Azure/azure-c-util/pull/43#discussion_r427743632) [](ancestors = 427743632,427743537)

tests/interlocked_win32_ut/interlocked_win32_ut.c
parth21999: If I declare it as destination[2] I must provide a brace enclosed initializer list. Creating the array on the stack gives an alignment error in InterlockedCompareExchange128.

---
In reply to: [427743989](https://github.com/Azure/azure-c-util/pull/43#discussion_r427743989) [](ancestors = 427743989)

tests/interlocked_win32_ut/interlocked_win32_ut.c
parth21999: I didn't add it in the unit test because that is checked in the integration test.

---
In reply to: [427744286](https://github.com/Azure/azure-c-util/pull/43#discussion_r427744286) [](ancestors = 427744286)

tests/interlocked_win32_ut/interlocked_win32_ut.c
parth21999: I'm doing that in the integration tests, so I didn't add that here. The INTERLOCKED_WIN32 requirements don't say anything about  target.

---
In reply to: [427745007](https://github.com/Azure/azure-c-util/pull/43#discussion_r427745007) [](ancestors = 427745007,427744935)

tests/interlocked_win32_ut/interlocked_win32_ut.c
parth21999: Yeah the 'volatile void**' is wrong.

---
In reply to: [427745365](https://github.com/Azure/azure-c-util/pull/43#discussion_r427745365) [](ancestors = 427745365)

tests/interlocked_win32_ut/mock_interlocked.c
parth21999: The Interlocked functions defined in windows.h are macros, so I have to undef before redefining them to be the mocked versions.

---
In reply to: [427739418](https://github.com/Azure/azure-c-util/pull/43#discussion_r427739418) [](ancestors = 427739418)

tests/interlocked_win32_ut/mock_interlocked.c
parth21999: I changed InterlockedExchange8 to use CHAR

---
In reply to: [427691435](https://github.com/Azure/azure-c-util/pull/43#discussion_r427691435) [](ancestors = 427691435)

tests/interlocked_win32_ut/interlocked_win32_ut.c
parth21999: I test this in the integration tests.

---
In reply to: [427744374](https://github.com/Azure/azure-c-util/pull/43#discussion_r427744374) [](ancestors = 427744374)

tests/interlocked_win32_ut/interlocked_win32_ut.c
```diff
@@ -0,0 +1,731 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#ifdef __cplusplus
+#include <cstdlib>
+#include <cstddef>
+#include <cstdbool>
```
mattdurak, line 7: this one isn't needed (bool is an intrinsic type in c++)

tests/interlocked_int/interlocked_int.c
```diff
@@ -0,0 +1,949 @@
...
+{
+    ///arrange
+    volatile int32_t addend = INT32_MAX;
+    int32_t value = INT32_MIN;
+
+    ///act
+    int32_t return_val = interlocked_add(&addend, value);
+
+    ///assert
+    ASSERT_ARE_EQUAL(int32_t, -1, interlocked_or((volatile int32_t*)&addend, 0));
```
mattdurak, line 64: >(volatile int32_t*) [](start = 49, length = 19)

Unnecessary cast.

Also, interlocked_add is more of a convention for reading (obviously not available for 16/8-bit, but should use interlocked_add/interlocked_add_64 where possible). /consider

mattdurak: I believe you can remove the casts here and everywhere in this file

---
In reply to: [428944893](https://github.com/Azure/azure-c-util/pull/43#discussion_r428944893) [](ancestors = 428944893)

mattdurak:
> MOCKABLE_FUNCTION(, int32_t, interlocked_add, volatile int32_t*, addend, int32_t, value);

Since InterlockedAdd64 exists, it would be really nice to include it as well, as Andrie said. Ok to make a separate task to add it in.

---
Refers to: inc/azure_c_util/interlocked.h:23 in 918c9b4. [](commit_id = 918c9b4f100a8650b5443dcba46eeddde1a39634, deletion_comment = False)

tests/interlocked_win32_ut/interlocked_win32_ut.c
```diff
@@ -0,0 +1,700 @@
...
+    int32_t return_val = interlocked_compare_exchange(&destination, exchange, comperand);
+
+    ///assert
+    ASSERT_ARE_EQUAL(char_ptr, umock_c_get_expected_calls(), umock_c_get_actual_calls());
+    ASSERT_ARE_EQUAL(int32_t, INT32_MAX, return_val);
+}
+
+/*Tests_SRS_INTERLOCKED_WIN32_43_013: [interlocked_compare_exchange_128 shall call InterlockedCompareExchange128 from windows.h.]*/
+/*Tests_SRS_INTERLOCKED_WIN32_43_014 : [interlocked_compare_exchange_128 shall return true if* comperand_result equals the original value of* destination.]*/
+TEST_FUNCTION(interlocked_compare_exchange_128_calls_InterlockedCompareExchange128_true_case)
```
mattdurak, line 200: >interlocked_compare_exchange_128_calls_InterlockedCompareExchange128_true_case [](start = 14, length = 78)

all of this needs `#ifdef _WIN64`. Same for int test

tests/interlocked_int/interlocked_int.c
parth21999: I think it would be better to do all the reads with interlocked_or so that things are consistent through the file. 

---
In reply to: [428944893](https://github.com/Azure/azure-c-util/pull/43#discussion_r428944893) [](ancestors = 428944893)

parth21999: This cast can be removed, but in the places where a pointer to an unsigned integer is being passed, a cast must be used.

---
In reply to: [428945102](https://github.com/Azure/azure-c-util/pull/43#discussion_r428945102) [](ancestors = 428945102,428944893)

adapters/interlocked_win32.c
```diff
@@ -1,4 +1,330 @@
...
-#include "azure_c_util/interlocked.h"
\ No newline at end of file
+IMPLEMENT_MOCKABLE_FUNCTION(, int32_t, interlocked_add, volatile int32_t*, addend, int32_t, value) 
+{
+    /*Codes_SRS_INTERLOCKED_43_001 [ interlocked_add shall atomically add *addend with value and store the result in *addend.]*/
+    /*Codes_SRS_INTERLOCKED_43_032: [ interlocked_add shall return the result of the addition.]*/
+    /*Codes_SRS_INTERLOCKED_WIN32_43_001: [ interlocked_add shall call InterlockedAdd from windows.h. ]*/
+    /*Codes_SRS_INTERLOCKED_WIN32_43_002: [ interlocked_add shall return the result of the addition. ]*/
+
+    return InterlockedAdd((volatile LONG*)addend, value);
```
anporumb, line 14: >(volatile LONG*) [](start = 26, length = 16)

why is the cast necessary?

anporumb: Here and everywhere. Point being: either the cast "does something" (changes value / signedness etc) (in which case it is wrong), or it doesn't do anything, in which case it is extraneous. 

There are legitimate cases where casts are necessary. Is this necessary and if so... why... I asked this in iteration 6 too - looking back at the comments,

---
In reply to: [430678730](https://github.com/Azure/azure-c-util/pull/43#discussion_r430678730) [](ancestors = 430678730)

tests/interlocked_int/interlocked_int.c
```diff
@@ -0,0 +1,999 @@
...
+{
+    TEST_MUTEX_RELEASE(g_testByTest);
+}
+
+/*Tests_SRS_INTERLOCKED_43_001: [ interlocked_add shall atomically add 32-bit integers *addend and value and store the result in *addend.]*/
+/*Tests_SRS_INTERLOCKED_43_032: [interlocked_add shall return the result of the addition.]*/
+TEST_FUNCTION(interlocked_add_does_addition)
+{
+    ///arrange
+    volatile int32_t addend = INT32_MAX;
```
anporumb, line 57: >volatile int32_t addend = INT32_MAX; [](start = 4, length = 36)

as a matter of code engineering, we should initialize interlocked values with interlocked APIs. 

In this specific case, as it is only 1 thread executing, it works "as is".

Consider though initializing the values with InterlockedExchange(...).

Here and everywhere.

tests/interlocked_int/interlocked_int.c
```diff
@@ -0,0 +1,999 @@
...
+    int32_t return_val = interlocked_add(&addend, value);
+
+    ///assert
+    ASSERT_ARE_EQUAL(int32_t, INT32_MIN, interlocked_or(&addend, 0));
+    ASSERT_ARE_EQUAL(int32_t, INT32_MIN, return_val);
+}
+
+/*Tests_SRS_INTERLOCKED_43_001: [ interlocked_add shall atomically add 32-bit integers *addend and value and store the result in *addend.]*/
+/*Tests_SRS_INTERLOCKED_43_032: [interlocked_add shall return the result of the addition.]*/
+TEST_FUNCTION(interlocked_add_overflows_lower_bound)
```
anporumb, line 86: >overflows [](start = 30, length = 9)

it "underflows"... 

Here and everywhere.

inc/azure_c_util/interlocked.h
```diff
@@ -29,7 +30,7 @@ MOCKABLE_FUNCTION(, int32_t, interlocked_compare_exchange, volatile int32_t*, de
 MOCKABLE_FUNCTION(, bool, interlocked_compare_exchange_128, volatile int64_t*, destination, int64_t, exchange_high, int64_t, exchange_low, int64_t*, comperand_result);
```
anporumb, line 30: should be protetced just like its implementation, otherwise it is only detectable at link time,

missing #ifdef _WIN64

tests/interlocked_int/interlocked_int.c
```diff
@@ -0,0 +1,999 @@
...
+    int16_t return_val = interlocked_compare_exchange_16(&destination, exchange, comperand);
+
+    ///assert
+    ASSERT_ARE_EQUAL(int16_t, INT16_MIN, interlocked_or_16(&destination, 0));
+    ASSERT_ARE_EQUAL(int16_t, INT16_MAX, return_val);
+}
+
+/*Tests_SRS_INTERLOCKED_43_009: [interlocked_compare_exchange_16 shall compare the 16-bit integers pointed to by destination and comperand. If they are equal, *destination is set to exchange. These operations are performed atomically.]*/
+/*Tests_SRS_INTERLOCKED_43_040: [ interlocked_compare_exchange_16 shall return the initial value of *destination. ]*/
+TEST_FUNCTION(interlocked_compare_exchange_16_not_exchanges_when_equal)
```
anporumb, line 332: >when_equal [](start = 60, length = 10)

when_not_equal... ?

tests/interlocked_win32_ut/mock_interlocked.c
```diff
@@ -0,0 +1,72 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#include "windows.h"
+#include "mock_interlocked.h"
+
```
anporumb, line 6: why is this file needed? what happens if it doesn't exist?

tests/interlocked_win32_ut/mock_interlocked.h
```diff
@@ -0,0 +1,38 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#include "windows.h"
+#include "umock_c/umock_c_prod.h"
+
+MOCKABLE_FUNCTION(, LONG, mock_InterlockedAdd, volatile LONG*, addend, LONG, value);
```
anporumb, line 7: >MOCKABLE_FUNCTION [](start = 0, length = 17)

missing __cplusplus thingie, extern "C"{...

anporumb: how did this even compile for the .dll tests... 

---
In reply to: [430698572](https://github.com/Azure/azure-c-util/pull/43#discussion_r430698572) [](ancestors = 430698572)

tests/interlocked_win32_ut/mock_interlocked.h
```diff
@@ -0,0 +1,38 @@
...
+MOCKABLE_FUNCTION(, SHORT, mock_InterlockedIncrement16, volatile SHORT*, addend);
+MOCKABLE_FUNCTION(, LONG64, mock_InterlockedIncrement64, volatile LONG64*, addend);
+MOCKABLE_FUNCTION(, LONG, mock_InterlockedOr, volatile LONG*, destination, LONG, value);
+MOCKABLE_FUNCTION(, SHORT, mock_InterlockedOr16, volatile SHORT*, destination, SHORT, value);
+MOCKABLE_FUNCTION(, LONG64, mock_InterlockedOr64, volatile LONG64*, destination, LONG64, value);
+MOCKABLE_FUNCTION(, char, mock_InterlockedOr8, volatile char*, destination, char, value);
+MOCKABLE_FUNCTION(, LONG, mock_InterlockedXor, volatile LONG*, destination, LONG, value);
+MOCKABLE_FUNCTION(, SHORT, mock_InterlockedXor16, volatile SHORT*, destination, SHORT, value);
+MOCKABLE_FUNCTION(, LONG64, mock_InterlockedXor64, volatile LONG64*, destination, LONG64, value);
+MOCKABLE_FUNCTION(, char, mock_InterlockedXor8, volatile char*, destination, char, value);
```
anporumb, line 38: >); [](start = 88, length = 2)

file doesn't end with a empty line.

anporumb, line 38: >char [](start = 77, length = 4)

will wait for the next iteration after 13.

adapters/interlocked_win32.c
mattdurak: int32_t is typedefed to int (on win32 at least). LONG is long.
There is definitely a warning assigning between long and int even though they are 32-bit signed integers on Windows. I think that's why these casts remain.

---
In reply to: [430679721](https://github.com/Azure/azure-c-util/pull/43#discussion_r430679721) [](ancestors = 430679721,430678730)

tests/interlocked_int/interlocked_int.c
parth21999: Yeah that was a find-replace artifact. Fixed it.

---
In reply to: [430689511](https://github.com/Azure/azure-c-util/pull/43#discussion_r430689511) [](ancestors = 430689511)

tests/interlocked_int/interlocked_int.c
```diff
@@ -0,0 +1,1055 @@
...
+    TEST_MUTEX_RELEASE(g_testByTest);
+}
+
+/*Tests_SRS_INTERLOCKED_43_001: [ interlocked_add shall atomically add 32-bit integers *addend and value and store the result in *addend.]*/
+/*Tests_SRS_INTERLOCKED_43_032: [interlocked_add shall return the result of the addition.]*/
+TEST_FUNCTION(interlocked_add_does_addition)
+{
+    ///arrange
+    volatile int32_t addend;
+	interlocked_exchange(&addend, INT32_MAX);
```
anporumb, line 58: >	 [](start = 0, length = 1)

evil TAB character detected

tests/interlocked_win32_ut/interlocked_win32_ut.c
```diff
@@ -0,0 +1,751 @@
...
+
+static void* my_gballoc_malloc(size_t size)
+{
+    return malloc(size);
+}
+
+static void my_gballoc_free(void* ptr)
+{
+    free(ptr);
+}
```
dcristoloveanu, line 24: These are not needed

tests/interlocked_int/interlocked_int.c
```diff
@@ -0,0 +1,1055 @@
...
+    int32_t value = INT32_MIN;
+
+    ///act
+    int32_t return_val = interlocked_add(&addend, value);
+
+    ///assert
+    ASSERT_ARE_EQUAL(int32_t, -1, interlocked_or(&addend, 0));
```
dcristoloveanu, line 65: > 0)); [](start = 57, length = 5)

Consider adding some information of what failed in the assert (ASSERT_ARE_EQUAL allows you to say ASSERT_ARE_EQUAL(int32_t, -1, interlocked_or(&addend, 0), "addend was not modified");
This makes sense if there are multiple asserts in a test and they are of the same type or same style. Having some extra bit of text (besides the expected value, actual value and line no makes reading the result of the test a bit easier).
Not mandatory, but a consider.

PR #44, opened 2020-05-27
Interlocked linux requirements

parth21999:


devdoc/interlocked_linux_requirements.md
```diff
@@ -0,0 +1,345 @@
+interlocked linux
+================
+
+## Overview
+
+`interlocked linux` is the Linux-specific implementation of the `interlocked` module. Each function calls the corresponding `atomic` function from `stdatomic.h`.
```
mattdurak, line 6: >`atomic` function from `stdatomic.h`. [](start = 124, length = 37)

Should add a note that this is always using the memory order `memory_order_seq_cst` (I believe that's the default for all of the calls when you don't call the _explicit version and specify).

This should be done to match the Interlocked win32 api which takes a full memory barrier for each operation.

devdoc/interlocked_linux_requirements.md
```diff
@@ -0,0 +1,345 @@
...
+
+## Overview
+
+`interlocked linux` is the Linux-specific implementation of the `interlocked` module. Each function calls the corresponding `atomic` function from `stdatomic.h`.
+
+
+## Exposed API
+
+```c
+MOCKABLE_FUNCTION(, int32_t, interlocked_add, volatile int32_t*, addend, int32_t, value);
```
mattdurak, line 12: >volatile [](start = 46, length = 8)

I think all of the types for the stdatomic functions should have the `_Atomic` qualifier. That's not available on win32 since we don't have C11 atomics. It's possible this means we should have some kind of atomic thing defined to make this correct and cross-platform.

Something like:

```
#ifdef WIN32
#define volatile_atomic volatile
#else
#define volatile_atomic volatile _Atomic
#endif
```

devdoc/interlocked_linux_requirements.md
```diff
@@ -0,0 +1,345 @@
...
+```
+
+## interlocked_add
+
+```c
+MOCKABLE_FUNCTION(, int32_t, interlocked_add, volatile int32_t*, addend, int32_t, value);
+```
+**SRS_INTERLOCKED_LINUX_43_001: [** `interlocked_add` shall call `atomic_fetch_add` from `stdatomic.h`. **]**
+
+**SRS_INTERLOCKED_LINUX_43_002: [** `interlocked_add` shall return the result of the addition. **]**
```
mattdurak, line 52: >the result of the addition [](start = 67, length = 26)

consider being explicit for documentation purposes, like: "the initial value in `addend` plus `value`."

devdoc/interlocked_linux_requirements.md
```diff
@@ -0,0 +1,345 @@
...
+**SRS_INTERLOCKED_LINUX_43_010: [** `interlocked_and_8` shall return the initial value of `*destination`. **]**
+
+## interlocked_compare_exchange
+
+```c
+MOCKABLE_FUNCTION(, int32_t, interlocked_compare_exchange, volatile int32_t*, destination, int32_t, exchange, int32_t, comperand);
+```
+**SRS_INTERLOCKED_LINUX_43_011: [** `interlocked_compare_exchange` shall call `atomic_compare_exchange_strong` from `stdatomic.h`. **]**
+
+**SRS_INTERLOCKED_LINUX_43_012: [** `interlocked_compare_exchange` shall return the initial value of `*destination`. **]**
```
mattdurak, line 106: >initial value of `*destination` [](start = 84, length = 31)

I also think these need to be explicit (especially because there is some logic to handle here)

If `atomic_compare_exchange_strong` returns `true` then `interlocked_compare_exchange` shall return `comperand`.

If `atomic_compare_exchange_strong` returns `false` then:

 - `interlocked_compare_exchange` shall call `atomic_load_explicit` to get the current value of `destination`.

 - If the loaded value is different than `comparand` then `interlocked_compare_exchange` shall return the loaded value.

Of course this leads to an issue we probably need to address here. What if `atomic_compare_exchange_strong` returns `false` but `atomic_load_explicit` returns `comperand`? You can't return `comperand` if the compare-exchange failed (because that breaks the contract for this interface that we return the value only if we did the exchange). So you either need to:

1. Change this function to return a flag like C11 atomics do (this can then easily be implemented either way), but then callers we convert from InterlockedCompareExchange need to have big code changes.
2. If the load after compare exchange is not consistent, then loop over and repeat this. Not quite as performant in all cases, but should be correct. This means you will need to add more requirements here of course

devdoc/interlocked_linux_requirements.md
```diff
@@ -0,0 +1,345 @@
...
+MOCKABLE_FUNCTION(, int64_t, interlocked_xor_64, volatile int64_t*, destination, int64_t, value);
+MOCKABLE_FUNCTION(, int8_t, interlocked_xor_8, volatile int8_t*, destination, int8_t, value);
+```
+
+## interlocked_add
+
+```c
+MOCKABLE_FUNCTION(, int32_t, interlocked_add, volatile int32_t*, addend, int32_t, value);
+```
+**SRS_INTERLOCKED_LINUX_43_001: [** `interlocked_add` shall call `atomic_fetch_add` from `stdatomic.h`. **]**
```
mattdurak, line 50: >atomic_fetch_add [](start = 66, length = 16)

So for InterlockedX calls, you didn't need this because you were passing all arguments "as is". Since the atomic_X functions have slightly different parameters, we should follow the convention we use in other specs. So this requirement should be like:

`interlocked_add` shall call `atomic_fetch_add` with `addend` as `object` and `value` as `operand`.

Things like interlocked compare exchange use different orders of arguments, so it's nice to be very explicit here.

mattdurak, line 50: >from `stdatomic.h` [](start = 84, length = 18)

Feel free to drop this part "from stdatomic.h", I don't know that it really adds much to having it in each spec.

devdoc/interlocked_linux_requirements.md
```diff
@@ -0,0 +1,345 @@
...
+
+**SRS_INTERLOCKED_LINUX_43_020: [** `interlocked_compare_exchange_pointer` shall return the initial value of `*destination`. **]**
+
+## interlocked_decrement
+
+```c
+MOCKABLE_FUNCTION(, int32_t, interlocked_decrement, volatile int32_t*, addend);
+
+```
+**SRS_INTERLOCKED_LINUX_43_021: [** `interlocked_decrement` shall call `atomic_fetch_sub` from `stdatomic.h`. **]**
```
mattdurak, line 143: >atomic_fetch_sub [](start = 72, length = 16)

Again, this should really say what arguments you are passing (because you should say "1 as `operand`", which is not stated currently in your spec)

devdoc/interlocked_linux_requirements.md
mattdurak: Here and everywhere in this file

---
In reply to: [431449568](https://github.com/Azure/azure-c-util/pull/44#discussion_r431449568) [](ancestors = 431449568)

devdoc/interlocked_linux_requirements.md
parth21999: If 'atomic_compare_exchange_strong' returns false, why do we have to check that the loaded value is different than 'comparand'? Why will it be the same? What should happen if it is the same?


---
In reply to: [431445821](https://github.com/Azure/azure-c-util/pull/44#discussion_r431445821) [](ancestors = 431445821)

mattdurak: Discussed offline:

1. Need to check because we need to to atomic_compare_exchange and atomic_fetch as two separate instructions, so the value can change in between
2. If we don't check and we return `comparand` but atomic_compare_exchange returned false, then we broke our contract (we should only return the same value as `comparand` if the exchange happened.
3. If it is the same, we need to retry (do the atomic_compare_exchange and possibly the atomic_fetch again)

---
In reply to: [431491999](https://github.com/Azure/azure-c-util/pull/44#discussion_r431491999) [](ancestors = 431491999,431445821)

devdoc/interlocked_linux_requirements.md
```diff
@@ -0,0 +1,345 @@
...
+**SRS_INTERLOCKED_LINUX_43_001: [** `interlocked_add` shall call `atomic_fetch_add` from `stdatomic.h`. **]**
+
+**SRS_INTERLOCKED_LINUX_43_002: [** `interlocked_add` shall return the result of the addition. **]**
+
+## interlocked_add_64
+
+```c
+MOCKABLE_FUNCTION(, int64_t, interlocked_add_64, volatile int64_t*, addend, int64_t, value);
+```
+**SRS_INTERLOCKED_LINUX_43_064: [** `interlocked_add_64` shall call `atomic_fetch_add` from `stdatomic.h`. **]**
```
anporumb, line 59: >atomic_fetch_add [](start = 69, length = 16)

how is it that both interlocked_add (that works on 32 bit integers) and interlocked_add_64 (that works on 64 bit integers) call the same function from stdatomic.h?

anporumb: Well - it turns out that they do! If that is not overloading, I don't know what is! How very interesting... 

---
In reply to: [431912405](https://github.com/Azure/azure-c-util/pull/44#discussion_r431912405) [](ancestors = 431912405)

mattdurak: C11 generics look a bit like C++ overloading

---
In reply to: [431933927](https://github.com/Azure/azure-c-util/pull/44#discussion_r431933927) [](ancestors = 431933927,431912405)

devdoc/interlocked_linux_requirements.md
```diff
@@ -0,0 +1,369 @@
...
+**SRS_INTERLOCKED_LINUX_43_010: [** `interlocked_and_8` shall return the initial value of `*destination`. **]**
+
+## interlocked_compare_exchange
+
+```c
+MOCKABLE_FUNCTION(, int32_t, interlocked_compare_exchange, volatile_atomic int32_t*, destination, int32_t, exchange, int32_t, comperand);
+```
+**SRS_INTERLOCKED_LINUX_43_011: [** `interlocked_compare_exchange` shall call `atomic_compare_exchange_strong` with `destination` as `object`, `&comperand` as `expected` and `exchange` as `desired`. **]**
+
+**SRS_INTERLOCKED_LINUX_43_012: [** `interlocked_compare_exchange` shall return a value equal to `comperand` if and only if `atomic_compare_exchange` returns `true`. **]**
```
mattdurak, line 106: >atomic_compare_exchange [](start = 125, length = 23)

Should be `atomic_compare_exchange_strong`

devdoc/interlocked_linux_requirements.md
```diff
@@ -0,0 +1,369 @@
...
+
+**SRS_INTERLOCKED_LINUX_43_066: [** `interlocked_compare_exchange` shall call `atomic_fetch` if `atomic_compare_exchange` returns `false`. **]**
+
+**SRS_INTERLOCKED_LINUX_43_067: [** `interlocked_compare_exchange` shall return the value returned by `atomic_fetch` if the returned value is not equal to `comperand`. **]**
+
+**SRS_INTERLOCKED_LINUX_43_068: [** `interlocked_compare_exchange` shall call `atomic_compare_exchange` if the value returned by `atomic_fetch` is equal to `comperand`. **]**
```
mattdurak, line 112: >call `atomic_compare_exchange` [](start = 73, length = 30)

I'd say something like "shall repeat the compare exchange if the value returned by `atomic_fetch` is equal to `comperand`"

Because you want to do all of the requirements, not just the compare exchange again.

inc/azure_c_util/interlocked.h
```diff
@@ -18,42 +18,46 @@
 extern "C" {
 #endif
 
+#ifdef _WIN32
+#define volatile_atomic volatile
```
mattdurak, line 22: >volatile_atomic [](start = 8, length = 15)

I know I told you to add this and we need something like that. I'm not sure if we should maybe rename it to something else though that has "interlocked" in the name (so it doesn't collide with anything else) or not (because the current name says exactly what it is, which is nice)... Dan or Andrei might have an opinion about this.

But I think it's fine for now as it's a simple "find and replace" rename. You can probably just leave it for now as we do need to declare the types as `_Atomic` to use C11 atomics from what I understand.

PR #45, opened 2020-05-28
fix SM int tests with better time management

anporumb:


tests/sm_int/sm_int.c
```diff
@@ -739,24 +739,38 @@ sm_exec_begin
...
 {
     SM_GO_TO_STATE* goToState = (SM_GO_TO_STATE*)lpThreadParameter;
 
-    LogInfo("set state thread: will now switch state to %" PRI_MU_ENUM "", MU_ENUM_VALUE(SM_STATES, (SM_STATES)(SM_CREATED + goToState->targetState)));
+    LogInfo("switchesTo thread: will now switch state to %" PRI_MU_ENUM "", MU_ENUM_VALUE(SM_STATES, (SM_STATES)(SM_CREATED + goToState->targetState)));
 
+    ASSERT_IS_TRUE(SetEvent(goToState->targetStateAPICalledInNextLine));
     switch (goToState->targetState)
     {
         case 0:/*SM_CREATED*/
```
mattdurak, line 776: >0 [](start = 13, length = 1)

nit: why not actual enum values like below?

mattdurak:
> static DWORD WINAPI switchesToCreated(

poetry: This name confused me, maybe it's `switchesFrom` to match the naming below that starts the thread?

Could also name these two threads something like "goToState" and "unblockStateTransition" or something

---
Refers to: tests/sm_int/sm_int.c:850 in af8c643. [](commit_id = af8c643a9afac33d6a2c5c353bb7beff1ce80c09, deletion_comment = False)

tests/sm_int/sm_int.c
```diff
@@ -1001,12 +1036,17 @@ TEST_FUNCTION(STATE_and_API)
                 }
             }
 
+            LogInfo("went to state =%" PRI_MU_ENUM "; and called =%" PRI_MU_ENUM " cleaner thread might already have run", MU_ENUM_VALUE(SM_STATES, (SM_STATES)(i + SM_CREATED)), MU_ENUM_VALUE(SM_APIS, (SM_APIS)(j + SM_OPEN_BEGIN)));
```
mattdurak, line 1039: >cleaner [](start = 83, length = 7)

poetry: or maybe the switchesToCreated is the cleaner thread?

tests/sm_int/sm_int.c
anporumb: reason is poor. Our enums start at '1' with the user-specified values (because _INVALID is "0"). So these numbers have been "indexes" corresponding to enum values. But having the enum values spelled out "as they are" is much better. Changed. 

---
In reply to: [432047519](https://github.com/Azure/azure-c-util/pull/45#discussion_r432047519) [](ancestors = 432047519)

tests/sm_int/sm_int.c
anporumb: Absolutely corrected.

---
In reply to: [432052699](https://github.com/Azure/azure-c-util/pull/45#discussion_r432052699) [](ancestors = 432052699)

anporumb:
> static DWORD WINAPI switchesToCreated(

Did renames, to be more consistent across names(hopefully!)

---
In reply to: [635531784](https://github.com/Azure/azure-c-util/pull/45#issuecomment-635531784) [](ancestors = 635531784)

---
Refers to: tests/sm_int/sm_int.c:850 in af8c643. [](commit_id = af8c643a9afac33d6a2c5c353bb7beff1ce80c09, deletion_comment = False)

PR #46, opened 2020-06-02
Interlocked linux tests and implementation

parth21999:


adapters/interlocked_linux.c
```diff
@@ -1,4 +1,296 @@
...
+            /*Codes_SRS_INTERLOCKED_LINUX_43_066: [ interlocked_compare_exchange shall call atomic_load if atomic_compare_exchange_strong returns false. ]*/
+            int32_t loaded_value = atomic_load(destination);
+            if(loaded_value != comperand)
+            {
+                /*Codes_SRS_INTERLOCKED_LINUX_43_067: [ interlocked_compare_exchange shall return the value returned by atomic_load if the returned value is not equal to comperand. ]*/
+                return loaded_value;
+            }
+        }
+        /*Codes_SRS_INTERLOCKED_LINUX_43_068: [ interlocked_compare_exchange shall repeat the call to atomic_compare_exchange_strong if the value returned by atomic_load is equal to comperand. ]*/
+    }
```
mattdurak, line 75: Sorry, we didn't get a chance to discuss this yesterday, but after you mentioned that aces modifies the second arg ("comparand"), I said we actually can change this.

Spec says:
"Atomically, compares the value pointed to by object for equality
with that in expected, and if true, replaces the value pointed to by object with
desired, and *if false, updates the value in expected with the value pointed to by
object.* "

So really we just need to return comparand_copy always, much simpler :)

I think all we need is this (obviously, you can remove most of the specs too to simplify)

```
int32_t comperand_copy = comperand;
(void)atomic_compare_exchange_strong(destination, &comperand_copy, exchange);
return comparand_copy;
```

comparand_copy is the same as comparand if atomic_compare_exchange returns true.

comparand_copy is the original/current value in destination if atomic_compare_exchange returns false.

adapters/interlocked_linux.c
```diff
@@ -1,4 +1,296 @@
...
+{
+    /*Codes_SRS_INTERLOCKED_LINUX_43_031: [ interlocked_exchange_64 shall call atomic_exchange with target as object and value as desired. ]*/
+    /*Codes_SRS_INTERLOCKED_LINUX_43_032: [ interlocked_exchange_64 shall return the initial value pointed to by target. ]*/
+    return atomic_exchange(target, value);
+}
+
+IMPLEMENT_MOCKABLE_FUNCTION(, int8_t, interlocked_exchange_8, volatile_atomic int8_t*, target, int8_t, value)
+{
+    /*Codes_SRS_INTERLOCKED_LINUX_43_033: [ interlocked_exchange_8 shall call atomic_exchange with target as object and value as desired. ]*/
+	/*Codes_SRS_INTERLOCKED_LINUX_43_034: [ interlocked_exchange_8 shall return the initial value pointed to by target. ]*/
```
mattdurak, line 196: >	 [](start = 0, length = 1)

rogue tab detected :)

adapters/interlocked_linux.c
parth21999: Since comperand is passed by value to interlocked_compare_exchange, is it necessary to make a copy of it?

---
In reply to: [433989674](https://github.com/Azure/azure-c-util/pull/46#discussion_r433989674) [](ancestors = 433989674)

mattdurak: It's not necessary but I think it's cleaner to avoid modifying arguments (IMO all func args should be const, but declaring them as const is overly verbose).

Making a copy doesn't hurt as the compiler will optimize it away anyway.

---
In reply to: [433992371](https://github.com/Azure/azure-c-util/pull/46#discussion_r433992371) [](ancestors = 433992371,433989674)

tests/interlocked_linux_ut/mock_interlocked.c
```diff
@@ -0,0 +1,79 @@
...
+#include <stdatomic.h>
+#include "mock_interlocked.h"
+
+#undef atomic_fetch_add
+#define atomic_fetch_add mock_atomic_fetch_add
+#define mock_atomic_fetch_add(X, Y) _Generic((X), \
+    volatile_atomic int16_t*: mock_atomic_fetch_add_16, \
+    volatile_atomic int32_t*: mock_atomic_fetch_add_32, \
+    volatile_atomic int64_t*: mock_atomic_fetch_add_64 \
+)(X, Y)
```
mattdurak, line 13: Yay for getting C generics working for this mocking!

But this reminds me that *maybe* we should make sure this library works without C11 on Linux.

I think that the cleanest way is to just remove all of the support for this in the CMakeLists.txt in that case. Adding a comment there...

tests/CMakeLists.txt
```diff
@@ -31,13 +31,15 @@ if(${run_unittests})
         add_subdirectory(timer_win32_ut)
         add_subdirectory(interlocked_hl_win32_ut)
         add_subdirectory(interlocked_win32_ut)
+    else()
+        add_subdirectory(interlocked_linux_ut)
```
mattdurak, line 35: elseif...

Needs to check for C11 support at least I think (that should be sufficient, though it's possible for a compiler to support C11 without atomics, we can deal with that if it were ever necessary).

[https://cmake.org/cmake/help/latest/prop_gbl/CMAKE_C_KNOWN_FEATURES.html#prop_gbl:CMAKE_C_KNOWN_FEATURES](https://cmake.org/cmake/help/latest/prop_gbl/CMAKE_C_KNOWN_FEATURES.html#prop_gbl:CMAKE_C_KNOWN_FEATURES)

I think you can just check it by seeing if `c_std_11` is supported in CMAKE_C_COMPILE_FEATURES.

I think the syntax is:

`elseif (c_std_11 IN_LIST CMAKE_C_COMPILE_FEATURES)`

tests/CMakeLists.txt
```diff
@@ -31,13 +31,15 @@ if(${run_unittests})
...
     endif()
 endif()
 
 if(${run_int_tests})
+    add_subdirectory(interlocked_int)
```
mattdurak, line 40: >add_subdirectory(interlocked_int) [](start = 4, length = 33)

Then this is also if c11 (same check as above) OR WIN32

tests/interlocked_linux_ut/interlocked_linux_ut.c
```diff
@@ -0,0 +1,833 @@
...
+#define ENABLE_MOCKS
+#include "mock_interlocked.h"
+#undef ENABLE_MOCKS
+
+static int16_t expected_atomic_compare_exchange_16_value;
+static int counter_16;
+static bool hook_mock_atomic_compare_exchange_16(volatile_atomic int16_t* object, int16_t* expected, int16_t desired)
+{
+    ASSERT_ARE_EQUAL(int16_t, expected_atomic_compare_exchange_16_value, *expected);
+    return counter_16--==0;
```
mattdurak, line 41: >== [](start = 23, length = 2)

let's put spaces around the "==" just to make this more readable.

tests/interlocked_linux_ut/interlocked_linux_ut.c
```diff
@@ -0,0 +1,833 @@
...
+
+    ///act
+    uint8_t return_val = (uint8_t)interlocked_or_8((volatile_atomic int8_t*)&destination, value);
+
+    ///assert
+    ASSERT_ARE_EQUAL(char_ptr, umock_c_get_expected_calls(), umock_c_get_actual_calls(), "Actual calls differ from expected calls");
+    ASSERT_ARE_EQUAL(uint8_t, 0xF0, return_val, "Return value is incorrect.");
+}
+
+// TEST_FUNCTION(interlocked_xor_calls_atomic_fetch_xor)
```
mattdurak, line 766: >//  [](start = 0, length = 3)

accidentally commented out still from testing

tests/interlocked_linux_ut/interlocked_linux_ut.c
```diff
@@ -0,0 +1,833 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#ifdef __cplusplus
+#include <cstdlib>
+#include <cstddef>
```
mattdurak, line 6: should be replaced by cstdint/stdint.c I think (we use stddef for size_t but it's already included by stdint, which has all the intX_t types).

tests/interlocked_linux_ut/interlocked_linux_ut.c
```diff
@@ -0,0 +1,710 @@
...
+#include "mock_interlocked.h"
+#undef ENABLE_MOCKS
+
+static int16_t expected_atomic_compare_exchange_16_value;
+static bool hook_mock_atomic_compare_exchange_16(volatile_atomic int16_t* object, int16_t* expected, int16_t desired)
+{
+    ASSERT_ARE_EQUAL(int16_t, expected_atomic_compare_exchange_16_value, *expected);
+    int16_t original_expected = *expected;
+    *expected = *object;
+    return original_expected == *object;
```
mattdurak, line 42: Actually, maybe this is easier to just implement as a "real" here for the compare exchange mocks.

Leave the ASSERT, then replace these three lines with `return atomic_compare_exchange(object, expected, desired);`

You don't rename the real functions in the test, so this will call the real atomic functions without mocks, but it's called within a mock call for all the nice call recording/mocking functionality.

build/devops_gated.yml
```diff
@@ -94,7 +94,7 @@ jobs:
     - vstest
 
   steps:
-  - task: BatchScript@1
+  - task: BatchScript@1z
```
mattdurak, line 97: >z [](start = 23, length = 1)

seems like a random z was accidentally inserted here

parth21999: whoops

mattdurak: :)

---
In reply to: [434183719](https://github.com/Azure/azure-c-util/pull/46#discussion_r434183719) [](ancestors = 434183719)

PR #49, opened 2020-07-07
Add rc_string specs and header

dcristoloveanu:
Add rc_string specs and header to be used in bsi_op_v2.


devdoc/rc_string_requirements.md
```diff
@@ -0,0 +1,39 @@
...
+
+```c
+    typedef struct RC_STRING_TAG
+    {
+        const char* string;
+    } RC_STRING;
+
+    THANDLE_TYPE_DECLARE(RC_STRING);
+
+MOCKABLE_FUNCTION(, THANDLE(RC_STRING), rc_string_create, const char*, string);
```
mattdurak, line 18: >const char*, string); [](start = 58, length = 21)

Couldn't a second "move create" function also be useful here to avoid copying in some cases? (Assumes we just care about "move create" which was allocated with "malloc" and will call "free" rather than a custom free function).

inc/azure_c_util/rc_string.h
```diff
@@ -0,0 +1,34 @@
...
+extern "C"
+{
+#endif
+
+    typedef struct RC_STRING_TAG
+    {
+        const char* string;
+    } RC_STRING;
+
+    THANDLE_TYPE_DECLARE(RC_STRING);
```
mattdurak, line 26: nit: seems like it shouldn't be indented if line 28 isn't.

devdoc/rc_string_requirements.md
dcristoloveanu: I think we'll end up having all of them (with move and with custom free). I want to start with this and use it and then we can add the other 2 versions as needed to be used. For example I can think of having the with custom free quite useful for baking the strings directly in a const buffer.

---
In reply to: [451153205](https://github.com/Azure/azure-c-util/pull/49#discussion_r451153205) [](ancestors = 451153205)

inc/azure_c_util/rc_string.h
dcristoloveanu: I indented all of them :-)

---
In reply to: [451153708](https://github.com/Azure/azure-c-util/pull/49#discussion_r451153708) [](ancestors = 451153708)

PR #51, opened 2020-07-24
bring in latest pal

anporumb:


inc/azure_c_util/containing_record.h
```diff
@@ -16,6 +16,12 @@
 // Calculate the address of the base of the structure given its type, and an
 // address of a field within the structure.
 //
-#define containingRecord(address, type, field) ((type *)((uintptr_t)(address) - offsetof(type,field)))
+
+#ifdef WIN32
+/*windows.h brings CONTAINING_RECORD*/
+#include "windows.h"
```
dcristoloveanu, line 22: Any benefit from using in one case one containing record and in another case the other? Looks to me that this is simple enough so that we could just use one (our own).

inc/azure_c_util/thandle.h
```diff
@@ -106,7 +106,7 @@ static T* THANDLE_MALLOC(T)(void(*dispose)(T*))
     {                                                                                                                                                               \
         /*Codes_SRS_THANDLE_02_014: [ THANDLE_MALLOC shall initialize the reference count to 1, store dispose and return a T* . ]*/                                 \
         handle_impl->dispose = dispose;                                                                                                                             \
-        INIT_REF_VAR(handle_impl->refCount);                                                                                                                        \
```
dcristoloveanu, line 109: >T_REF_VAR(handle_impl- [](start = 11, length = 22)

Nice.

src/constbuffer_array.c
```diff
@@ -433,7 +433,7 @@ IMPLEMENT_MOCKABLE_FUNCTION(, void, constbuffer_array_dec_ref, CONSTBUFFER_ARRAY
     else
     {
         /* Codes_SRS_CONSTBUFFER_ARRAY_01_016: [ Otherwise constbuffer_array_dec_ref shall decrement the reference count for constbuffer_array_handle. ]*/
-        if (DEC_REF(CONSTBUFFER_ARRAY_HANDLE_DATA, constbuffer_array_handle) == DEC_RETURN_ZERO)
+        if (DEC_REF(CONSTBUFFER_ARRAY_HANDLE_DATA, constbuffer_array_handle) == 0)
```
dcristoloveanu, line 436: >0 [](start = 80, length = 1)

Seems like if we're still using DEC_REF here, we should keep using DEC_RETURN_ZERO, right? IIRC that was defined in REF_COUNT so that ref count owns which is the actual value when the reference count is 0. Just looks a bit odd to change it only here in one place.

anporumb: Your memories are accurate, however, recently refcount_os went away. And DEC_REF is now interlocked_xxx based. And interlocked_decrement returns 0. And DEC_RETURN_ZERO doesn't even exist anymore. (we had DEC_RETURN_ZERO for OS reasons, C11 whatever).

But Parth's work enabled us to stop caring about OS in interlocked_xxx so why not use interlocked_xxx for refcounting? :) [and I need it on a separate track with gballoc that will be revealed in a future CR]

---
In reply to: [460211383](https://github.com/Azure/azure-c-util/pull/51#discussion_r460211383) [](ancestors = 460211383)

inc/azure_c_util/containing_record.h
anporumb: Absolutely no benefit. That's why I switched to having just 1 CONTAINING_RECORD with the casing from windows.h across the repos. If we're in Windows land - use that. If we are in linux land - use the #define. But we spend 99.9999% of our time in Windows land - so let's optimize the common case.

There's a problem with "our own casing" (which I find annoying). Pressing F1 on "our own" doesn't open help. Pressing F1 on something that exists however... will bring me the page with the help.

---
In reply to: [460209663](https://github.com/Azure/azure-c-util/pull/51#discussion_r460209663) [](ancestors = 460209663)

PR #71, opened 2020-10-21
Adding constbuffer_array_create_from_offset_and_count

mattdurak:
Helper to create a subset of a constbuffer array using the same underlying memory/not copying the handle array


devdoc/agenttime_requirements.md
```diff
@@ -1,24 +0,0 @@
-AgentTime
```
dcristoloveanu, line 1: haha!

src/constbuffer_array.c
```diff
@@ -129,6 +142,61 @@ IMPLEMENT_MOCKABLE_FUNCTION(, CONSTBUFFER_ARRAY_HANDLE, constbuffer_array_create
     return result;
 }
 
+static void constbuffer_array_offset_and_count_free(void* context)
+{
+    CONSTBUFFER_ARRAY_HANDLE constbuffer_array_handle = context;
+    constbuffer_array_dec_ref(constbuffer_array_handle);
+}
+
+IMPLEMENT_MOCKABLE_FUNCTION(, CONSTBUFFER_ARRAY_HANDLE, constbuffer_array_create_from_offset_and_count, CONSTBUFFER_ARRAY_HANDLE, original, uint32_t, offset, uint32_t, buffer_count)
```
dcristoloveanu, line 151: >offset [](start = 150, length = 6)

When I think of offset I think of offset within a buffer, so I had to twist my mind a bit to realize that it is the start buffer index in the array.
Consider renaming to something like start_buffer_index. If it's not too much.

Otherwise cheers for it!

dcristoloveanu (approved):
:shipit:

src/constbuffer_array.c
mattdurak: Makes sense, index everywhere now

---
In reply to: [509800197](https://github.com/Azure/azure-c-util/pull/71#discussion_r509800197) [](ancestors = 509800197)

PR #77, opened 2020-11-08
Remove mocked windows.h and its usage from interlocked_hl tests

dcristoloveanu:
Remove mocked window.h and its usage from interlocked_hl tests


anporumb (approved):
:shipit:

PR #82, opened 2020-11-20
Fix log critical and terminate to be ... critical

dcristoloveanu:
Fix log critical and terminate to be ... critical


mattdurak (approved):
:shipit:

PR #91, opened 2021-01-05
Add specs for external command helper

mattdurak:
Module handles running another command and piping the output to an array of strings (with logging).


dcristoloveanu (approved):
:shipit:

PR #95, opened 2021-01-20
chore: open up real_sm to linux

jebrando:


tests/reals/CMakeLists.txt
```diff
@@ -3,15 +3,13 @@
...
 cmake_minimum_required(VERSION 2.8.11)
 
-if(MSVC)
 set(REAL_SM_C_FILES
     real_sm.c
 )
 set(REAL_SM_H_FILES
     real_sm.h
     real_sm_renames.h
 )
```
mattdurak, line 12: I think we should just dump these in the main lists below now, no need for the separate variable

mattdurak (approved):
:shipit:

PR #98, opened 2021-01-21
Add includecheck leg

dcristoloveanu:
Add includecheck leg

### DIFF
--- a/src/interlocked_hl.c
+++ b/src/interlocked_hl.c
@@ -4,7 +4,7 @@
 #include <inttypes.h>
 #include <stddef.h>                 // for NULL
 
-#include "umock_c/umock_c_prod.h"   // for MU_COUNT_ARG_0, MU_DISPATCH_EMPTY...
+#include "umock_c/umock_c_prod.h"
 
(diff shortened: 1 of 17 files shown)


src/interlocked_hl.c
```diff
@@ -2,7 +2,9 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 #include <inttypes.h>
-#include <limits.h>
+#include <stddef.h>                 // for NULL
+
+#include "umock_c/umock_c_prod.h"   // for MU_COUNT_ARG_0, MU_DISPATCH_EMPTY...
```
mattdurak, line 7: Is the MU_* stuff actually why we include this? Sounds like macro_utils.h

mattdurak (approved):

src/interlocked_hl.c
dcristoloveanu: It's not. It's for the IMPLEMENT_MOCKABLE. Fixed by removing the "for"

jebrando (approved):
:shipit:

PR #102, opened 2021-01-28
Remove extra logging from THANDLE

mattdurak:
Allow code like `THANDLE_INITIALIZE(RC_STRING)(&var, NULL)` because we use it in a few places so the logging is noisy.


anporumb (approved):
:shipit:

PR #105, opened 2021-02-09
use w4

anporumb:


src/rc_string.c
```diff
@@ -188,19 +188,25 @@ IMPLEMENT_MOCKABLE_FUNCTION(, THANDLE(RC_STRING), rc_string_create_with_custom_f
...
     {
         /*Codes_SRS_RC_STRING_02_002: [ rc_string_recreate shall perform same steps as rc_string_create to return a THANDLE(RC_STRING) with the same content as source. ]*/
         THANDLE(RC_STRING) temp = rc_string_create_impl(self->string);
-        THANDLE_INITIALIZE_MOVE(RC_STRING)(&result, &temp);
-        /*return as is*/
+
+        if (temp == NULL)
+        {
+            LogError("unable to recreate the string %" PRI_RC_STRING " with individual storage", RC_STRING_VALUE(self));
+        }
```
jebrando, line 206: Is there any test changes with this change?

jebrando (approved):
:shipit:

src/rc_string.c
anporumb: No, because this just logs something, and there's no "STRICT_EXPECTED_CALL(I_will_log_something)". The only change is the assignment to "result=NULL" above.

PR #106, opened 2021-02-18
Add sm_fault and faulted state to state machine

mattdurak:
- Fix sm_int chaos test so that end calls are always paired with begin
- Add sm_fault, which disallows new operations except close
- Add retry loops when state changes in barrier and close end so we don't silently fail and get stuck
- Fail fast if sm_exec_end is called too many times and the count would go negative

### DIFF
--- a/devdoc/sm_requirements.md
+++ b/devdoc/sm_requirements.md
@@ -57,7 +57,7 @@ Barriers - since they are exclusive - are realized by switching to a state calle
 
 Close is realized by prohibiting all calls (including competing `sm_close_begin` calls) by setting a bit with `InterlockedOr`. `sm_close_begin` will wait for the state to reach `SM_OPENED` and the number of executing calls to be `0`. This allows an ongoing barrier to finish (and return to `SM_OPENED` state), or the executing APIs to finish.
 
-Fault is realized by prohibiting all calls (except for `sm_close_begin` calls and all `_end` calls) by setting a bit with `InterlockedOr`. `sm_fault` will not set the fault bit if the state is `SM_CREATED` or `SM_CLOSING`. This means that the module cannot fault while it is closed and any currently executing operations can complete.
+Fault is realized by prohibiting all calls (except for `sm_close_begin` calls and all `_end` calls) by setting a bit with `InterlockedOr`. `sm_fault` will not set the fault bit if the state is `SM_CREATED`. This means that the module cannot fault while it is closed and any currently executing operations can complete.
 
 `sm` will verify all sequence of calls. When a _begin call is called in an unexpected state, `sm` will refuse to grant the execution. `sm_exec_end` calls do not have a return value, but `sm` does protect internally against mismatched such calls. For example, `n` is decremented by `sm_exec_end`, but `sm` does not allow `n` to reach negative values.
 
--- a/devdoc/sm_requirements.md
+++ b/devdoc/sm_requirements.md
@@ -9,9 +9,9 @@
 
 2. The module's APIs are callable after _open has executed (_open being the exception - it can only be called after _create).
 
-3. The module may go into a "faulted" state which disallows all calls, other than closing the module (which resets the faulted state). The "faulted" state means that something catastrophic happened so the module must be closed and re-opened. The "faulted" state is not possible when the module is closed.
+3. The module may go into a "faulted" state which disallows all calls, other than closing the module. The "faulted" state means that something catastrophic happened so the module must be closed and destroyed. It is a terminal state and _open will fail.
 
-4. The module has a _close function that reverts the effects of _open and allows calling _open again.
+4. The module has a _close function that reverts the effects of _open and allows calling _open again. Calling _close when the module is "faulted" cleans up the module but leaves it in a "faulted" state.
 
 5. the APIs that can be called in the _open state can be further divided into 3 categories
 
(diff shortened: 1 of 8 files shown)


devdoc/sm_requirements.md
```diff
@@ -274,4 +289,19 @@ MOCKABLE_FUNCTION(, void, sm_barrier_end, SM_HANDLE, sm);
...
-**SRS_SM_02_073: [** `sm_barrier_end` shall switch the state to `SM_OPENED`. **]**
\ No newline at end of file
+**SRS_SM_02_073: [** `sm_barrier_end` shall switch the state to `SM_OPENED`. **]**
+
+### sm_fault
+```c
+MOCKABLE_FUNCTION(, void, sm_fault, SM_HANDLE, sm);
+```
+
+`sm_fault` puts `sm` in a faulted state that requires the module to be closed to clear the fault.
```
dcristoloveanu, line 299: >requires the module to be closed to clear the fault. [](start = 45, length = 52)

This statement really made me wondering why we need the fault (why not just call close?)

dcristoloveanu: I've been thinking about this as an evolution of what we have. We do have some intertwined code paths that rely on Faulted today, but my thinking goes like this:

The different between faulted and closed has started to appear when we started switching to FAULTED state to disallow accepting new requests (I think it might have started in BSDL at some point, but don't quote me on it).

FAULTED wants to solve a very specific issue: when a callback arises from inside the guts of the module and we know that the module is in a bad state now, we would like to not accept any requests anymore, so that at least we do not make it worse on ourselfes.

Now the problems:
i) FAULTED does not tell the user that something is wrong. Let's say sp_index is switching to FAULTED, the module on top of it will keep hammering without any knowledge of the FAULTED state (it will get its writes rejected with a MU_FAILURE probably).
ii) FAULTED requires the user to call close to use the module again. So in case sp_index is usable after a close and reopen, why does the user have to do that. It should be that the sp_index module when it switches to FAULTED, should actually close.
And from this comes the next level thinking: if sp_index would simply close itself when it gets internally faulted, then there is no need for a FAULTED state.
Now when we are closing internally because we are FAULTED I think it would be a good idea to let the user of the module know that "sp_index is closed now", potentially with an error code. Which smells like a callback to me.

What do you think? Would this be a problem for the case at hand that you have?
If we go this route it sounds to me that sm does not need a FAULTED state.

---
In reply to: [578974556](https://github.com/Azure/c-util/pull/106#discussion_r578974556) [](ancestors = 578974556)

dcristoloveanu: I think this is a good moment for a chat about it.

---
In reply to: [578978940](https://github.com/Azure/c-util/pull/106#discussion_r578978940) [](ancestors = 578978940,578974556)

dcristoloveanu (changes requested):
:clock1:

devdoc/sm_requirements.md
anporumb: I want to be in the chat too.

---
In reply to: [578979241](https://github.com/Azure/c-util/pull/106#discussion_r578979241) [](ancestors = 578979241,578978940,578974556)

dcristoloveanu (approved):
:shipit:

devdoc/sm_requirements.md
mattdurak: Capturing some of the discussion/conclusion for "historical reasons"/"future developers":
- Will add the faulted state for now, doesn't necessarily need to be used, but it fits the SF model for some modules (SF expects ReportFault and then it calls BeginClose "later", so self-closing doesn't fit the semantics as "nicely", which may or may not be a good reason to follow its model)
- Should try the "self-close" model for some modules that aren't as tied to the SF model directly, if applicable. This can be done even if sm has a `sm_fault`, modules don't need to use `sm_fault`.
  - Note that self-close likely needs to "allow close while something is executing" (e.g. Thread Local pending operation count to exclude the current thread from the wait) or it will deadlock in many cases.

The callback to notify the user is sort of orthogonal (e.g. it's needed anyway when a callback fails, may not be needed for synchronous failures)

---
In reply to: [579301007](https://github.com/Azure/c-util/pull/106#discussion_r579301007) [](ancestors = 579301007,578979241,578978940,578974556)

dcristoloveanu (changes requested):
:clock1:

devdoc/sm_requirements.md
mattdurak: Pending change: Faulted will be terminal. After closing from faulted, we enter a terminal faulted state and open will no longer work (must be destroyed and re-created)

---
In reply to: [579417557](https://github.com/Azure/c-util/pull/106#discussion_r579417557) [](ancestors = 579417557,579301007,578979241,578978940,578974556)

tests/sm_ut/sm_ut.c
```diff
@@ -531,6 +578,36 @@ TEST_FUNCTION(sm_close_end_switches_state_to_SM_CREATED) /*allows sm_open_begin
     sm_destroy(sm);
 }
 
+TEST_FUNCTION(sm_close_end_switches_state_to_SM_CREATED_leaves_SM_FAULTED_BIT_set)
```
dcristoloveanu, line 581: Let's add the tested specs

dcristoloveanu (approved):
:shipit:

tests/sm_ut/sm_ut.c
mattdurak: Good point, added some specs about calls being allowed when faulted and this not resetting faulted

---
In reply to: [579550722](https://github.com/Azure/c-util/pull/106#discussion_r579550722) [](ancestors = 579550722)

PR #114, opened 2021-04-21
have external command helper reals

anporumb:


tests/reals/real_external_command_helper.c
```diff
@@ -0,0 +1,11 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#include "real_gballoc_hl_renames.h" // IWYU pragma: keep
+#include "real_pipe_renames.h" // IWYU pragma: keep
+#include "real_rc_string_renames.h" // IWYU pragma: keep
+#include "real_rc_string_array_renames.h" // IWYU pragma: keep
+
+#include "real_external_command_helper_renames.h" // IWYU pragma: keep
+
+#include "../../src/external_command_helper.c"
```
mattdurak, line 11: >external_command_helper [](http://example.com/codeflow?start=20&length=23)

Kind of curious about what unit tests would use the real command helper to actually execute an external program...seems a few steps further from things that we normally need reals for (like buffers/interlocked)

mattdurak (approved):
:shipit:

tests/reals/real_external_command_helper.c
anporumb: I need this for install_dkds_event_manifest

PR #115, opened 2021-05-04
Add a THANDLE_TUPLE_ARRAY, which is like rc_string_array …

mattdurak:
…but generalized for multiple different THANDLEs instead of a single THANDLE(RC_STRING). Needed to have an array of pairs (or tuples) of strings


devdoc/thandle_tuple_array_requirements.md
```diff
@@ -0,0 +1,84 @@
...
+### Example Usage
+
+```c
+// In the header
+#define MY_TUPLE_FIELDS \
+    RC_STRING, foo, \
+    RC_STRING, bar, \
+    RC_STRING, baz
+
+DECLARE_THANDLE_TUPLE_ARRAY(MY_TUPLE, MY_TUPLE_FIELDS);
```
anporumb, line 39: >DECLARE_THANDLE_TUPLE_ARRAY [](http://example.com/codeflow?start=0&length=27)

It took me a while to understand what these are. 

I do need to ask why isn't this broken in 2 separate concerns:

Concept 1: have a THANDLE'd struct - this is an independent data type that is like-a-C-struct, except that ALL fields are THANDLEs and the structure itself is a THANDLE.

Example: (codeflow cam compile anything)

DECLARE_THANDLE_STRUCT(FIELDS_S, MY_TUPLE_FIELDS).

DEFINE_THANDLE_STRUCT(FIELDS_S, MY_TUPLE_FIELDS).

... and then have the usual THANDLE macros working on the type (allegedly - somehow).

Concept2: have a THANDLE'd array - this another data type that is-like-C-array and its base type is THANDLE and the complete array is THANDLE. So this new THANDLE array could have its individual elements as either "basic" THANDLEs or structs or other THANDLE_ARRAY. 

... I am curious.

devdoc/thandle_tuple_array_requirements.md
```diff
@@ -0,0 +1,84 @@
...
+
+// In .c file:
+DEFINE_THANDLE_TUPLE_ARRAY(MY_TUPLE, MY_TUPLE_FIELDS)
+
+// Callers
+THANDLE_TUPLE_ARRAY_TYPE(MY_TUPLE) tuple = THANDLE_TUPLE_ARRAY_CREATE(MY_TUPLE)(4);
```
anporumb, line 45: >tuple [](http://example.com/codeflow?start=35&length=5)

is THANDLE_INC_REF/DEC_REF working for tuple here? Would be nice with an example.

anporumb: Truth being said, documentation doesn't say anything about that.

So these arrays are not refcounted, why are they called with THANDLE in their name? I assumed they'd be refcounted (still unsure) by the name only.

devdoc/thandle_tuple_array_requirements.md
mattdurak: hmm, well to give some context, today we have `rc_string_array`. It's just like this but it's a wrapper of `THANDLE(RC_STRING)`'s in an array (so array of thandle + count).

I need to instead pass out an array of 2 strings. So I wanted to generalize something better than `rc_string_array_of_pairs`. I could have made it just be an arbitrary number of strings, but it's easy to use any THANDLE and not limit it to strings.

Now I guess some of this could the generalized further to THANDLE, but it wasn't something I needed in my use case. I'm tempted to just keep it as is to meet my needs because I don't need any additional ref counting really (I just want to have a nice way to pass the array and have a destroy function so the caller doesn't need to do that)

devdoc/thandle_tuple_array_requirements.md
mattdurak: Hmm yeah maybe just `TUPLE_ARRAY` is the analogous name to `rc_string_array`, since the THANDLE's are just used for the fields...
Though it does hold THANDLE's (and maybe it's good to have that called out in the name), it's just not a THANDLE itself...

inc/c_util/thandle_tuple_array.h
```diff
@@ -0,0 +1,147 @@
...
+        } \
+        else \
+        { \
+            result->tuple_array = NULL; \
+            bool failed = false; \
+            if (count > 0) \
+            { \
+                /*Codes_SRS_THANDLE_TUPLE_ARRAY_42_002: [ THANDLE_TUPLE_ARRAY_CREATE(name) shall allocate memory for count structs which hold the tuple. ]*/ \
+                required_size = count * sizeof(name); \
+                result->tuple_array = (name *)malloc(required_size); \
```
anporumb, line 73: >malloc [](http://example.com/codeflow?start=46&length=6)

Same can be achieved using just 1 malloc and flexible arrays. Would imply having line 37 written as name tuple_array[].

Saves 1 malloc.

inc/c_util/thandle_tuple_array.h
```diff
@@ -0,0 +1,147 @@
...
+            LogError("malloc(%zu) failed", required_size); \
+        } \
+        else \
+        { \
+            result->tuple_array = NULL; \
+            bool failed = false; \
+            if (count > 0) \
+            { \
+                /*Codes_SRS_THANDLE_TUPLE_ARRAY_42_002: [ THANDLE_TUPLE_ARRAY_CREATE(name) shall allocate memory for count structs which hold the tuple. ]*/ \
+                required_size = count * sizeof(name); \
```
anporumb, line 72: >* [](http://example.com/codeflow?start=38&length=1)

In theory can overflow. IN practice... unlikely unless count comes from some uninitialized/freed memory maybe (0xCCCCCCCC etc).

inc/c_util/thandle_tuple_array.h
```diff
@@ -0,0 +1,147 @@
...
+        { \
+            /*Codes_SRS_THANDLE_TUPLE_ARRAY_42_007: [ THANDLE_TUPLE_ARRAY_DESTROY(name) shall iterate over all of the elements in tuple_array and call THANDLE_DEC_REF(type) for each field. ]*/ \
+            for (uint32_t i = 0; i < tuple_array->count; ++i) \
+            { \
+                MU_FOR_EACH_2(THANDLE_TUPLE_ARRAY_FREE_MEMBER, __VA_ARGS__) \
+            } \
+            /*Codes_SRS_THANDLE_TUPLE_ARRAY_42_008: [ THANDLE_TUPLE_ARRAY_DESTROY(name) shall free the memory allocated in tuple_array. ]*/ \
+            if (tuple_array->tuple_array != NULL) \
+            { \
+                free((void*)tuple_array->tuple_array); \
```
anporumb, line 133: >(void*) [](http://example.com/codeflow?start=21&length=7)

what happens if you remove this cast?

tests/thandle_tuple_array_ut/thandle_tuple_array_ut.c
```diff
@@ -0,0 +1,479 @@
...
+    STRICT_EXPECTED_CALL(malloc(sizeof(TUPLE_ONE)));
+    STRICT_EXPECTED_CALL(THANDLE_INITIALIZE(RC_STRING)(IGNORED_ARG, NULL));
+
+    // act
+    THANDLE_TUPLE_ARRAY_TYPE(TUPLE_ONE)* result = THANDLE_TUPLE_ARRAY_CREATE(TUPLE_ONE)(1);
+
+    // assert
+    ASSERT_ARE_EQUAL(char_ptr, umock_c_get_expected_calls(), umock_c_get_actual_calls());
+    ASSERT_IS_NOT_NULL(result);
+    ASSERT_ARE_EQUAL(uint32_t, 1, result->count);
```
anporumb, line 175: >ASSERT_ARE_EQUAL [](http://example.com/codeflow?start=4&length=16)

missing assert that the element of the array is NULL (per SRS_THANDLE_TUPLE_ARRAY_42_003)

tests/thandle_tuple_array_ut/thandle_tuple_array_ut.c
```diff
@@ -0,0 +1,479 @@
...
+    STRICT_EXPECTED_CALL(THANDLE_INITIALIZE(RC_STRING)(IGNORED_ARG, NULL));
+    STRICT_EXPECTED_CALL(THANDLE_INITIALIZE(RC_STRING)(IGNORED_ARG, NULL));
+
+    // act
+    THANDLE_TUPLE_ARRAY_TYPE(TUPLE_THREE)* result = THANDLE_TUPLE_ARRAY_CREATE(TUPLE_THREE)(1);
+
+    // assert
+    ASSERT_ARE_EQUAL(char_ptr, umock_c_get_expected_calls(), umock_c_get_actual_calls());
+    ASSERT_IS_NOT_NULL(result);
+    ASSERT_ARE_EQUAL(uint32_t, 1, result->count);
```
anporumb, line 200: >ASSERT_ARE_EQUAL [](http://example.com/codeflow?start=4&length=16)

should assert that every element of the array is NULL)

anporumb: sort of here and everywhere

anporumb (approved):
:shipit:

inc/c_util/thandle_tuple_array.h
mattdurak: Sure, can fail in theory. Not really easy to write a test with the sizeof but the check is easy to have for "theoretical" reasons, I will add it.

inc/c_util/thandle_tuple_array.h
mattdurak: Yes, doesn't hurt to optimize this

inc/c_util/thandle_tuple_array.h
mattdurak: can't free a const pointer (and that's good because it catches mistakes of calling free instead of destroy)

tests/thandle_tuple_array_ut/thandle_tuple_array_ut.c
mattdurak: yeah makes sense

devdoc/thandle_tuple_array_requirements.md
mattdurak: I added a note about this, saying that this could be changed to be a THANDLE as an improvement.

devdoc/thandle_tuple_array_requirements.md
mattdurak: Added the note that this is not ref counted. I'm leaving the name because I want it clear to callers that tuple member access is with THANDLE's.

PR #131, opened 2021-09-03
constbuffer serialization

anporumb:


inc/c_util/constbuffer.h
```diff
@@ -32,6 +32,25 @@ typedef struct CONSTBUFFER_TAG
...
+#define CONSTBUFFER_TO_FIXED_SIZE_BUFFER_RESULT_VALUES \
+    CONSTBUFFER_TO_FIXED_SIZE_BUFFER_RESULT_OK, \
+    CONSTBUFFER_TO_FIXED_SIZE_BUFFER_RESULT_ERROR, \
+    CONSTBUFFER_TO_FIXED_SIZE_BUFFER_RESULT_INSUFFICIENT_BUFFER, \
+    CONSTBUFFER_TO_FIXED_SIZE_BUFFER_RESULT_INVALID_ARG
+
+MU_DEFINE_ENUM(CONSTBUFFER_TO_FIXED_SIZE_BUFFER_RESULT, CONSTBUFFER_TO_FIXED_SIZE_BUFFER_RESULT_VALUES)
+
+#define CONSTBUFFER_FROM_BUFFER_RESULT_VALUES   \
+    CONSTBUFFER_FROM_BUFFER_RESULT_OK ,\
```
dcristoloveanu, line 47: >  [](http://example.com/codeflow?start=37&length=1)

Nitpick: space after ,

inc/c_util/constbuffer.h
```diff
@@ -32,6 +32,25 @@ typedef struct CONSTBUFFER_TAG
 
 typedef void(*CONSTBUFFER_CUSTOM_FREE_FUNC)(void* context);
 
+/*what function should CONSTBUFFER_HANDLE_to_buffer use to allocate the returned serialized form. NULL means malloc from gballoc_hl_malloc_redirect.h of this lib.*/
+typedef void*(*CONSTBUFFER_to_buffer_alloc)(size_t size, void* context);
```
dcristoloveanu, line 36: Is there a specific use case we need this now? Up to now everything goes through the redirected malloc/free except in very very specific cases like ownership transfer and some THANDLE/COM_WRAPPER cases which are mainly for testing.

If we have a clear use case upstream sure. If we don't and this won't get used then consider not having it.

inc/c_util/constbuffer_format.h
```diff
@@ -0,0 +1,17 @@
...
+
+#define CONSTBUFFER_VERSION_OFFSET 0
+#define CONSTBUFFER_VERSION_SIZE (sizeof(uint8_t))
+
+#define CONSTBUFFER_SIZE_OFFSET (CONSTBUFFER_VERSION_OFFSET + CONSTBUFFER_VERSION_SIZE)
+#define CONSTBUFFER_SIZE_SIZE (sizeof(uint32_t))
+
+#define CONSTBUFFER_CONTENT_OFFSET (CONSTBUFFER_SIZE_OFFSET + CONSTBUFFER_SIZE_SIZE)
+
+#endif  /* CONSTBUFFER_FORMAT_H */
```
dcristoloveanu, line 17: A-ha, so these are here to have them sort of "internal". Interesting.

inc/c_util/constbuffer.h
```diff
@@ -53,7 +72,15 @@ MOCKABLE_INTERFACE(constbuffer,
 
     FUNCTION(, const CONSTBUFFER*, CONSTBUFFER_GetContent, CONSTBUFFER_HANDLE, constbufferHandle),
 
-    FUNCTION(, bool, CONSTBUFFER_HANDLE_contain_same, CONSTBUFFER_HANDLE, left, CONSTBUFFER_HANDLE, right)
+    FUNCTION(, bool, CONSTBUFFER_HANDLE_contain_same, CONSTBUFFER_HANDLE, left, CONSTBUFFER_HANDLE, right),
+
+    FUNCTION(, uint32_t, CONSTBUFFER_get_serialization_size, CONSTBUFFER_HANDLE, source),
+
+    FUNCTION(, unsigned char*, CONSTBUFFER_to_buffer, CONSTBUFFER_HANDLE, source, CONSTBUFFER_to_buffer_alloc, alloc, void*, alloc_context, uint32_t*, size),
```
dcristoloveanu, line 79: >size [](http://example.com/codeflow?start=151&length=4)

is this serialized_size? I presume.

src/constbuffer.c
```diff
@@ -394,3 +398,256 @@ IMPLEMENT_MOCKABLE_FUNCTION(, bool, CONSTBUFFER_HANDLE_contain_same, CONSTBUFFER
...
+        {
+            /*Codes_SRS_CONSTBUFFER_02_066: [ If size is 0 then CONSTBUFFER_from_buffer shall fail and return CONSTBUFFER_FROM_BUFFER_RESULT_INVALID_ARG. ]*/
+            LogError("cannot deserialize from size=%" PRIu32 "", size);
+            result = CONSTBUFFER_FROM_BUFFER_RESULT_INVALID_ARG;
+        }
+        else
+        {
+            uint8_t version;
+            read_uint8_t(source + CONSTBUFFER_VERSION_OFFSET, &version);
+            if (version != CONSTBUFFER_VERSION)
```
dcristoloveanu, line 604: >CONSTBUFFER_VERSION [](http://example.com/codeflow?start=27&length=19)

Technically this is the serializer for v1. When v2 would appear I would bet that the v1 serializer will still have to exist (because that is the only way we can do upgrades without interruption of service). So maybe we should consider having constbuffer_serializer_v1 as a unit. And this unit would only do v1 serialization/deserialization. Hardcoded. So when v2 comes along we can slam a v2 next to it.

Of course we can also do the rename when v2 appears :-). Delay until maybe not needed.

src/constbuffer.c
```diff
@@ -394,3 +398,256 @@ IMPLEMENT_MOCKABLE_FUNCTION(, bool, CONSTBUFFER_HANDLE_contain_same, CONSTBUFFER
...
+            {
+                /*Codes_SRS_CONSTBUFFER_02_067: [ If source byte at offset 0 is not 1 (current version) then CONSTBUFFER_from_buffer shall fail and return CONSTBUFFER_FROM_BUFFER_RESULT_INVALID_DATA. ]*/
+                LogError("different version (%" PRIu8 ") detected. This module only knows about version %" PRIu8 "", version, CONSTBUFFER_VERSION);
+                result = CONSTBUFFER_FROM_BUFFER_RESULT_INVALID_DATA;
+            }
+            else
+            {
+                /*Codes_SRS_CONSTBUFFER_02_068: [ If source's size is less than sizeof(uint8_t) + sizeof(uint32_t) then CONSTBUFFER_from_buffer shall fail and return CONSTBUFFER_FROM_BUFFER_RESULT_INVALID_DATA. ]*/
+                if (size < sizeof(uint8_t) + sizeof(uint32_t))
```
dcristoloveanu, line 613: >sizeof(uint8_t) + sizeof(uint32_t) [](http://example.com/codeflow?start=27&length=34)

I'd say CONSTBUFFER_CONTENT_OFFSET. Or the sum of the sizes from the format so it is clear why sizeof(uint8_t) + sizeof(uint32_t)

src/constbuffer.c
```diff
@@ -394,3 +398,256 @@ IMPLEMENT_MOCKABLE_FUNCTION(, bool, CONSTBUFFER_HANDLE_contain_same, CONSTBUFFER
...
+                {
+                    LogError("cannot deserialize when the numbe of serialized bytes cannot be determined. size=%" PRIu32 "", size);
+                    result = CONSTBUFFER_FROM_BUFFER_RESULT_INVALID_DATA;
+                }
+                else
+                {
+                    /*Codes_SRS_CONSTBUFFER_02_069: [ CONSTBUFFER_from_buffer shall read the number of serialized content bytes from offset 1 of source. ]*/
+                    uint32_t content_size;
+                    read_uint32_t(source + CONSTBUFFER_SIZE_OFFSET, &content_size);
```
dcristoloveanu, line 622: >CONSTBUFFER_SIZE_OFFSET [](http://example.com/codeflow?start=43&length=23)

Here we use it, hihi :-)

src/constbuffer.c
```diff
@@ -394,3 +398,256 @@ IMPLEMENT_MOCKABLE_FUNCTION(, bool, CONSTBUFFER_HANDLE_contain_same, CONSTBUFFER
...
+    {
+        LogError("invalid arguments CONSTBUFFER_HANDLE source=%p, unsigned char* destination=%p, uint32_t destination_size=%" PRIu32 ", uint32_t* serialized_size=%p",
+            source, destination, destination_size, serialized_size);
+        result = CONSTBUFFER_TO_FIXED_SIZE_BUFFER_RESULT_INVALID_ARG;
+    }
+    else
+    {
+        if (UINT32_MAX - (sizeof(uint8_t) + sizeof(uint32_t)) < source->alias.size)
+        {
+            /*when this is true, an output for "serialized_size" cannot be provided as the complete serialization would exceed UINT32_MAX*/
```
dcristoloveanu, line 533: >/*when this is true, an output for "serialized_size" cannot be provided as the complete serialization would exceed UINT32_MAX*/ [](http://example.com/codeflow?start=12&length=127)

Missing LogError?

src/constbuffer.c
```diff
@@ -394,3 +398,256 @@ IMPLEMENT_MOCKABLE_FUNCTION(, bool, CONSTBUFFER_HANDLE_contain_same, CONSTBUFFER
...
+
+            /*Codes_SRS_CONSTBUFFER_02_074: [ If there are any failures then CONSTBUFFER_to_fixed_size_buffer shall fail and return CONSTBUFFER_TO_FIXED_SIZE_BUFFER_RESULT_ERROR. ]*/
+            result = CONSTBUFFER_TO_FIXED_SIZE_BUFFER_RESULT_ERROR;
+        }
+        else
+        {
+            *serialized_size = sizeof(uint8_t) + sizeof(uint32_t) + (uint32_t)source->alias.size;
```
dcristoloveanu, line 540: >sizeof(uint8_t) + sizeof(uint32_t) [](http://example.com/codeflow?start=31&length=34)

CONSTBUFFER_CONTENT_OFFSET

src/constbuffer.c
```diff
@@ -394,3 +398,256 @@ IMPLEMENT_MOCKABLE_FUNCTION(, bool, CONSTBUFFER_HANDLE_contain_same, CONSTBUFFER
...
+            }
+            else
+            {
+                if (destination_size - (sizeof(uint8_t) + sizeof(uint32_t)) < source->alias.size)
+                {
+                    /*Codes_SRS_CONSTBUFFER_02_058: [ If the size of serialization exceeds destination_size then CONSTBUFFER_to_fixed_size_buffer shall fail, write in serialized_size how much it would need and return CONSTBUFFER_TO_FIXED_SIZE_BUFFER_RESULT_INSUFFICIENT_BUFFER. ]*/
+                    LogError("cannot serialize CONSTBUFFER_HANDLE source=%p because its serialized size (sizeof(uint8_t) + sizeof(uint32_t) + alias->size=%zu)=%zu exceeds available destination_size=%" PRIu32 "",
+                        source, source->alias.size, sizeof(uint8_t) + sizeof(uint32_t) + source->alias.size, destination_size);
+                    result = CONSTBUFFER_TO_FIXED_SIZE_BUFFER_RESULT_INSUFFICIENT_BUFFER;
+                }
```
dcristoloveanu, line 555: Why do we need 2 checks? I think we only need one: we know what is the serialized size, if the destimation buffer size is less we should error, printing the sizes and that should be it.

src/constbuffer.c
```diff
@@ -394,3 +398,256 @@ IMPLEMENT_MOCKABLE_FUNCTION(, bool, CONSTBUFFER_HANDLE_contain_same, CONSTBUFFER
...
+    else
+    {
+        /*Codes_SRS_CONSTBUFFER_02_046: [ If alloc is NULL then CONSTBUFFER_to_buffer shall use malloc as provided by gballoc_hl_redirect.h. ]*/
+        if (alloc == NULL)
+        {
+            alloc = &calls_malloc;
+        }
+
+        /*Codes_SRS_CONSTBUFFER_02_054: [ If there are any failures then CONSTBUFFER_to_buffer shall fail and return NULL. ]*/
+        if (THE_LESSER - sizeof(uint8_t) - sizeof(uint32_t) < source->alias.size)
```
dcristoloveanu, line 474: >THE_LESSER [](http://example.com/codeflow?start=12&length=10)

Please file a task in the bucket of stuff that never gets done so that we start the fire for forging the statue of the unknown hero that will change size_t to uint32_t.

tests/constbuffer_ut/constbuffer_ut.c
```diff
@@ -1554,4 +1584,973 @@ TEST_FUNCTION(CONSTBUFFER_CreateFromOffsetAndSizeWithCopy_unhappy_path) /*this f
     CONSTBUFFER_DecRef(origin);
 }
 
+/*Tests_SRS_CONSTBUFFER_02_041: [ If source is NULL then CONSTBUFFER_get_serialization_size shall fail and return 0. ]*/
+TEST_FUNCTION(CONSTBUFFER_get_serialization_size_with_source_NULL_fails)
```
dcristoloveanu, line 1588: >CONSTBUFFER_get_serialization_size [](http://example.com/codeflow?start=14&length=34)

Please add those guiding regoins for your manager :-)

/* CONSTBUFFER_get_serialization_size */

dcristoloveanu (approved):
:shipit:

inc/c_util/constbuffer.h
anporumb: Yeah, correct assumption supported by specs :). Will rename for consistency reasons

inc/c_util/constbuffer.h
anporumb: First: specs says you can pass "NULL" and then it uses malloc. That doesn't seem like a huge hinderance to me.

Second: We do not have *at the moment* a use case. However, here's a simple use case: assume you're a receiver of CONSTBUFFER_HANDLEs that need to be serialized in some byte [] array. The size of the byte[] should grow with the CONSTBUFFERs that are received. 

It could be done like 
1. get the serialized size. realloc. serialize in the new grown space. Has a drawback of calling 2 APIs (more mocks!) and sort of does the arithmetic of verification twice.

or (what I had in mind):

2. pass in a custom allocator that does the growth 

err...too much explanation to write here, I've added a sample and will ping you should you want to revisit this topic.

src/constbuffer.c
anporumb: I will rename this macro to CONSTBUFFER_VERSION_V1. And CONSTBUFFER_VERSION_V2 will come when time comes.

src/constbuffer.c
anporumb: I liked this idea very much so I've changed everywhere to use the macros for SIZES.

inc/c_util/constbuffer.h
anporumb: Removed space.

src/constbuffer.c
anporumb: changed

src/constbuffer.c
anporumb: Correct. I was too pedantic about verifying at every step.

src/constbuffer.c
anporumb: Task 10783889: Change CONSTBUFFER_HANDLE internal size from size_t to uint32_t

[https://msazure.visualstudio.com/One/_workitems/edit/10783889](https://msazure.visualstudio.com/One/_workitems/edit/10783889)

src/constbuffer.c
anporumb: The long lost LogError has been found and re-written in its rightful place!

inc/c_util/constbuffer.h
dcristoloveanu: Yeah, I get there would be cases of custom allocator that can be used.

The hindrance of passing a NULL is really nothing, the developer can write a NULL when calling this without their hand falling off :-).

What I was challenging is that "if we don't need it now it sounds that maybe we should not add it now".

But it sounds like maybe you are thinking of using this in the near future in the BSDL land, so I'm fine with leaving it in. And probably if not used in a while it will suffer of the "2 are loading, 2 are unloading" syndrome and we just remove it :-).

PR #136, opened 2021-09-11
Add InterlockedHL_CompareExchangeIf specs, header, code and tests

dcristoloveanu:
Add InterlockedHL_CompareExchangeIf specs, header, code and tests.
Example where this function is useful for switching to a given value (like a faulted state) if we are in any of 2 or more other states.


src/interlocked_hl.c
```diff
@@ -163,6 +163,53 @@ IMPLEMENT_MOCKABLE_FUNCTION(, INTERLOCKED_HL_RESULT, InterlockedHL_WaitForNotVal
...
+            else
+            {
+                /* Codes_SRS_INTERLOCKED_HL_01_014: [ If target changed meanwhile then InterlockedHL_CompareExchangeIf shall return INTERLOCKED_HL_CHANGED and shall not peform any exchange of values. ]*/
+                result = INTERLOCKED_HL_CHANGED;
+            }
+        }
+        else
+        {
+            /* Codes_SRS_INTERLOCKED_HL_01_017: [ If compare returns false then  InterlockedHL_CompareExchangeIf shall not perform any exchanges and return INTERLOCKED_HL_OK. ]*/
+            result = INTERLOCKED_HL_OK;
```
jebrando, line 204: >result = INTERLOCKED_HL_OK; [](http://example.com/codeflow?start=12&length=27)

One thing that makes me think, should we change the return value here to differentiate between a positive compare and a negative compare?

jebrando (approved):
:shipit:

src/interlocked_hl.c
```diff
@@ -163,6 +163,53 @@ IMPLEMENT_MOCKABLE_FUNCTION(, INTERLOCKED_HL_RESULT, InterlockedHL_WaitForNotVal
     return result;
 }
 
+IMPLEMENT_MOCKABLE_FUNCTION(, INTERLOCKED_HL_RESULT, InterlockedHL_CompareExchangeIf, int32_t volatile_atomic*, target, int32_t, exchange, INTERLOCKED_COMPARE_EXCHANGE_IF, compare, int32_t*, original_target)
```
anporumb, line 166: >InterlockedHL_CompareExchangeIf [](http://example.com/codeflow?start=53&length=31)

possibly missing "reals" for this new function.

anporumb (approved):
:shipit:

src/interlocked_hl.c
dcristoloveanu: reals added.

src/interlocked_hl.c
dcristoloveanu: I hear you. Let's leave that for when we hit that case though.

PR #141, opened 2021-09-30
have TARRAY

anporumb:


inc/c_util/tarray.h
```diff
@@ -0,0 +1,177 @@
...
+    {                                                                                                                                                           \
+        /*Codes_SRS_TARRAY_02_006: [ If tarray->capacity is greater than or equal to capacity then TARRAY_ENSURE_CAPACITY(T) shall succeed and return 0. ]*/    \
+        if (capacity <= tarray->capacity)                                                                                                                       \
+        {                                                                                                                                                       \
+            /*we already have the capacity*/                                                                                                                    \
+            result = 0;                                                                                                                                         \
+        }                                                                                                                                                       \
```
dcristoloveanu, line 127: So, this array never shrinks... Why not have it smart on shrinking too?

inc/c_util/tarray.h
```diff
@@ -0,0 +1,177 @@
...
+        else                                                                                                                                                    \
+        {                                                                                                                                                       \
+            if(capacity > (((uint32_t)1)<<31))                                                                                                                  \
```
dcristoloveanu, line 130: >if(capacity > (((uint32_t)1)<<31))  [](http://example.com/codeflow?start=12&length=35)

Once could challenge that we can still grow here. Buuuut yeah, we can say these huge values are probably not of interest. At least we should document in the .md this limit. Maybe someone reads that. Maybe.

inc/c_util/tarray.h
```diff
@@ -0,0 +1,177 @@
...
+            {                                                                                                                                                   \
+                /*Codes_SRS_TARRAY_02_009: [ If there are any failures then TARRAY_ENSURE_CAPACITY(T) shall fail and return a non-zero value. ]*/               \
+                LogError("capacity (%" PRIu32 ")cannot be improved by doubling", capacity);                                                                     \
```
dcristoloveanu, line 133: >) [](http://example.com/codeflow?start=47&length=1)

Nitpick: space

tests/tarray_ut/tarray_ut.c
```diff
@@ -0,0 +1,333 @@
...
+static TEST_MUTEX_HANDLE g_testByTest;
+
+#define THANDLE_MALLOC_FUNCTION malloc
+#define THANDLE_FREE_FUNCTION free
+
+/*TARRAY with regular types*/
+TARRAY_TYPE_DECLARE(uint32_t);
+TARRAY_TYPE_DEFINE(uint32_t);
+
+/*TARRAY works with THANDLEs too*/
```
dcristoloveanu, line 46: Seems that this should be before ln 56. Here there should be a comment that it works with structs

dcristoloveanu (approved):
:shipit:

tests/tarray_ut/tarray_ut.c
anporumb: Moved to line 56

inc/c_util/tarray.h
anporumb: added a space

inc/c_util/tarray.h
anporumb: This exists in order to avoid an overflow down the road where a multiplication result would extend past the UINT32_MAX.

However, it is now documented in .md as a separate spec.

inc/c_util/tarray.h
anporumb: I guess it is one of those cases "when time comes". Looks like it would be some easy implementation.

PR #146, opened 2021-10-04
THANDLE / TARRAY with LL

anporumb:


tests/reals/real_tarray_undo_op.h
```diff
@@ -0,0 +1,40 @@
...
+#include "c_util/tarray_ll.h"
+#include "c_util/tarray.h"
+
+#include "../tarray_int/play_undo_op_types.h"
+#include "../tarray_int/play_undo_op_tarray_types.h"
+
+#define R2(X) REGISTER_GLOBAL_MOCK_HOOK(X, real_##X);
+
+#define REGISTER_TARRAY_UNDO_OP_GLOBAL_MOCK_HOOK() \
+    REGISTER_GLOBAL_MOCK_HOOK(TARRAY_ENSURE_CAPACITY(UNDO_OP), TARRAY_LL_ENSURE_CAPACITY(real_UNDO_OP)) \
```
mattdurak, line 18: >UNDO_OP [](http://example.com/codeflow?start=53&length=7)

Why have this real for a "test" type here? I thought generally in the reals folder we only put re-usable reals that are needed by consumers of this lib. Seems like this is for a type defined in tarray_int.

inc/c_util/tarray_ll.h
```diff
@@ -0,0 +1,183 @@
...
+#include <inttypes.h>
+#endif
+
+#include "c_pal/interlocked.h"
+
+#include "c_util/thandle_ll.h"
+
+#include "umock_c/umock_c_prod.h"
+
+/*file acronyms: C = code, T = data Type. C is a decoration for code, T is a decoration for Data types. Having 2 decorations allows*/
```
mattdurak, line 19: >C = code, T = data Type. C is a decoration for code, T is a decoration for Data types [](http://example.com/codeflow?start=17&length=85)

I don't know if I fully understand what "C" means from this comment...maybe an example or something would help or maybe I need to read more of the code.

mattdurak: Maybe also "decoration" should be "alias" or "placeholder", but it's still not completely clear to me

inc/c_util/tarray.h
```diff
@@ -10,167 +10,38 @@
...
-/*given a type "T" TARRAY(T) expands to the name of the type. */
-#define TARRAY(T) THANDLE(TARRAY_TYPEDEF_NAME(T))
+#define TARRAY_FREE_DEFINE(T) TARRAY_LL_FREE_DEFINE(T, T)
+
+#define TARRAY_CREATE(C) TARRAY_LL_CREATE(C)
+#define TARRAY_ENSURE_CAPACITY(C) TARRAY_LL_ENSURE_CAPACITY(C)
 
-/*because TARRAY is a THANDLE, all THANDLE's macro APIs are useable with TARRAY.*/
-/*the below are just shortcuts of THANDLE's public ones*/
 #define TARRAY_INITIALIZE(T) THANDLE_INITIALIZE(TARRAY_TYPEDEF_NAME(T))
```
mattdurak, line 29: >THANDLE_INITIALIZE [](http://example.com/codeflow?start=29&length=18)

Shouldn't these be TARRAY_LL?

inc/c_util/tarray_ll.h
mattdurak: I read a bit further, maybe this would help me: "C = the type name to use for function names, T = the type name to use as a type" (or something like that.

inc/c_util/tarray_ll.h
```diff
@@ -0,0 +1,183 @@
...
+    if(result == NULL)                                                                                                      \
+    {                                                                                                                       \
+        LogError("failure in malloc(" MU_TOSTRING(TARRAY_TYPEDEF_NAME(T)) "=%zu", sizeof(TARRAY_TYPEDEF_NAME(T)));          \
+        /*return as is*/                                                                                                    \
+    }                                                                                                                       \
+    else                                                                                                                    \
+    {                                                                                                                       \
+        /*Codes_SRS_TARRAY_02_002: [ TARRAY_CREATE(T) shall call malloc to allocate result->arr. ]*/                        \
+        result->arr = malloc(1 * sizeof(T));                                                                                \
```
mattdurak, line 87: > malloc(1 * sizeof(T));   [](http://example.com/codeflow?start=21&length=25)

unrelated to splitting the LL I guess, but why not have the  caller give an initial capacity to avoid allocating an array of 1 first if the initial size is known?

inc/c_util/tarray.h
```diff
@@ -10,167 +10,38 @@
...
-    }                                                                                                                                                           \
-    return result;                                                                                                                                              \
-}
-
 /*macro to be used in headers*/                                                                                     \
 #define TARRAY_TYPE_DECLARE(T)                                                                                      \
-    TARRAY_DEFINE_STRUCT_TYPE(T)                                                                                    \
-    THANDLE_TYPE_DECLARE(TARRAY_TYPEDEF_NAME(T))                                                                    \
+    /*TARRAY_DEFINE_STRUCT_TYPE(T)                                                                                   */ \
+    /*THANDLE_TYPE_DECLARE(TARRAY_TYPEDEF_NAME(T))                                                                   */ \
```
mattdurak, line 39: commented out? Should this be removed or uncommented? Here and below

inc/c_util/tarray_ll.h
```diff
@@ -0,0 +1,183 @@
...
+#define TARRAY_DEFINE_STRUCT_TYPE(T)                                                \
+typedef struct TARRAY_STRUCT_TYPE_NAME_TAG(T)                                       \
+{                                                                                   \
+    uint32_t capacity;                                                              \
+    T* arr;                                                                         \
+} TARRAY_TYPEDEF_NAME(T);                                                           \
+
+/*TARRAY is-a THANDLE*/
+/*given a type "T" TARRAY(T) expands to the name of the type. */
+#define TARRAY(T) THANDLE(TARRAY_TYPEDEF_NAME(T))
```
mattdurak, line 36: Why aren't these TARRAY_LL?

inc/c_util/tarray_ll.h
```diff
@@ -0,0 +1,183 @@
...
+                }                                                                                                                                               \
+            }                                                                                                                                                   \
+        }                                                                                                                                                       \
+    }                                                                                                                                                           \
+    return result;                                                                                                                                              \
+}
+
+/*macro to be used in headers*/                                                                                     \
+#define TARRAY_LL_TYPE_DECLARE(C, T)                                                                                \
+    /*hint: have TARRAY_DEFINE_STRUCT_TYPE(T) before TARRAY_LL_TYPE_DECLARE*/                                       \
```
mattdurak, line 172: >TARRAY_DEFINE_STRUCT_TYPE [](http://example.com/codeflow?start=17&length=25)

This comment makes me wonder what I'm supposed to do when defining my own TARRAY and whether I need to use the LL outside of tests...but I guess I should look at some examples in the tests first.

mattdurak: Although it seems like we should have an example in the comments here or (probably better) have an MD at least with the "how to use this" example. I see some SRS in this file but no MD in the PR

tests/reals/real_tarray_undo_op.c
```diff
@@ -0,0 +1,22 @@
...
+
+#include "real_tarray_undo_op.h"
+
+#define THANDLE_MALLOC_FUNCTION malloc
+#define THANDLE_FREE_FUNCTION free
+
+THANDLE_LL_TYPE_DECLARE(TARRAY_TYPEDEF_NAME(real_UNDO_OP), TARRAY_TYPEDEF_NAME(UNDO_OP));
+THANDLE_LL_TYPE_DEFINE(TARRAY_TYPEDEF_NAME(real_UNDO_OP), TARRAY_TYPEDEF_NAME(UNDO_OP));
+
+TARRAY_LL_TYPE_DEFINE(real_UNDO_OP, UNDO_OP);
```
mattdurak, line 19: >TARRAY_LL_TYPE_DEFINE [](http://example.com/codeflow?start=0&length=21)

So `TARRAY_TYPE_DEFINE` isn't needed here?

tests/tarray_int/play_undo_op_tarray_types.h
```diff
@@ -0,0 +1,32 @@
...
+
+#include "umock_c/umock_c_prod.h"
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
+    THANDLE_TYPE_DECLARE(TARRAY_TYPEDEF_NAME(UNDO_OP));
+
+    TARRAY_TYPE_DECLARE(UNDO_OP);
```
mattdurak, line 24: >TARRAY_TYPE_DECLARE [](http://example.com/codeflow?start=4&length=19)

Can this UNDO_OP be used without the LL declare/define?

inc/c_util/tarray.h
```diff
@@ -10,167 +10,38 @@
...
+#define TARRAY_ENSURE_CAPACITY_DECLARE(T) TARRAY_LL_ENSURE_CAPACITY_DECLARE(T, T)
+#define TARRAY_ENSURE_CAPACITY_DEFINE(T) TARRAY_LL_ENSURE_CAPACITY_DEFINE(T, T)
 
-/*TARRAY is-a THANDLE*/
-/*given a type "T" TARRAY(T) expands to the name of the type. */
-#define TARRAY(T) THANDLE(TARRAY_TYPEDEF_NAME(T))
+#define TARRAY_FREE_DEFINE(T) TARRAY_LL_FREE_DEFINE(T, T)
+
+#define TARRAY_CREATE(C) TARRAY_LL_CREATE(C)
+#define TARRAY_ENSURE_CAPACITY(C) TARRAY_LL_ENSURE_CAPACITY(C)
```
mattdurak, line 27: After reading thandle.h, I'm wondering if most of these can be removed and the LL can just be directly used in lines 37-47. Do these need to be exposed with these names?

mattdurak (approved):
:shipit:

inc/c_util/tarray_ll.h
anporumb: It is a good suggestion, I will take it with another task. Here's the task: 

Task 11115085: TARRAY_CREATE can take an initial capacity

([https://msazure.visualstudio.com/One/_workitems/edit/11115085](https://msazure.visualstudio.com/One/_workitems/edit/11115085))

tests/tarray_int/play_undo_op_tarray_types.h
anporumb: In PROD code there shouldn't be any need to use LL variety of macros. They are only needed for crafting "reals" during testing.

inc/c_util/tarray.h
anporumb: Of course they should, changed now :)

inc/c_util/tarray_ll.h
anporumb: Added a comment near TARRAY that tarray_int tests can be used a template for tests/reals.

inc/c_util/tarray.h
anporumb: These are just hints.

Just like the ones in tarray_ll.h. 

Now they are properly marked as such.

inc/c_util/tarray_ll.h
anporumb: They are now.

tests/reals/real_tarray_undo_op.h
anporumb: Your observation is awesome. 

I will move it to some other place.

It **used** to be a subfolder in tarray_int but then CMake would wine about building twice the same lib (and that does happen, as the same folder is parsed for .dll and for .exe tests). So that's why it landed in reals, but I agree this is not the right place.

Moved now.

inc/c_util/tarray_ll.h
anporumb: This is one of those comments that only creates chaos. Removed.

inc/c_util/tarray.h
anporumb: TARRAY is just a wrapper over TARRAY_LL where C==T. I don't see a problem with they way these are written, as they are a bit of separation of concerns:
concern1 : how is TARRAY linked to TARRAY_LL (addressed here)

concern 2: what's the content of a TARRAY_DECLARE/DEFINE (addressed in lines 36-47)

PR #149, opened 2021-10-08
Try Semmle

dcristoloveanu:


dcristoloveanu:
Too old PR, If we'll come back to it will re-open

PR #165, opened 2021-11-16
Add CONSTBUFFER_CreateWithOffsetAndSize

avranju:


inc/c_util/constbuffer.h
```diff
@@ -57,6 +57,8 @@ MOCKABLE_INTERFACE(constbuffer,
     /*this creates a new constbuffer from a memory area*/
     FUNCTION(, CONSTBUFFER_HANDLE, CONSTBUFFER_Create, const unsigned char*, source, uint32_t, size),
 
+    FUNCTION(, CONSTBUFFER_HANDLE, CONSTBUFFER_CreateWithOffsetAndSize, const unsigned char*, source, uint32_t, offset, uint32_t, size),
```
dcristoloveanu, line 60: >CONSTBUFFER_CreateWithOffsetAndSize [](http://example.com/codeflow?start=35&length=35)

My first reaction was: can we just use CONSTBUFFER_CreateWithCustomFree?
It seems this new function wants to say the buffer starts ar a given offset in a memory area (let's call that start).
So we have some memory area starting at start and spanning n bytes.

And I understand we want to create a buffer [start + offset, start+offset+size]

One could use CONSTBUFFER_CreateWithCustomFree, pass start+offset as source argument, pass free as free function and start as context.

dcristoloveanu: Then I realized this actually copies the memory, so I guess this does a different job.

But I think we would want as much as possible to use the constructors that do not copy the memory. Copying memory is not really desired, in the past we had to make a serious push to eliminate copying the memory for the user payloads as they did turn out to be quite an interesting bottleneck at some point.

dcristoloveanu: If we need the memory copy I am fine with having this extra constructor for easier use - it is a bit easier to use than CONSTBUFFER_CreateWithCustomFree which is totally generic though.

dcristoloveanu:
>     FUNCTION(, CONSTBUFFER_HANDLE, CONSTBUFFER_CreateFromOffsetAndSize, CONSTBUFFER_HANDLE, handle, uint32_t, offset, uint32_t, size),

I cannot help but notice that we will have CONSTBUFFER_CreateWithOffsetAndSize and CONSTBUFFER_CreateFromOffsetAndSize.

Which will not be confusing at all :-).

I think maybe CONSTBUFFER_CreateFromOffsetAndSize should be renamed to be CONSTBUFFER_CreateFromConstBufferWithOffsetAndSize

---
Refers to: inc/c_util/constbuffer.h:69 in 9c9900d. [](commit_id = 9c9900d413511b22ad0b0a1f5d18c8ba6b3857da, deletion_comment = False)

dcristoloveanu:
>     FUNCTION(, CONSTBUFFER_HANDLE, CONSTBUFFER_CreateFromOffsetAndSize, CONSTBUFFER_HANDLE, handle, uint32_t, offset, uint32_t, size),

Same for CONSTBUFFER_CreateFromOffsetAndSizeWithCopy

---
In reply to: [971143262](https://github.com/Azure/c-util/pull/165#issuecomment-971143262) [](http://example.com/codeflow?ancestors=971143262)

---
Refers to: inc/c_util/constbuffer.h:69 in 9c9900d. [](commit_id = 9c9900d413511b22ad0b0a1f5d18c8ba6b3857da, deletion_comment = False)

inc/c_util/constbuffer.h
dcristoloveanu, line 60: >CONSTBUFFER_CreateWithOffsetAndSize [](http://example.com/codeflow?start=35&length=35)

Clearly missing specs, so I will stop and wait for the specs to come in.

dcristoloveanu (changes requested):
:clock1:

inc/c_util/constbuffer.h
dcristoloveanu: Ah, I got it this is only here for the spike that you have in the main repo, so that is why it does not have specs/tests :-).

Did not know the context until I saw the other one and read all mails.

avranju: So, looks like we can actually [do pointer arithmetic](https://docs.microsoft.com/en-us/dotnet/api/system.intptr.add?view=net-5.0) in .NET with `IntPtr`. So then maybe we don't need a new constructor in C `CONSTBUFFER` because we can just use `CONSTBUFFER_Create` (as it was doing so till now) and just add the offset from the .NET side.

```c#
IntPtr address = IntPtr.Add(pinnedPayload.AddrOfPinnedObject(), (int)this.offset);
```

As for copying data, that seems to be the current design, i.e., anytime someone accesses the `NativeHandle` property, an alloc and buffer copy happens on the native side (albeit once per `ConstBuffer` instance). Are you proposing we change this design and have the C side just have a direct pointer to a pinned managed buffer or something?

avranju: I'll do this. Given that we can do pointer arithmetic on `IntPtr` in .NET I'll abandon this PR. We can talk separately about whether a design change is in order for the managed `ConstBuffer` type.

avranju:
Turns out we may not need a new constructor function in C just to do pointer arithmetic.

PR #167, opened 2022-01-11
have uuid_to/from string in c_util

anporumb:


inc/c_util/uuid.h
```diff
@@ -1,63 +1,35 @@
...
-
-#ifdef __cplusplus
-}
-#endif
-
-#endif /* UUID_H */
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#ifndef C_UTIL_UUID_H
```
dcristoloveanu, line 4: >C_UTIL_UUID_H [](http://example.com/codeflow?start=8&length=13)

Can we rename this to uuid_string or something like that (as that is what it does - conversions between uuid and string)

We clearly had to change the multiple include guards so I suspect we'll hit some level of weirdness with the fact that we will have multiple uuid.h files. At least when  searching we'll always be baffled "which uuid.h" is this one!?"

dcristoloveanu (approved):
:shipit:

inc/c_util/uuid.h
anporumb: oooh, i like this. I will rename the include and the files, it makes so much more sense to have it named uuid_string.h/c etc

stay tuned!

PR #181, opened 2022-06-11
Add helpers for DList

parth21999:


devdoc/doublylinkedlist_requirements.md
```diff
@@ -20,13 +20,36 @@ typedef struct DLIST_ENTRY_TAG
...
+MU_DEFINE_ENUM(DLIST_CONDITION_FUNCTION_RESULT, DLIST_CONDITION_FUNCTION_VALUES);
+
+#define DLIST_CONDITION_FUNCTION_RESULT_VALUES \
+    DLIST_CONDITION_FUNCTION_SATISFIED, \
+    DLIST_CONDITION_FUNCTION_NOT_SATISFIED, \
+    DLIST_CONDITION_FUNCTION_ERROR
+
+MU_DEFINE_ENUM(DLIST_CONDITION_FUNCTION_RESULT, DLIST_CONDITION_FUNCTION_RESULT_VALUES);
+
+typedef DLIST_CONDITION_FUNCTION_RESULT (*DLIST_MATCH_FUNCTION)(PDLIST_ENTRY listEntry, const void* matchContext);
```
darobs, line 37: DLIST_MATCH_FUNCTION_RESULT?

parth21999: fixed.

devdoc/doublylinkedlist_requirements.md
```diff
@@ -20,13 +20,36 @@ typedef struct DLIST_ENTRY_TAG
...
 
+#define DLIST_MATCH_FUNCTION_RESULT_VALUES \
+    DLIST_MATCH_FUNCTION_MATCHING, \
+    DLIST_MATCH_FUNCTION_NOT_MATCHING, \
+    DLIST_MATCH_FUNCTION_ERROR
+
+MU_DEFINE_ENUM(DLIST_MATCH_FUNCTION_RESULT, DLIST_MATCH_FUNCTION_RESULT_VALUES);
+
+#define DLIST_CONDITION_FUNCTION_RESULT_VALUES \
+    DLIST_CONDITION_FUNCTION_SATISFIED, \
```
dcristoloveanu, line 31: >SATISFIED [](http://example.com/codeflow?start=29&length=9)

I'd probably had used TRUE/FALSE/ERROR. Sounds a bit more intuitive when writing the user code. But I guess this will work too, so I'll let you decide.

devdoc/doublylinkedlist_requirements.md
```diff
@@ -104,3 +127,46 @@ extern PDLIST_ENTRY DList_RemoveHeadList(PDLIST_ENTRY listHead);
...
 Note: The Flink & Blink of the returned PDLIST_ENTRY shall be undefined.
+
+### DList_Find
+```c
+MOCKABLE_FUNCTION(, PDLIST_ENTRY, DList_Find, PDLIST_ENTRY, listHead, DLIST_MATCH_FUNCTION, matchFunction, const void*, matchContext);
+```
+
+**SRS_DLIST_43_001: [** `DList_Find` shall call `matchFunction` on each entry in the list defined by `listHead` along with `matchContext`. **]**
+
+**SRS_DLIST_43_002: [** If the call to `matchFunction` for an entry returns `DLIST_MATCH_FUNCTION_MATCHING`, `DList_Find` shall return that entry. **]**
```
dcristoloveanu, line 138: >_DLIST_43_002: [** If the call to `matchFunction` for an entry returns `DLIST_MATCH_FUNCTION_MAT [](http://example.com/codeflow?start=5&length=96)

What if ERROR is returned? What is specced to happen then?
I think you need to have a return enum for this function and an out arg that actually returns the found element.

devdoc/doublylinkedlist_requirements.md
```diff
@@ -104,3 +127,46 @@ extern PDLIST_ENTRY DList_RemoveHeadList(PDLIST_ENTRY listHead);
...
+**SRS_DLIST_43_003: [** If no calls to `matchFunction` return `DLIST_MATCH_FUNCTION_MATCHING`, `DList_Find` shall return `NULL`. **]**
+
+### DList_RemoveIf
+```c
+MOCKABLE_FUNCTION(, PDLIST_ENTRY, DList_RemoveIf, PDLIST_ENTRY, listHead, DLIST_CONDITION_FUNCTION, conditionFunction, const void*, conditionContext, DLIST_ENTRY_DESTROY_FUNCTION, destroyFunction, const void*, destroyContext);
+```
+
+**SRS_DLIST_43_004: [** `DList_RemoveIf` shall call `conditionFunction` on each entry in the list defined by `listHead` along with `conditionContext`. **]**
+
+**SRS_DLIST_43_005: [** If the call to `conditionFunction` for an entry returns `DLIST_CONDITION_FUNCTION_SATISFIED`:**]**
```
dcristoloveanu, line 149: >FUNCTION_SATISFIED [](http://example.com/codeflow?start=97&length=18)

Missing handling of ERROR.

inc/c_util/doublylinkedlist.h
```diff
@@ -4,26 +4,54 @@
...
+{
+#endif
 MOCKABLE_FUNCTION(, void, DList_InitializeListHead, PDLIST_ENTRY, listHead);
 MOCKABLE_FUNCTION(, int, DList_IsListEmpty, const PDLIST_ENTRY, listHead);
 MOCKABLE_FUNCTION(, void, DList_InsertTailList, PDLIST_ENTRY, listHead, PDLIST_ENTRY, listEntry);
 MOCKABLE_FUNCTION(, void, DList_InsertHeadList, PDLIST_ENTRY, listHead, PDLIST_ENTRY, listEntry);
 MOCKABLE_FUNCTION(, void, DList_AppendTailList, PDLIST_ENTRY, listHead, PDLIST_ENTRY, ListToAppend);
 MOCKABLE_FUNCTION(, int, DList_RemoveEntryList, PDLIST_ENTRY, listEntry);
 MOCKABLE_FUNCTION(, PDLIST_ENTRY, DList_RemoveHeadList, PDLIST_ENTRY, listHead);
+MOCKABLE_FUNCTION(, PDLIST_ENTRY, DList_Find, PDLIST_ENTRY, listHead, DLIST_MATCH_FUNCTION, matchFunction, const void*, matchContext);
```
dcristoloveanu, line 52: >PDLIST_ENTRY [](http://example.com/codeflow?start=20&length=12)

This is not accounting at all for the error cases returned by DLIST_MATCH_FUNCTION.
Since we are implementing the error enum value it needs to be handled.
Shall wait for an update.

dcristoloveanu (changes requested):
:clock1:

devdoc/doublylinkedlist_requirements.md
parth21999: changed to TRUE/FALSE

devdoc/doublylinkedlist_requirements.md
parth21999: added specs to handle ERROR.

devdoc/doublylinkedlist_requirements.md
parth21999: Added specs to handle ERROR.

inc/c_util/doublylinkedlist.h
parth21999: Added specs to handle ERROR, changed return to int.

devdoc/doublylinkedlist_requirements.md
```diff
@@ -104,3 +127,54 @@ extern PDLIST_ENTRY DList_RemoveHeadList(PDLIST_ENTRY listHead);
...
+MOCKABLE_FUNCTION(, int, DList_Find, PDLIST_ENTRY, listHead, DLIST_MATCH_FUNCTION, matchFunction, const void*, matchContext, PDLIST_ENTRY*, foundEntry);
+```
+
+**SRS_DLIST_43_001: [** `DList_Find` shall call `matchFunction` on each entry in the list defined by `listHead` along with `matchContext`. **]**
+
+**SRS_DLIST_43_002: [** If the call to `matchFunction` for an entry returns `DLIST_MATCH_FUNCTION_MATCHING`, `DList_Find` shall set `foundEntry` to that entry. **]**
+
+**SRS_DLIST_43_003: [** If no calls to `matchFunction` return `DLIST_MATCH_FUNCTION_MATCHING`, `DList_Find` shall set `foundEntry` to `NULL`. **]**
+
+**SRS_DLIST_43_015: [** `DList_Find` shall succeed and return zero. **]**
```
dcristoloveanu, line 142: Why not have an enum return?
FOUND/NOT_FOUND/ERROR?
That would allow checking it in a switch statement.
The way it is now one has to first check the int return and then as a second check one has to check whether foundEntry is NULL or not. 

I mean, both work, but my personal preference would be to have an enum as return.

devdoc/doublylinkedlist_requirements.md
```diff
@@ -104,3 +127,54 @@ extern PDLIST_ENTRY DList_RemoveHeadList(PDLIST_ENTRY listHead);
...
+
+**SRS_DLIST_43_016: [** If there are any failures, `DList_Find` shall fail and return a non-zero value. **]**
+
+### DList_RemoveIf
+```c
+MOCKABLE_FUNCTION(, int, DList_RemoveIf, PDLIST_ENTRY, listHead, DLIST_CONDITION_FUNCTION, conditionFunction, const void*, conditionContext, DLIST_ENTRY_DESTROY_FUNCTION, destroyFunction, const void*, destroyContext, PDLIST_ENTRY*, newHead);
```
dcristoloveanu, line 148: >newHead [](http://example.com/codeflow?start=232&length=7)

I'm not sure why we need this?

dcristoloveanu: I can't see right now a use case for this, but maybe I am blind. Can you clarify?

dcristoloveanu: Still waiting for this.

parth21999: The head of the list could be removed. The user will need a pointer to the new head of the list.

dcristoloveanu: Can you write me the user code for that scenario please?

devdoc/doublylinkedlist_requirements.md
parth21999: Changing to enum

devdoc/doublylinkedlist_requirements.md
parth21999: Tried to come up with one. In the process, I understood why it won't be needed. Removing.

dcristoloveanu (approved):
:shipit:

inc/c_util/doublylinkedlist.h
```diff
@@ -4,26 +4,61 @@
...
+MU_DEFINE_ENUM(DLIST_MATCH_FUNCTION_RESULT, DLIST_MATCH_FUNCTION_RESULT_VALUES);
+
+#define DLIST_CONDITION_FUNCTION_RESULT_VALUES \
+    DLIST_CONDITION_FUNCTION_TRUE, \
+    DLIST_CONDITION_FUNCTION_FALSE, \
+    DLIST_CONDITION_FUNCTION_ERROR
+
+MU_DEFINE_ENUM(DLIST_CONDITION_FUNCTION_RESULT, DLIST_CONDITION_FUNCTION_RESULT_VALUES);
+
+typedef DLIST_MATCH_FUNCTION_RESULT (*DLIST_MATCH_FUNCTION)(PDLIST_ENTRY listEntry, const void* matchContext);
```
anporumb, line 41: >const [](http://example.com/codeflow?start=84&length=5)

In general, I don't think EVER we should have a const void* pointer. The reason is that "just" void* should be enough.

To change the memory pointed to by the void*, it *needs* to be cast to something. The moment when it is cast... that's when we should consider const/non const.

Const is just a "pass (or not) the ownership". having const here means "never pass the ownership" - which might be extra requirement as to "don't change anything".

matchContext belongs to the user, so *the user* has ownership on it.

./conisder because we're going to cast the const away every time when it is needed anyway.

anporumb: here and everywhere

inc/c_util/doublylinkedlist.h
```diff
@@ -4,26 +4,61 @@
...
+{
+#endif
 MOCKABLE_FUNCTION(, void, DList_InitializeListHead, PDLIST_ENTRY, listHead);
 MOCKABLE_FUNCTION(, int, DList_IsListEmpty, const PDLIST_ENTRY, listHead);
 MOCKABLE_FUNCTION(, void, DList_InsertTailList, PDLIST_ENTRY, listHead, PDLIST_ENTRY, listEntry);
 MOCKABLE_FUNCTION(, void, DList_InsertHeadList, PDLIST_ENTRY, listHead, PDLIST_ENTRY, listEntry);
 MOCKABLE_FUNCTION(, void, DList_AppendTailList, PDLIST_ENTRY, listHead, PDLIST_ENTRY, ListToAppend);
 MOCKABLE_FUNCTION(, int, DList_RemoveEntryList, PDLIST_ENTRY, listEntry);
 MOCKABLE_FUNCTION(, PDLIST_ENTRY, DList_RemoveHeadList, PDLIST_ENTRY, listHead);
+MOCKABLE_FUNCTION(, DLIST_FIND_RESULT, DList_Find, PDLIST_ENTRY, listHead, DLIST_MATCH_FUNCTION, matchFunction, const void*, matchContext, PDLIST_ENTRY*, foundEntry);
```
anporumb, line 59: >DLIST_FIND_RESULT [](http://example.com/codeflow?start=20&length=17)

Consider switching the result of this function with "found entry". That is, have

MOCKABLE_FUNCTION(, PDLIST_ENTRY, DList_Find, PDLIST_ENTRY, listHead, DLIST_MATCH_FUNCTION, matchFunction, const void*, matchContext, DLIST_FIND_RESULT*, findResult);

When the function returns non-NULL-> the user cna go about his marry way.

When the function returns NULL -> the user can inspect findResult to be verbose about logging/retry whatever.

inc/c_util/doublylinkedlist.h
```diff
@@ -4,26 +4,61 @@
...
+MOCKABLE_FUNCTION(, int, DList_RemoveIf, PDLIST_ENTRY, listHead, DLIST_CONDITION_FUNCTION, conditionFunction, const void*, conditionContext, DLIST_ENTRY_DESTROY_FUNCTION, destroyFunction, const void*, destroyContext);
```
anporumb, line 60: >DLIST_ENTRY_DESTROY_FUNCTION [](http://example.com/codeflow?start=141&length=28)

What's the meaning of this function??

I cannot find it documented anywhere. I am thinking it returns (int)5.

inc/c_util/doublylinkedlist.h
```diff
@@ -4,26 +4,61 @@
...
+
+#define DLIST_CONDITION_FUNCTION_RESULT_VALUES \
+    DLIST_CONDITION_FUNCTION_TRUE, \
+    DLIST_CONDITION_FUNCTION_FALSE, \
+    DLIST_CONDITION_FUNCTION_ERROR
+
+MU_DEFINE_ENUM(DLIST_CONDITION_FUNCTION_RESULT, DLIST_CONDITION_FUNCTION_RESULT_VALUES);
+
+typedef DLIST_MATCH_FUNCTION_RESULT (*DLIST_MATCH_FUNCTION)(PDLIST_ENTRY listEntry, const void* matchContext);
+typedef DLIST_CONDITION_FUNCTION_RESULT (*DLIST_CONDITION_FUNCTION)(PDLIST_ENTRY listEntry, const void* conditionContext, bool* continueProcessing);
```
anporumb, line 42: >DLIST_CONDITION_FUNCTION [](http://example.com/codeflow?start=42&length=24)

`DLIST_CONDITION_FUNCTION` returns DLIST_CONDITION_FUNCTION_TRUE/FALSE/ERROR.

Can `DLIST_CONDITION_FUNCTION` return TRUE and have \*continueProcessing set to "false"?

Can `DLIST_CONDITION_FUNCTION` return `DLIST_CONDITION_FUNCTION_FALSE` and *continueProcessing set to "true"?

devdoc/doublylinkedlist_requirements.md
```diff
@@ -20,13 +20,43 @@ typedef struct DLIST_ENTRY_TAG
...
+
+#define DLIST_CONDITION_FUNCTION_RESULT_VALUES \
+    DLIST_CONDITION_FUNCTION_TRUE, \
+    DLIST_CONDITION_FUNCTION_FALSE, \
+    DLIST_CONDITION_FUNCTION_ERROR
+
+MU_DEFINE_ENUM(DLIST_CONDITION_FUNCTION_RESULT, DLIST_CONDITION_FUNCTION_RESULT_VALUES);
+
+typedef DLIST_MATCH_FUNCTION_RESULT (*DLIST_MATCH_FUNCTION)(PDLIST_ENTRY listEntry, const void* matchContext);
+typedef DLIST_CONDITION_FUNCTION_RESULT (*DLIST_CONDITION_FUNCTION)(PDLIST_ENTRY listEntry, const void* conditionContext, bool* continueProcessing);
```
anporumb, line 45: >DLIST_CONDITION_FUNCTION_RESULT [](http://example.com/codeflow?start=8&length=31)

I cannot find anywhere explained what this function is supposed to return under which conditions.

devdoc/doublylinkedlist_requirements.md
```diff
@@ -20,13 +20,43 @@ typedef struct DLIST_ENTRY_TAG
...
 extern void DList_InitializeListHead(PDLIST_ENTRY listHead);
 extern int DList_IsListEmpty(const PDLIST_ENTRY listHead);
 extern void DList_InsertTailList(PDLIST_ENTRY listHead, PDLIST_ENTRY listEntry);
 extern void DList_InsertHeadList(PDLIST_ENTRY listHead, PDLIST_ENTRY entry);
 extern void DList_AppendTailList(PDLIST_ENTRY listHead, PDLIST_ENTRY ListToAppend);
 extern int DList_RemoveEntryList(PDLIST_ENTRY listEntry);
 extern PDLIST_ENTRY DList_RemoveHeadList(PDLIST_ENTRY listHead);
+extern DLIST_FIND_RESULT DList_Find(PDLIST_ENTRY listHead, DLIST_MATCH_FUNCTION matchFunction, const void* matchContext, PDLIST_ENTRY* foundEntry);
+extern int DList_RemoveIf(PDLIST_ENTRY listHead, DLIST_CONDITION_FUNCTION conditionFunction, const void* conditionContext, DLIST_ENTRY_DESTROY_FUNCTION destroyFunction, const void* destroyContext);
+extern int DList_ForEach(PDLIST_ENTRY listHead, DLIST_ACTION_FUNCTION actionFunction, const void* actionContext);
```
anporumb, line 58: >DList_ForEach [](http://example.com/codeflow?start=11&length=13)

I don't think ForEach should take an "actionFunction". Keeping it very easy, it would just enumerate the DLIST_ENTRYs in the list. It is up to the user to perform "some" action on it.

This all sounds very cpluspus-ish


Oh wait, maybe it is just poetry. I saw "aCTIOn" and I thought... "it does something". But maybe it just enumerates.

./consider ACTION => ENUMERATE.

inc/c_util/doublylinkedlist.h
parth21999: 1) Yes. The current node will be removed and DList_RemoveIf will stop iterating over the list.

2) Yes. The current node will not be removed and DList_RemoveIf will continue iterating over the list.

devdoc/doublylinkedlist_requirements.md
```diff
@@ -20,13 +20,43 @@ typedef struct DLIST_ENTRY_TAG
...
+#define DLIST_CONDITION_FUNCTION_RESULT_VALUES \
+    DLIST_CONDITION_FUNCTION_TRUE, \
+    DLIST_CONDITION_FUNCTION_FALSE, \
+    DLIST_CONDITION_FUNCTION_ERROR
+
+MU_DEFINE_ENUM(DLIST_CONDITION_FUNCTION_RESULT, DLIST_CONDITION_FUNCTION_RESULT_VALUES);
+
+typedef DLIST_MATCH_FUNCTION_RESULT (*DLIST_MATCH_FUNCTION)(PDLIST_ENTRY listEntry, const void* matchContext);
+typedef DLIST_CONDITION_FUNCTION_RESULT (*DLIST_CONDITION_FUNCTION)(PDLIST_ENTRY listEntry, const void* conditionContext, bool* continueProcessing);
+typedef int (*DLIST_ENTRY_DESTROY_FUNCTION)(PDLIST_ENTRY listEntry, const void* destroyContext);
```
anporumb, line 46: >destroyContext [](http://example.com/codeflow?start=80&length=14)

what's the user supposed to do with destroyContext? What do I pass there?

Why does "removeIf" take a "DESTROY" (scary?...).

Remove if should do just that.... remove the DLIST_ENTRY from the list, hand it back to the user. The user can then do whatever he pleases with it... I suppose DESTROY spooked me here.

./consider rename DESTROY -> ONREMOVE. 

"destroy" means "totally free all used resources"... but this in no way does that. Or is mandated to do that.

inc/c_util/doublylinkedlist.h
parth21999: Agreed. Did it this way because singlylinkedlist had it this way. Removing.

devdoc/doublylinkedlist_requirements.md
```diff
@@ -20,13 +20,43 @@ typedef struct DLIST_ENTRY_TAG
...
+typedef int (*DLIST_ACTION_FUNCTION)(PDLIST_ENTRY listEntry, const void* actionContext, bool* continueProcessing);
+
 extern void DList_InitializeListHead(PDLIST_ENTRY listHead);
 extern int DList_IsListEmpty(const PDLIST_ENTRY listHead);
 extern void DList_InsertTailList(PDLIST_ENTRY listHead, PDLIST_ENTRY listEntry);
 extern void DList_InsertHeadList(PDLIST_ENTRY listHead, PDLIST_ENTRY entry);
 extern void DList_AppendTailList(PDLIST_ENTRY listHead, PDLIST_ENTRY ListToAppend);
 extern int DList_RemoveEntryList(PDLIST_ENTRY listEntry);
 extern PDLIST_ENTRY DList_RemoveHeadList(PDLIST_ENTRY listHead);
+extern DLIST_FIND_RESULT DList_Find(PDLIST_ENTRY listHead, DLIST_MATCH_FUNCTION matchFunction, const void* matchContext, PDLIST_ENTRY* foundEntry);
```
anporumb, line 56: >DList_Find [](http://example.com/codeflow?start=25&length=10)

How do I find **all** elements of the DLIST that match some condition?

Match condition doesn't return a "continueProcessing"... so it looks like it finds the first one, and then ... it stops.

If we love so much predicates, should have some way of continuing... ./consider

devdoc/doublylinkedlist_requirements.md
```diff
@@ -104,3 +134,48 @@ extern PDLIST_ENTRY DList_RemoveHeadList(PDLIST_ENTRY listHead);
 **SRS_DLIST_06_013: [** DList_RemoveHeadList shall return listHead if that's the only item in the list. **]**
 
 Note: The Flink & Blink of the returned PDLIST_ENTRY shall be undefined.
+
+### DList_Find
+```c
+MOCKABLE_FUNCTION(, DLIST_FIND_RESULT, DList_Find, PDLIST_ENTRY, listHead, DLIST_MATCH_FUNCTION, matchFunction, const void*, matchContext, PDLIST_ENTRY*, foundEntry);
+```
+
```
anporumb, line 142: traditionally we have one phrase that explains what the function does in non-requirement terms here.

anporumb: Here and everywhere so I don't repeat myself.

devdoc/doublylinkedlist_requirements.md
```diff
@@ -104,3 +134,48 @@ extern PDLIST_ENTRY DList_RemoveHeadList(PDLIST_ENTRY listHead);
...
+**S_R_S_DLIST_43_001: [** `DList_Find` shall call `matchFunction` on each entry in the list defined by `listHead` along with `matchContext`. **]**
```
anporumb, line 143: >S_R_S_DLIST_43_001 [](http://example.com/codeflow?start=2&length=18)

all values for all parameters are valid? What happens if foundEntry is `NULL`?

Or if listHead is `NULL`?

parth21999: The spec says: "no input error checking is provided for this set of APIs". 

So I didn't add these checks for the new helpers either. Undefined behavior if input is invalid.

src/doublylinkedlist.c
```diff
@@ -1,6 +1,13 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+#ifdef __cplusplus
+#include <cstddef>
+#else
+#include <stddef.h>
+#include <stdbool.h>
+#endif
```
anporumb, line 9: this only ever gets compiled as "C". so #include <stdboo.h> is enough

src/doublylinkedlist.c
```diff
@@ -105,3 +112,30 @@ void DList_InsertHeadList(PDLIST_ENTRY listHead, PDLIST_ENTRY entry)
...
+IMPLEMENT_MOCKABLE_FUNCTION(, DLIST_FIND_RESULT, DList_Find, PDLIST_ENTRY, listHead, DLIST_MATCH_FUNCTION, matchFunction, const void*, matchContext, PDLIST_ENTRY*, foundEntry)
+{
+    (void)listHead;
+    (void)matchFunction;
+    (void)matchContext;
+    (void)foundEntry;
+    return DLIST_FIND_RESULT_NOT_FOUND;
+}
+
+IMPLEMENT_MOCKABLE_FUNCTION(, int, DList_RemoveIf, PDLIST_ENTRY, listHead, DLIST_CONDITION_FUNCTION, conditionFunction, const void*, conditionContext, DLIST_ENTRY_DESTROY_FUNCTION, destroyFunction, const void*, destroyContext)
```
anporumb, line 125: >DLIST_ENTRY_DESTROY_FUNCTION, destroyFunction, [](http://example.com/codeflow?start=151&length=46)

I am thinking more and more that this parameter is not needed here.

anporumb: REmove means just that - remove. DEstroy means "deallocate used resources" and frankly, user might not want to do that.

What if user want just to move the DLIST_ENTRY from one list to another??

devdoc/doublylinkedlist_requirements.md
```diff
@@ -20,13 +20,43 @@ typedef struct DLIST_ENTRY_TAG
...
+    DLIST_CONDITION_FUNCTION_TRUE, \
+    DLIST_CONDITION_FUNCTION_FALSE, \
+    DLIST_CONDITION_FUNCTION_ERROR
+
+MU_DEFINE_ENUM(DLIST_CONDITION_FUNCTION_RESULT, DLIST_CONDITION_FUNCTION_RESULT_VALUES);
+
+typedef DLIST_MATCH_FUNCTION_RESULT (*DLIST_MATCH_FUNCTION)(PDLIST_ENTRY listEntry, const void* matchContext);
+typedef DLIST_CONDITION_FUNCTION_RESULT (*DLIST_CONDITION_FUNCTION)(PDLIST_ENTRY listEntry, const void* conditionContext, bool* continueProcessing);
+typedef int (*DLIST_ENTRY_DESTROY_FUNCTION)(PDLIST_ENTRY listEntry, const void* destroyContext);
+typedef int (*DLIST_ACTION_FUNCTION)(PDLIST_ENTRY listEntry, const void* actionContext, bool* continueProcessing);
```
anporumb, line 47: >int [](http://example.com/codeflow?start=8&length=3)

what's the meaning of this int?

anporumb (changes requested):
:clock1:

inc/c_util/doublylinkedlist.h
```diff
@@ -4,26 +4,61 @@
...
 MOCKABLE_FUNCTION(, void, DList_InitializeListHead, PDLIST_ENTRY, listHead);
 MOCKABLE_FUNCTION(, int, DList_IsListEmpty, const PDLIST_ENTRY, listHead);
 MOCKABLE_FUNCTION(, void, DList_InsertTailList, PDLIST_ENTRY, listHead, PDLIST_ENTRY, listEntry);
 MOCKABLE_FUNCTION(, void, DList_InsertHeadList, PDLIST_ENTRY, listHead, PDLIST_ENTRY, listEntry);
 MOCKABLE_FUNCTION(, void, DList_AppendTailList, PDLIST_ENTRY, listHead, PDLIST_ENTRY, ListToAppend);
 MOCKABLE_FUNCTION(, int, DList_RemoveEntryList, PDLIST_ENTRY, listEntry);
 MOCKABLE_FUNCTION(, PDLIST_ENTRY, DList_RemoveHeadList, PDLIST_ENTRY, listHead);
+MOCKABLE_FUNCTION(, DLIST_FIND_RESULT, DList_Find, PDLIST_ENTRY, listHead, DLIST_MATCH_FUNCTION, matchFunction, const void*, matchContext, PDLIST_ENTRY*, foundEntry);
+MOCKABLE_FUNCTION(, int, DList_RemoveIf, PDLIST_ENTRY, listHead, DLIST_CONDITION_FUNCTION, conditionFunction, const void*, conditionContext, DLIST_ENTRY_DESTROY_FUNCTION, destroyFunction, const void*, destroyContext);
+MOCKABLE_FUNCTION(, int, DList_ForEach, PDLIST_ENTRY, listHead, DLIST_ACTION_FUNCTION, actionFunction, const void*, actionContext);
```
dcristoloveanu, line 61: >ch, PDLIST_ENTRY, lis [](http://example.com/codeflow?start=36&length=21)

I guess based on the last in person discussion I'll have to see this again.

dcristoloveanu (changes requested):
:clock1:

devdoc/doublylinkedlist_requirements.md
parth21999: keeping it as action.

devdoc/doublylinkedlist_requirements.md
parth21999: Removing DList_RemoveIf in next iteration.

devdoc/doublylinkedlist_requirements.md
parth21999: Removing DList_Find in next iteration.

devdoc/doublylinkedlist_requirements.md
parth21999: Removing in next iteration.

devdoc/doublylinkedlist_requirements.md
parth21999: 0 on success, non-zero on failure. Added description in next iteration.

devdoc/doublylinkedlist_requirements.md
parth21999: Added description for DList_ForEach in next iteration.

src/doublylinkedlist.c
parth21999: Removing DList_RemoveIf in next iteration.

src/doublylinkedlist.c
parth21999: Removing in next iteration

devdoc/doublylinkedlist_requirements.md
```diff
@@ -20,13 +20,17 @@ typedef struct DLIST_ENTRY_TAG
     struct DLIST_ENTRY_TAG *Blink;
 } DLIST_ENTRY, *PDLIST_ENTRY;
 
+typedef int (*DLIST_ACTION_FUNCTION)(PDLIST_ENTRY listEntry, void* actionContext, bool* continueProcessing);
+
 extern void DList_InitializeListHead(PDLIST_ENTRY listHead);
```
anporumb, line 25: >extern [](http://example.com/codeflow?start=0&length=6)

all C function have by default "extern" linkage, so it can be omitted.

./consider

devdoc/doublylinkedlist_requirements.md
```diff
@@ -20,13 +20,17 @@ typedef struct DLIST_ENTRY_TAG
...
 extern int DList_IsListEmpty(const PDLIST_ENTRY listHead);
 extern void DList_InsertTailList(PDLIST_ENTRY listHead, PDLIST_ENTRY listEntry);
```
anporumb, line 27: >DList_InsertTailList [](http://example.com/codeflow?start=12&length=20)

shouldn't all thse be copy&paste from header? Together with MOICKABLE_FUNCTION?

anporumb:
> #define CONTAINING_RECORD(address, type, field) ((type *)((char *)(address) - offsetof(type,field)))

not defined here.

---
Refers to: devdoc/doublylinkedlist_requirements.md:39 in d26d2d4. [](commit_id = d26d2d4d93ab15ba3f827d1f317db57684b3c9bf, deletion_comment = False)

anporumb (approved):
:shipit:

devdoc/doublylinkedlist_requirements.md
parth21999: copy pasting from header in next iteration.

parth21999:
> #define CONTAINING_RECORD(address, type, field) ((type *)((char *)(address) - offsetof(type,field)))

Original author included this. Maybe to hint to the reader on how to obtain address of struct from PDLIST_ENTRY. Leaving as is.

---
In reply to: [1155852922](https://github.com/Azure/c-util/pull/181#issuecomment-1155852922) [](http://example.com/codeflow?ancestors=1155852922,1155852922)

---
Refers to: devdoc/doublylinkedlist_requirements.md:39 in d26d2d4. [](commit_id = d26d2d4d93ab15ba3f827d1f317db57684b3c9bf, deletion_comment = False)

dcristoloveanu:
> #define CONTAINING_RECORD(address, type, field) ((type *)((char *)(address) - offsetof(type,field)))

I wanted to reply to the comment about this but could not! Because I think the original comment from Andrei was deleted.
Anyway, here is what I said:

"I think at some point CONTAINING_RECORD was in DLIST.
But as I recall we moved it out as it is more generic.
I suggest replacing this with a note that says that typically DLIST_ENTRY is used as part of a structure and the obtaining of the parent structure can be done with CONTAINING_RECORD. If you want to go the extra mile you can even give an example."

---
Refers to: devdoc/doublylinkedlist_requirements.md:39 in 341cf36. [](commit_id = 341cf36ad590e004ad718b5b09f4880e75f5eb69, deletion_comment = False)

devdoc/doublylinkedlist_requirements.md
```diff
@@ -104,3 +108,17 @@ extern PDLIST_ENTRY DList_RemoveHeadList(PDLIST_ENTRY listHead);
 **SRS_DLIST_06_013: [** DList_RemoveHeadList shall return listHead if that's the only item in the list. **]**
 
 Note: The Flink & Blink of the returned PDLIST_ENTRY shall be undefined.
+
+### DList_ForEach
+```c
+MOCKABLE_FUNCTION(, int, DList_ForEach, PDLIST_ENTRY, listHead, DLIST_ACTION_FUNCTION, actionFunction, void*, actionContext);
```
dcristoloveanu, line 114: >actionFunction [](http://example.com/codeflow?start=87&length=14)

This should be checked for NULL.

dcristoloveanu, line 114: >listHead [](http://example.com/codeflow?start=54&length=8)

This should also be checked for NULL.
Frankly like I mentioned in my other comment all other functions should be consistent with how we write code and shall check their arguments.

dcristoloveanu: The new code at least should have the checks since we're writing it now.
You can create a separate task for the existing functions to check their args.

dcristoloveanu: Since this comment was already ignored once, I'll still wait for it until it's done :-)

parth21999: Added check.

parth21999: Sorry, didn't mean to ignore the comment. I was just trying to follow the precedent in the module. 

Added arg checks for new code. 

Added task in backlog to add checks everywhere: [https://msazure.visualstudio.com/DefaultCollection/One/_backlogs/backlog/Azure%20Messaging%20Store/Backlog%20items/?workitem=12862713](https://msazure.visualstudio.com/DefaultCollection/One/_backlogs/backlog/Azure%20Messaging%20Store/Backlog%20items/?workitem=12862713)

parth21999:
> #define CONTAINING_RECORD(address, type, field) ((type *)((char *)(address) - offsetof(type,field)))

Understood. Replacing this with example.

---
In reply to: [1155945393](https://github.com/Azure/c-util/pull/181#issuecomment-1155945393) [](http://example.com/codeflow?ancestors=1155945393,1155945393)

---
Refers to: devdoc/doublylinkedlist_requirements.md:39 in 341cf36. [](commit_id = 341cf36ad590e004ad718b5b09f4880e75f5eb69, deletion_comment = False)

PR #195, opened 2022-08-08
Harid/cutil interlocked hl int tests

haridamineni:


tests/interlocked_hl_int/interlocked_hl_int.c
```diff
@@ -0,0 +1,287 @@
...
+{
+}
+
+TEST_FUNCTION_INITIALIZE(function_initialize)
+{
+}
+
+TEST_FUNCTION_CLEANUP(function_cleanup)
+{
+}
```
dcristoloveanu, line 36: Can be removed if there is no setup needed

tests/interlocked_hl_int/interlocked_hl_int.c
```diff
@@ -0,0 +1,287 @@
...
+TEST_FUNCTION(interlocked_hl_decrement_and_wake_operates_successfully)
+{
+    // + have a helper thread which waits for decremented value
+    // + main thread creates helper thread and decrements value
+    // + helper thread wakes up when the global address value is decremented by 1
+    // arrange
+    globalValue = 10;
+    THREAD_HANDLE helper_thread_handle;
+    THREADAPI_RESULT thread_result = ThreadAPI_Create(&helper_thread_handle, decrement_and_wake_helper_thread_function, NULL);
+    ASSERT_ARE_EQUAL(THREADAPI_RESULT, THREADAPI_OK, thread_result);
```
dcristoloveanu, line 62: Can have the assert in the same line with the call, so you can remove the thread_result variable alltogether

tests/interlocked_hl_int/interlocked_hl_int.c
```diff
@@ -0,0 +1,287 @@
...
+
+    // act
+    //sleep so that the helper thread can go into wait mode
+    ThreadAPI_Sleep(5000);
+    INTERLOCKED_HL_RESULT result = InterlockedHL_DecrementAndWake(&globalValue);
+    ASSERT_ARE_EQUAL(INTERLOCKED_HL_RESULT, INTERLOCKED_HL_OK, result);
```
dcristoloveanu, line 68: Can combine assert and the call also.

tests/interlocked_hl_int/interlocked_hl_int.c
```diff
@@ -0,0 +1,287 @@
...
+    ASSERT_ARE_EQUAL(THREADAPI_RESULT, THREADAPI_OK, ThreadAPI_Join(helper_thread_handle, &return_code));
+    ASSERT_ARE_EQUAL(int, 0, return_code);
+}
+
+/*
+Tests:
+InterlockedHL_WaitForValue
+InterlockedHL_SetAndWakeAll
+*/
+static int set_and_wakeall_helper_thread_function(void* context)
```
dcristoloveanu, line 82: >wakeall [](http://example.com/codeflow?start=19&length=7)

wake_all

tests/interlocked_hl_int/interlocked_hl_int.c
dcristoloveanu: Here and below

tests/interlocked_hl_int/interlocked_hl_int.c
```diff
@@ -0,0 +1,287 @@
...
+
+TEST_FUNCTION(interlocked_hl_set_and_wakeall_operates_successfully)
+{
+    // + create 10 helper threads which wait on a value
+    // + main thread creates helper threads and sets value for the helper threads to wake up
+    // + main thread joins on the helper threads and ensures that all threads wake up and terminate
+    // arrange
+    globalValue = 10;
+
+    enum { NUMBER_OF_HELPER_THREADS = 10 };
```
dcristoloveanu, line 98: > enum { NUMBER_OF_HELPER_THREADS = 10 }; [](http://example.com/codeflow?start=3&length=40)

Just wondering would a const size_t do the same? I am genuinely asking so that I know myself.

tests/interlocked_hl_int/interlocked_hl_int.c
```diff
@@ -0,0 +1,287 @@
...
+    ASSERT_ARE_EQUAL(INTERLOCKED_HL_RESULT, INTERLOCKED_HL_OK, result);
+    ASSERT_ARE_EQUAL(int64_t, 55, original_addend);
+    ASSERT_ARE_EQUAL(int64_t, 75, addend);
+}
+
+/*
+Tests:
+InterlockedHL_CompareExchangeIf
+*/
+bool helper_int32_compare_function(int32_t target, int32_t exchange)
```
dcristoloveanu, line 233: >helper_int32_compare_function [](http://example.com/codeflow?start=5&length=29)

Likely static

tests/interlocked_hl_int/interlocked_hl_int.c
```diff
@@ -0,0 +1,287 @@
...
+    ASSERT_ARE_EQUAL(INTERLOCKED_HL_RESULT, INTERLOCKED_HL_OK, result);
+    ASSERT_ARE_EQUAL(int32_t, 60, original_target);
+    ASSERT_ARE_EQUAL(int32_t, 88, target);
+}
+
+/*
+Tests:
+InterlockedHL_CompareExchange64If
+*/
+bool helper_int64_compare_function(int64_t target, int64_t exchange)
```
dcristoloveanu, line 262: >helper_int64_compare_function [](http://example.com/codeflow?start=5&length=29)

static

dcristoloveanu (approved):
:shipit:

tests/interlocked_hl_int/interlocked_hl_int.c
haridamineni: i tried to use const size_t variable. but, the compiler complained when used it for an array as in array[const_variable]. C does not allow it, looks like. so, had to take the enum route.

PR #200, opened 2022-08-23
[Specs Change] Adding builder for CONSTBUFFER 

ACKPaul:
Currently, it takes two malloc calls to get the final CONSTBUFFER_HANDLE. we want to reduce it to one malloc call so that we can allocate a buffer, write to it, and eventually get a CONSTBUFFER_HANDLE.

### DIFF
--- a/devdoc/constbuffer_requirements.md
+++ b/devdoc/constbuffer_requirements.md
@@ -363,7 +363,7 @@ CONSTBUFFER_FROM_BUFFER_RESULT CONSTBUFFER_from_buffer(const unsigned char* sour
 **SRS_CONSTBUFFER_02_073: [** If there are any failures then shall fail and return `CONSTBUFFER_FROM_BUFFER_RESULT_ERROR`. **]**
 
 
-### CONSTBUFFER_create_writiable_handle
+### CONSTBUFFER_create_writable_handle
 
 ```c
  CONSTBUFFER_WRITABLE_HANDLE CONSTBUFFER_create_writable_handle (uint32_t size)
@@ -392,7 +392,7 @@ CONSTBUFFER_FROM_BUFFER_RESULT CONSTBUFFER_from_buffer(const unsigned char* sour
 
 **S_RS_CONSTBUFFER_51_006: [** If `constbufferHandle` is `NULL`, then `CONSTBUFFER_get_writable_buffer` shall fail and return `NULL`. **]**
 
-**S_RS_CONSTBUFFER_51_007: [** `CONSTBUFFER_get_writable_buffer` shall succeed and return a handle to the non-CONST buffer of `constbufferWritableHandle`. **]**
+**S_RS_CONSTBUFFER_51_007: [** `CONSTBUFFER_get_writable_buffer` shall succeed and returns a pointer to the non-CONST buffer of `constbufferWritableHandle`. **]**
 
 
 
--- a/devdoc/constbuffer_requirements.md
+++ b/devdoc/constbuffer_requirements.md
@@ -104,8 +104,6 @@ MOCKABLE_FUNCTION(, CONSTBUFFER_HANDLE, CONSTBUFFER_Create, const unsigned char*
 
 **SRS_CONSTBUFFER_02_005: [** The non-NULL handle returned by `CONSTBUFFER_Create` shall have its ref count set to "1". **]** 
 
-**S_RS_CONSTBUFFER_51_014: [** `CONSTBUFFER_Create` shall set the state of `CONSTBUFFER_HANDLE` to `CONSTBUFFER_WRITIABLE_STATE_SEAL`. **]**
-
 ### CONSTBUFFER_CreateFromBuffer
 
 ```c
@@ -121,8 +119,6 @@ MOCKABLE_FUNCTION(, CONSTBUFFER_HANDLE, CONSTBUFFER_CreateFromBuffer, BUFFER_HAN
 **SRS_CONSTBUFFER_02_009: [** Otherwise, `CONSTBUFFER_CreateFromBuffer` shall return a non-NULL handle. **]**
 
 **SRS_CONSTBUFFER_02_010: [** The non-NULL handle returned by `CONSTBUFFER_CreateFromBuffer` shall have its ref count set to "1". **]** 
-
-**S_RS_CONSTBUFFER_51_015: [** `CONSTBUFFER_CreateFromBuffer` shall set the state of `CONSTBUFFER_HANDLE` to `CONSTBUFFER_WRITIABLE_STATE_SEAL`. **]**
 
 ### CONSTBUFFER_CreateWithMoveMemory
 
@@ -140,8 +136,6 @@ The memory is assumed to be freeable by a call to `free`.
 **SRS_CONSTBUFFER_01_002: [** `CONSTBUFFER_CreateWithMoveMemory` shall store the `source` and `size` and return a non-NULL handle to the newly created const buffer. **]**
 
 **SRS_CONSTBUFFER_01_003: [** The non-NULL handle returned by `CONSTBUFFER_CreateWithMoveMemory` shall have its ref count set to "1". **]**
-
-**S_RS_CONSTBUFFER_51_016: [** `CONSTBUFFER_CreateWithMoveMemory` shall set the state of `CONSTBUFFER_HANDLE` to `CONSTBUFFER_WRITIABLE_STATE_SEAL`. **]**
 
 **SRS_CONSTBUFFER_01_005: [** If any error occurs, `CONSTBUFFER_CreateWithMoveMemory` shall fail and return NULL. **]**
 
@@ -168,8 +162,6 @@ The memory has to be free by calling the custom free function passed as argument
 
 **SRS_CONSTBUFFER_01_010: [** The non-NULL handle returned by `CONSTBUFFER_CreateWithCustomFree` shall have its ref count set to 1. **]**
 
-**S_RS_CONSTBUFFER_51_017: [** `CONSTBUFFER_CreateWithCustomFree` shall set the state of `CONSTBUFFER_HANDLE` to `CONSTBUFFER_WRITIABLE_STATE_SEAL`. **]**
-
 **SRS_CONSTBUFFER_01_011: [** If any error occurs, `CONSTBUFFER_CreateWithMoveMemory` shall fail and return NULL. **]**
 
 ### CONSTBUFFER_CreateFromOffsetAndSize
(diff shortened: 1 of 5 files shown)


devdoc/constbuffer_requirements.md
```diff
@@ -78,7 +78,13 @@ MOCKABLE_INTERFACE(constbuffer,
 
     FUNCTION(, CONSTBUFFER_TO_FIXED_SIZE_BUFFER_RESULT, CONSTBUFFER_to_fixed_size_buffer, CONSTBUFFER_HANDLE, source, unsigned char*, destination, uint32_t, destination_size, uint32_t*, serialized_size),
 
-    FUNCTION(, CONSTBUFFER_FROM_BUFFER_RESULT, CONSTBUFFER_from_buffer, const unsigned char*, source, uint32_t, size, uint32_t*, consumed, CONSTBUFFER_HANDLE*, destination)
+    FUNCTION(, CONSTBUFFER_FROM_BUFFER_RESULT, CONSTBUFFER_from_buffer, const unsigned char*, source, uint32_t, size, uint32_t*, consumed, CONSTBUFFER_HANDLE*, destination),
+
+    FUNCTION(, CONSTBUFFER_HANDLE, CONSTBUFFER_create_writable_handle, uint32_t, size),
```
dcristoloveanu, line 83: >CONSTBUFFER_HANDLE [](http://example.com/codeflow?start=15&length=18)

Looking at this I am wondering if we could have something to prohibit accidentally passing a writeable const buffer handle to other APIs (like passing a writeable CONST BUFFER to a constbuffer array for example)

Think something like a CONSTBUFFER_WRITEABLE_HANDLE and the compiler would not accept that when passed to constbuffer_array APIs for example.
I don't have a solution right now and here (typedef CONSTBUFFER_HANDLE CONSTBUFFER_WRITEABLE_HANDLE will not make the compiler cry, but would be a nice thing to have. 
Maybe something where CONSTBUFFER_HANDLE has a different qualifier of sorts ...?
Just thinking our loud

devdoc/constbuffer_requirements.md
```diff
@@ -98,6 +104,8 @@ MOCKABLE_FUNCTION(, CONSTBUFFER_HANDLE, CONSTBUFFER_Create, const unsigned char*
 
 **SRS_CONSTBUFFER_02_005: [** The non-NULL handle returned by `CONSTBUFFER_Create` shall have its ref count set to "1". **]** 
 
+**SRS_CONSTBUFFER_51_015: [** Shall set the state of `CONSTBUFFER_HANDLE` to `CONSTBUFFER_WRITIABLE_STATE_SEAL`. **]**
```
dcristoloveanu, line 107: >Shall [](http://example.com/codeflow?start=30&length=5)

We put the function name in the spec:
`CONSTBUFFER_Create` shall set the state ...

devdoc/constbuffer_requirements.md
```diff
@@ -340,6 +372,8 @@ CONSTBUFFER_FROM_BUFFER_RESULT CONSTBUFFER_from_buffer(const unsigned char* sour
 
 **SRS_CONSTBUFFER_02_065: [** If `destination` is `NULL` then `CONSTBUFFER_from_buffer` shall fail and return `CONSTBUFFER_FROM_BUFFER_RESULT_INVALID_ARG`. **]**
 
+**SRS_CONSTBUFFER_51_027: [** If state of `destination` is `CONSTBUFFER_WRITABLE_STATE_OPEN`, shall fail and return `CONSTBUFFER_FROM_BUFFER_RESULT_INVALID_ARG`. **]**
```
dcristoloveanu, line 375: destination is the out argument, this function is actually creating the buffer placed in destination, so we should definitely not look at it.

devdoc/constbuffer_requirements.md
dcristoloveanu: Here and everywhere

devdoc/constbuffer_requirements.md
```diff
@@ -354,4 +388,62 @@ CONSTBUFFER_FROM_BUFFER_RESULT CONSTBUFFER_from_buffer(const unsigned char* sour
...
 **SRS_CONSTBUFFER_02_073: [** If there are any failures then shall fail and return `CONSTBUFFER_FROM_BUFFER_RESULT_ERROR`. **]**
+
+
+### CONSTBUFFER_create_writiable_handle
+
+```c
+ CONSTBUFFER_HANDLE CONSTBUFFER_create_writable_handle (uint32_t size)
+```
+
+`CONSTBUFFER_create_writiable_handle` construct a new `CONSTBUFFER_HANDLE` of size `size` and state `CONSTBUFFER_WRITABLE_STATE_OPEN`.`CONSTBUFFER_HANDLE`supports two state `CONSTBUFFER_WRITABLE_STATE_SEAL` and `CONSTBUFFER_WRITABLE_STATE_OPEN`.If the state is `CONSTBUFFER_WRITABLE_STATE_OPEN` then buffers of `CONSTBUFFER_HANDLE` can be overridden (the same buffer can be reused to fill).
```
dcristoloveanu, line 402: >.If [](http://example.com/codeflow?start=244&length=3)

Nitpick: space after the dot. Allows for somewhat better formatting.

devdoc/constbuffer_requirements.md
```diff
@@ -354,4 +388,62 @@ CONSTBUFFER_FROM_BUFFER_RESULT CONSTBUFFER_from_buffer(const unsigned char* sour
...
+
+**SRS_CONSTBUFFER_51_002: [** `CONSTBUFFER_create_writiable_handle` shall allocate memory for the `CONSTBUFFER_HANDLE` **]**
+
+**SRS_CONSTBUFFER_51_003: [** If any error occurs, `CONSTBUFFER_create_writiable_handle` shall fail and return NULL. **]**
+
+**SRS_CONSTBUFFER_51_004: [** `CONSTBUFFER_create_writiable_handle` shall set the ref count of the newly created `CONSTBUFFER_HANDLE` to 1. **]**
+
+**SRS_CONSTBUFFER_51_005: [** `CONSTBUFFER_create_writiable_handle` shall set the state of the newly created `CONSTBUFFER_HANDLE` to `CONSTBUFFER_WRITABLE_STATE_OPEN`. **]**
+
+**SRS_CONSTBUFFER_51_006: [** `CONSTBUFFER_create_writiable_handle` shall store the `storage` and `size` to the newly created const buffer. **]**
```
dcristoloveanu, line 414: >`storage` [](http://example.com/codeflow?start=84&length=9)

storage doe snot exist here in the specs. I suspect you are referring that internally the const buffer should set its alias to wherever the bytes are allocated. But I think you don't actually have to spec that at all, it is an implementation detail. Tests will fail if we don't do that :-).

ACKPaul: Removed .

devdoc/constbuffer_requirements.md
ACKPaul: Added prefix all the places .

devdoc/constbuffer_requirements.md
ACKPaul: Got it. Fixed.

devdoc/constbuffer_requirements.md
ACKPaul: Fixed.

devdoc/constbuffer_requirements.md
ACKPaul: Created two structures CONSTBUFFER_HANDLE_DATA_TAG and CONSTBUFFER_WRITABLE_HANDLE_DATA_TAG with help of macro `MU_DEFINE_STRUCT` with the same fields. Now on compile time if the user by mistake passes one structure type instead of the other then the build will fail, and no run-time check is required.

devdoc/constbuffer_requirements.md
```diff
@@ -355,3 +361,49 @@ CONSTBUFFER_FROM_BUFFER_RESULT CONSTBUFFER_from_buffer(const unsigned char* sour
 **SRS_CONSTBUFFER_02_072: [** `CONSTBUFFER_from_buffer` shall succeed, write in `consumed` the total number of consumed bytes from `source`, write in `destination` the constructed `CONSTBUFFER_HANDLE` and return `CONSTBUFFER_FROM_BUFFER_RESULT_OK`. **]**
 
 **SRS_CONSTBUFFER_02_073: [** If there are any failures then shall fail and return `CONSTBUFFER_FROM_BUFFER_RESULT_ERROR`. **]**
+
+
+### CONSTBUFFER_create_writiable_handle
```
dcristoloveanu, line 366: >writiable [](http://example.com/codeflow?start=23&length=9)

typo: writable

devdoc/constbuffer_requirements.md
```diff
@@ -355,3 +361,49 @@ CONSTBUFFER_FROM_BUFFER_RESULT CONSTBUFFER_from_buffer(const unsigned char* sour
...
+**S_RS_CONSTBUFFER_51_005: [** `CONSTBUFFER_create_writiable_handle` shall succeed and return a non-`NULL` `CONSTBUFFER_WRITABLE_HANDLE`. **]**
+
+
+### CONSTBUFFER_get_writable_buffer
+
+```c
+ unsigned char* CONSTBUFFER_get_writable_buffer(CONSTBUFFER_WRITABLE_HANDLE constbufferWritableHandle)
+```
+
+`CONSTBUFFER_get_writable_buffer` return the handle of writable buffer.
```
dcristoloveanu, line 391: >return [](http://example.com/codeflow?start=34&length=6)

returns

devdoc/constbuffer_requirements.md
```diff
@@ -355,3 +361,49 @@ CONSTBUFFER_FROM_BUFFER_RESULT CONSTBUFFER_from_buffer(const unsigned char* sour
...
+
+**S_RS_CONSTBUFFER_51_006: [** If `constbufferHandle` is `NULL`, then `CONSTBUFFER_get_writable_buffer` shall fail and return `NULL`. **]**
+
+**S_RS_CONSTBUFFER_51_007: [** `CONSTBUFFER_get_writable_buffer` shall succeed and return a handle to the non-CONST buffer of `constbufferWritableHandle`. **]**
```
dcristoloveanu, line 395: >handle [](http://example.com/codeflow?start=92&length=6)

pointer

dcristoloveanu (approved):
:shipit:

devdoc/constbuffer_requirements.md
ACKPaul: Fixed.

devdoc/constbuffer_requirements.md
ACKPaul: Fixed.

PR #201, opened 2022-08-26
[Code and Test change] An optimization in CONSTBUFFER_CreateFromOffsetAndSize.

RohitG28:
An optimization in CONSTBUFER_CreateFromOffsetAndSize where we return the original handle as is if the offset is 0 and size is equal to handle's size.

### DIFF
--- a/src/constbuffer.c
+++ b/src/constbuffer.c
@@ -221,6 +221,12 @@ IMPLEMENT_MOCKABLE_FUNCTION(, CONSTBUFFER_HANDLE, CONSTBUFFER_CreateFromOffsetAn
             handle, offset, size);
         result = NULL;
     }
+    /*Codes_SRS_CONSTBUFFER_28_001: [ If offset is 0 and size is equal to handle's size then CONSTBUFFER_CreateFromOffsetAndSize shall increment the reference count of handle and return handle. ]*/
+    else if (offset == 0 && size == handle->alias.size)
+    {
+        (void)interlocked_increment(&handle->count);
+        result = handle;
+    }
     else
     {
         /*Codes_SRS_CONSTBUFFER_02_028: [ CONSTBUFFER_CreateFromOffsetAndSize shall allocate memory for a new CONSTBUFFER_HANDLE's content. ]*/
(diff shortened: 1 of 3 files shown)


src/constbuffer.c
```diff
@@ -221,6 +221,12 @@ IMPLEMENT_MOCKABLE_FUNCTION(, CONSTBUFFER_HANDLE, CONSTBUFFER_CreateFromOffsetAn
             handle, offset, size);
         result = NULL;
     }
+    /*Codes_SRS_CONSTBUFFER_28_001: [ If offset is 0 and size is equal to handle's size then CONSTBUFFER_CreateFromOffsetAndSize shall increment the reference count of handle and return handle. ]*/
+    else if (offset == 0 && size == handle->alias.size)
+    {
+        (void)interlocked_increment(&handle->count);
```
dcristoloveanu, line 227: >interlocked_increment [](http://example.com/codeflow?start=14&length=21)

REFCOUNT (which is the mechanism being used for ref counting here by ConstBuffer) has INC_REF to abstract the actual incrementing of the ref count. That's what should be used for it.

The same applies to ConstBuffer_IncRef actually.
And because we should not have the INC_REF done in 2 places (to avoid having to fix something in multiple places if possible), sonsider extracting an internal method:

static CONSTBUFFER_IncRefInternal, which should do the INC_REF.
Then the 2 places should just call this.

Similar to how we have extracted CONSTBUFFER_DecRef_internal.

dcristoloveanu (changes requested):
:clock1:

dcristoloveanu:
I only had one comment, not a big item, but would like to see that addressed.

src/constbuffer.c
dcristoloveanu: You are right, it turns out I just THOUGHT CONSTBUFFER uses REFCOUNT, it does use poor man's ref counting instead :-).
So we can have it as is now, we don't need to switch it right now. Another task for another day.

dcristoloveanu (approved):
:shipit:

PR #204, opened 2022-09-01
Add rc string sprintf

M-iceberg:


devdoc/rc_string_requirements.md
```diff
@@ -72,6 +72,33 @@ MOCKABLE_FUNCTION(, THANDLE(RC_STRING), rc_string_create, const char*, string);
...
+
+```c
+THANDLE(RC_STRING) rc_string_create_with_format(const char* format, ...);
+```
+
+`rc_string_create_with_format` creates a new ref counted string by using the format convention as in sprintf.
+
+If `format` is `NULL`, `rc_string_create_with_format` shall fail and return `NULL`. 
+
+Otherwise, `rc_string_create_with_format` shall determine the total number of characters written using the variadic arguments.
```
dcristoloveanu, line 85: >variadic [](http://example.com/codeflow?start=107&length=8)

this word actually does not appear in the C99 or C11 standard.
The standard way of talking about these is variable number of arguments

devdoc/rc_string_requirements.md
```diff
@@ -72,6 +72,33 @@ MOCKABLE_FUNCTION(, THANDLE(RC_STRING), rc_string_create, const char*, string);
...
+
+If `format` is `NULL`, `rc_string_create_with_format` shall fail and return `NULL`. 
+
+Otherwise, `rc_string_create_with_format` shall determine the total number of characters written using the variadic arguments. 
+
+`rc_string_create_with_format` shall allocate memory for the formatted string including the `NULL` terminator.
+
+`rc_string_create_with_format` shall copy the formatted string result including the `NULL` terminator into `string`. 
+
+`rc_string_create_with_format` shall allocate memory for  the `THANDLE(RC_STRING)`, ensuring all the bytes in `string` can be copied (including the zero terminator).
```
dcristoloveanu, line 91: > allocate memory for  the `THAN [](http://example.com/codeflow?start=36&length=31)

So, why do we need 2 allocations? I would like this to be done with 1 allocation. I think we know how to do that all the time:

get the length of the needed string
allocate the memory for the THANDLE(RC_STRING) AND the string that it needs to contain
fill in the alias pointer in the THANDLE(RC_STRING)
fill in the bytes of the string by using vsnprintf

dcristoloveanu (changes requested):
:clock1:

devdoc/rc_string_requirements.md
```diff
@@ -72,6 +72,26 @@ MOCKABLE_FUNCTION(, THANDLE(RC_STRING), rc_string_create, const char*, string);
...
+THANDLE(RC_STRING) rc_string_create_with_format(const char* format, ...);
+```
+
+`rc_string_create_with_format` creates a new ref counted string by using the format convention as in sprintf.
+
+If `format` is `NULL`, `rc_string_create_with_format` shall fail and return `NULL`. 
+
+Otherwise, `rc_string_create_with_format` shall determine the total number of characters written using the variable number of arguments.  
+
+`rc_string_create_with_format` shall allocate memory for the `THANDLE(RC_STRING)`.
```
dcristoloveanu, line 87: >allocate memory for the `THANDLE(RC_STRING)`. [](http://example.com/codeflow?start=37&length=45)

I think we should specify here that we're allocating memory for THANDLE(RC_STRING) and the number of bytes needed for the resulting formatted string.

dcristoloveanu (approved):
:shipit:

PR #205, opened 2022-09-06
modify rc_string_sprintf unit test

M-iceberg:


tests/rc_string_ut/rc_string_mocked.c
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#include <stdio.h>
+
+#define vsnprintf mocked_vsnprintf
+
+extern int mocked_vsnprintf(char* s, size_t n, const char* format, ...);
```
dcristoloveanu, line 8: >extern [](http://example.com/codeflow?start=0&length=6)

Likely not needed

dcristoloveanu, line 8: > ...); [](http://example.com/codeflow?start=66&length=6)

The signature of vsnprintf per C99 is:
int vsnprintf(char * restrict s, size_t n,
const char * restrict format,
va_list arg);

devdoc/rc_string_requirements.md
```diff
@@ -80,17 +80,21 @@ THANDLE(RC_STRING) rc_string_create_with_format(const char* format, ...);
...
+**SRS_RC_STRING_07_003: [** If `vsnprintf` failed to determine the total number of characters written, `rc_string_create_with_format` shall fail and return `NULL`. **]**
 
-`rc_string_create_with_format` shall fill in the bytes of the string by using `vsnprintf`.
+**SRS_RC_STRING_07_004: [** `rc_string_create_with_format` shall allocate memory for the `THANDLE(RC_STRING)` and the number of bytes for the resulting formatted string. **]**
 
-`rc_string_create_with_format` shall succeed and return a non-`NULL` handle. 
+**SRS_RC_STRING_07_005: [** `rc_string_create_with_format` shall fill in the bytes of the string by using `vsnprintf`. **]**
 
-If any error occurs, `rc_string_create_with_format` shall fail and return `NULL`.
+**SRS_RC_STRING_07_006: [** If `vsnprintf` failed to get the resulting formatted string, `rc_string_create_with_format` shall fail and return `NULL`. **]**
```
dcristoloveanu, line 93: >get [](http://example.com/codeflow?start=53&length=3)

I would say "construct"

src/rc_string.c
```diff
@@ -108,11 +109,61 @@ IMPLEMENT_MOCKABLE_FUNCTION(, THANDLE(RC_STRING), rc_string_create, const char*,
...
+    {
+        /*Codes_SRS_RC_STRING_07_001: [ If format is NULL, rc_string_create_with_format shall fail and return NULL. ]*/
+        LogError("Invalid arguments: const char* format=%s", MU_P_OR_NULL(format));
+    }
+    else
+    {
+        /*Codes_SRS_RC_STRING_07_002: [ Otherwise, `rc_string_create_with_format` shall determine the total number of characters written using the variable number of arguments. ]*/
+        va_list args;
+        va_start(args, format);
+        size_t string_length_with_terminator = vsnprintf(NULL, 0, format, args) + 1;
```
dcristoloveanu, line 127: >vsnprintf [](http://example.com/codeflow?start=47&length=9)

So, what if vsnprintf return -2?
Section 7.19.6.12 of C99 says this can happen.

The vsnprintf function returns the number of characters that would have been written
had n been sufficiently large, not counting the terminating null character, *oraneg ative
value if an encoding error occurred*. 

Would like to see a test for it :-)

tests/rc_string_ut/rc_string_ut.c
```diff
@@ -13,6 +15,11 @@
 #define ENABLE_MOCKS
 #include "c_pal/gballoc_hl.h"
 #include "c_pal/gballoc_hl_redirect.h"
+
+#include "umock_c/umock_c_prod.h"
```
dcristoloveanu, line 19: A wild umock_c_prod include has appeared!
Why do we need to do this here in the middle of the #define ENABLE_MOCKS
#undef ENABLE_MOCKS

?

tests/rc_string_ut/rc_string_ut.c
```diff
@@ -54,6 +66,17 @@ TEST_SUITE_INITIALIZE(suite_initialize)
...
     ASSERT_ARE_EQUAL(int, 0, umocktypes_charptr_register_types());
 
+    REGISTER_GLOBAL_MOCK_HOOK(mocked_vsnprintf, my_vsnprintf);
+    REGISTER_GLOBAL_MOCK_FAIL_RETURN(mocked_vsnprintf, -1);
+
+#ifdef WIN32
+    // The reason this is with a WIN32 block is because 
+    // in linux the va_list item is not a char*, but a struct
+    // of size 24 bytes.  It is made to be opaque so we are unable
+    // to set up the mock alias for this type.
```
dcristoloveanu, line 76: Interesting, so what if for Linux we're making this alias to a struct of 24 bytes (opaque as it is I assume we can see what it is actually behind in the code)?

tests/rc_string_ut/rc_string_ut.c
```diff
@@ -54,6 +66,17 @@ TEST_SUITE_INITIALIZE(suite_initialize)
     ASSERT_ARE_EQUAL(int, 0, umock_c_init(on_umock_c_error));
     ASSERT_ARE_EQUAL(int, 0, umocktypes_charptr_register_types());
 
+    REGISTER_GLOBAL_MOCK_HOOK(mocked_vsnprintf, my_vsnprintf);
+    REGISTER_GLOBAL_MOCK_FAIL_RETURN(mocked_vsnprintf, -1);
+
+#ifdef WIN32
+    // The reason this is with a WIN32 block is because 
+    // in linux the va_list item is not a char*, but a struct
```
dcristoloveanu, line 74: >char [](http://example.com/codeflow?start=42&length=4)

likely meant void*

src/rc_string.c
```diff
@@ -108,11 +109,61 @@ IMPLEMENT_MOCKABLE_FUNCTION(, THANDLE(RC_STRING), rc_string_create, const char*,
...
+        
+        if (string_length_with_terminator <= 0)
+        {
+            /*Codes_SRS_RC_STRING_07_003: [ If `vsnprintf` failed to determine the total number of characters written, `rc_string_create_with_format` shall fail and return `NULL`. ]*/
+            LogError("vsnprintf failed to determine the total number of characters written.");
+        }
+        else
+        {
+            /*Codes_SRS_RC_STRING_07_004: [ `rc_string_create_with_format` shall allocate memory for the `THANDLE(RC_STRING)`and the number of bytes for the resulting formatted string. ]*/
+            THANDLE(RC_STRING) temp_result = THANDLE_MALLOC_WITH_EXTRA_SIZE(RC_STRING)(rc_string_dispose, sizeof(RC_STRING_INTERNAL) - sizeof(RC_STRING) + string_length_with_terminator);
```
dcristoloveanu, line 137: >sizeof(RC_STRING_INTERNAL) - sizeof(RC_STRING) + string_length_with_terminator [](http://example.com/codeflow?start=106&length=78)

Funny thing is that "in theory" this can overflow.

tests/rc_string_ut/rc_string_ut.c
```diff
@@ -218,6 +241,174 @@ TEST_FUNCTION(when_underlying_calls_fail_rc_string_create_also_fails)
...
+    // act
+    THANDLE(RC_STRING) rc_string = rc_string_create_with_format("hell%c", 'o');
+
+    // assert
+    ASSERT_ARE_EQUAL(char_ptr, umock_c_get_expected_calls(), umock_c_get_actual_calls());
+    ASSERT_IS_NULL(rc_string);
+}
+
+/* Tests_SRS_RC_STRING_07_008: [ If any error occurs, `rc_string_create_with_format` shall fail and return `NULL`. ]*/
+TEST_FUNCTION(when_underlying_calls_fail_rc_string_create_with_format_also_fails)
```
dcristoloveanu, line 386: >when_underlying_calls_fail_rc_string_create_with_format_also_fails [](http://example.com/codeflow?start=14&length=66)

We should have a more adversarial test where the return value is -100 for example.

dcristoloveanu (dismissed):
:clock1:

tests/rc_string_ut/rc_string_ut.c
jebrando: We can't see what the 24 bytes are and it would possibly change with different version of linux.  Technically we can try to just make the size measure up, but the underlying structure would be unknow.  Would this work with the register?

dcristoloveanu: Yeah, but technically you don't need an alias :-D. Just need to define the 3-4 functions that umock needs in order to know and handle a type.

devdoc/rc_string_requirements.md
```diff
@@ -64,6 +64,8 @@ MOCKABLE_FUNCTION(, THANDLE(RC_STRING), rc_string_create, const char*, string);
 
 **SRS_RC_STRING_01_002: [** Otherwise, `rc_string_create` shall determine the length of `string`. **]**
 
+**SRS_RC_STRING_07_010: [** If the resulting memory size requested for the `THANDLE(RC_STRING)` and `string` results in an size_t overflow,`rc_string_create` shall fail and return `NULL`. **]**
```
dcristoloveanu, line 67: >7_010: [** If the resulting memory size requested for the `THANDLE(RC_STRING)` and `string` results in an size_t overflow,`rc_string_create [](http://example.com/codeflow?start=17&length=139)

Love it

src/rc_string.c
```diff
@@ -65,29 +66,41 @@ static void rc_string_dispose(RC_STRING* content)
...
+
     /* Codes_SRS_RC_STRING_01_002: [ Otherwise, rc_string_create shall determine the length of string. ]*/
     size_t string_length = strlen(string);
-    size_t string_length_with_terminator = string_length + 1;
-
-    /* Codes_SRS_RC_STRING_01_003: [ rc_string_create shall allocate memory for the THANDLE(RC_STRING), ensuring all the bytes in string can be copied (including the zero terminator). ]*/
-    THANDLE(RC_STRING) temp_result = THANDLE_MALLOC_WITH_EXTRA_SIZE(RC_STRING)(rc_string_dispose, sizeof(RC_STRING_INTERNAL) - sizeof(RC_STRING) + string_length_with_terminator);
-    if (temp_result == NULL)
+    size_t memory_allocation_max_size = SIZE_MAX - (sizeof(RC_STRING_INTERNAL) - sizeof(RC_STRING) + 1);
+    if (string_length > memory_allocation_max_size)
```
dcristoloveanu, line 74: >memory_allocation_max_size [](http://example.com/codeflow?start=24&length=26)

This could be replaced here with SIZE_MAX - (sizeof(RC_STRING_INTERNAL) - sizeof(RC_STRING) + 1

No need for an extra variable.

src/rc_string.c
```diff
@@ -65,29 +66,41 @@ static void rc_string_dispose(RC_STRING* content)
...
+    THANDLE(RC_STRING) result = NULL;
+
     /* Codes_SRS_RC_STRING_01_002: [ Otherwise, rc_string_create shall determine the length of string. ]*/
     size_t string_length = strlen(string);
-    size_t string_length_with_terminator = string_length + 1;
-
-    /* Codes_SRS_RC_STRING_01_003: [ rc_string_create shall allocate memory for the THANDLE(RC_STRING), ensuring all the bytes in string can be copied (including the zero terminator). ]*/
-    THANDLE(RC_STRING) temp_result = THANDLE_MALLOC_WITH_EXTRA_SIZE(RC_STRING)(rc_string_dispose, sizeof(RC_STRING_INTERNAL) - sizeof(RC_STRING) + string_length_with_terminator);
-    if (temp_result == NULL)
+    size_t memory_allocation_max_size = SIZE_MAX - (sizeof(RC_STRING_INTERNAL) - sizeof(RC_STRING) + 1);
```
dcristoloveanu, line 73: > + 1 [](http://example.com/codeflow?start=98&length=4)

Are you sure this + 1 is correct here?

src/rc_string.c
```diff
@@ -65,29 +66,41 @@ static void rc_string_dispose(RC_STRING* content)
...
+    if (string_length > memory_allocation_max_size)
     {
-        /* Codes_SRS_RC_STRING_01_006: [ If any error occurs, rc_string_create shall fail and return NULL. ]*/
-        LogError("THANDLE_MALLOC_WITH_EXTRA_SIZE(RC_STRING) failed, extra size is %zu, string_length_with_terminator=%zu", sizeof(RC_STRING_INTERNAL) - sizeof(RC_STRING) + string_length_with_terminator, string_length_with_terminator);
+        /* Codes_SRS_RC_STRING_07_010: [ If the resulting memory size requested for the `THANDLE(RC_STRING)`and `string` results in an size_t overflow, `rc_string_create` shall failand return `NULL`. ]*/
+        LogError("Requested memory size overflow.");
```
dcristoloveanu, line 77: >Error("Requested memory size overflow."); [](http://example.com/codeflow?start=11&length=41)

Let's print in this the string_length at least so that when the error happens we know how much was requested so that we overflowed

src/rc_string.c
dcristoloveanu: Ah I see, this is the +1 because we want the null terminator.
It works as is but consider changing the check in line to be

    if (string_length_with_terminator > SIZE_MAX - (sizeof(RC_STRING_INTERNAL) - sizeof(RC_STRING)))

Looks cleaner than having the +1 as it is now. And we need to compute string_length_with_terminator anyway for later computations

src/rc_string.c
```diff
@@ -110,9 +123,68 @@ IMPLEMENT_MOCKABLE_FUNCTION(, THANDLE(RC_STRING), rc_string_create, const char*,
...
+        int string_length_with_terminator = vsnprintf(NULL, 0, format, args) + 1;
+        
+        if (string_length_with_terminator <= 0)
+        {
+            /*Codes_SRS_RC_STRING_07_003: [ If `vsnprintf` failed to determine the total number of characters written, `rc_string_create_with_format` shall fail and return `NULL`. ]*/
+            LogError("vsnprintf failed to determine the total number of characters written.");
+        }
+        else
+        {   
+            size_t memory_size_need = sizeof(RC_STRING_INTERNAL) - sizeof(RC_STRING) + string_length_with_terminator;
```
dcristoloveanu, line 147: >memory_size_need [](http://example.com/codeflow?start=19&length=16)

Nitpick: memory_size_needed

src/rc_string.c
```diff
@@ -110,9 +123,68 @@ IMPLEMENT_MOCKABLE_FUNCTION(, THANDLE(RC_STRING), rc_string_create, const char*,
...
+            
+            if (memory_size_need > SIZE_MAX) 
+            {
+                /*Codes_SRS_RC_STRING_07_009: [ If the resulting memory size requested for the `THANDLE(RC_STRING)`and the resulting formatted string results in an size_t overflow in `malloc_flex`, `rc_string_create_with_format` shall failand return `NULL`. ]*/
+                LogError("Requested memory size overflow.");
```
dcristoloveanu, line 152: >or("Requested memory size overflow."); [](http://example.com/codeflow?start=22&length=38)

Let's print the string_length_with_terminator for example so we know how much memory was requested.

tests/rc_string_ut/rc_string_ut.c
```diff
@@ -218,6 +288,224 @@ TEST_FUNCTION(when_underlying_calls_fail_rc_string_create_also_fails)
...
+    // assert
+    ASSERT_ARE_EQUAL(char_ptr, umock_c_get_expected_calls(), umock_c_get_actual_calls());
+    ASSERT_IS_NULL(rc_string);
+}
+
+/* Tests_SRS_RC_STRING_07_002: [ Otherwise `rc_string_create_with_format` shall determine the total number of characters written using `vsnprintf` with the variable number of arguments. ]*/
+/* Tests_SRS_RC_STRING_07_004: [ `rc_string_create_with_format` shall allocate memory for the `THANDLE(RC_STRING)`and the number of bytes for the resulting formatted string. ]*/
+/* Tests_SRS_RC_STRING_07_005: [ `rc_string_create_with_format` shall fill in the bytes of the string by using `vsnprintf`. ]*/
+/* Tests_SRS_RC_STRING_07_007: [ `rc_string_create_with_format` shall succeed and return a non - `NULL` handle. ]*/
+#ifdef WIN32
```
dcristoloveanu, line 310: Why mark it to only run on Windows?
I assume it should run on Linux too.

tests/rc_string_ut/rc_string_ut.c
```diff
@@ -218,6 +288,224 @@ TEST_FUNCTION(when_underlying_calls_fail_rc_string_create_also_fails)
...
+    THANDLE(RC_STRING) rc_string = rc_string_create_with_format("hell%c", 'o');
+
+    // assert
+    ASSERT_ARE_EQUAL(char_ptr, umock_c_get_expected_calls(), umock_c_get_actual_calls());
+    ASSERT_IS_NULL(rc_string);
+
+}
+
+/*Tests_SRS_RC_STRING_07_003: [ If `vsnprintf` failed to determine the total number of characters written, `rc_string_create_with_format` shall fail and return `NULL`. ]*/
+TEST_FUNCTION(when_vsnprintf_determine_written_characters_length_fail_with_return_value_negative_two_rc_string_create_with_format_also_fails)
```
dcristoloveanu, line 393: >fail [](http://example.com/codeflow?start=65&length=4)

Nitpick: fails

tests/rc_string_ut/rc_string_ut.c
```diff
@@ -218,6 +288,224 @@ TEST_FUNCTION(when_underlying_calls_fail_rc_string_create_also_fails)
...
+    ASSERT_ARE_EQUAL(char_ptr, umock_c_get_expected_calls(), umock_c_get_actual_calls());
+    ASSERT_IS_NOT_NULL(rc_string);
+    ASSERT_ARE_EQUAL(char_ptr, "", rc_string->string);
+
+    // cleanup
+    THANDLE_ASSIGN(RC_STRING)(&rc_string, NULL);
+}
+
+/*Tests_SRS_RC_STRING_07_009: [ If the resulting memory size requested for the `THANDLE(RC_STRING)`and the resulting formatted string results in an size_t overflow in `malloc_flex`, `rc_string_create_with_format` shall failand return `NULL`. ]*/
+TEST_FUNCTION(rc_string_create_with_format_with_overflow_memory_allocation_size_fails)
```
dcristoloveanu, line 432: >with_format_with_overflow_memory_allocation_size [](http://example.com/codeflow?start=31&length=48)

Nitpick: with_overflow_for_memory_allocation_size

dcristoloveanu (changes requested):
:clock1:

dcristoloveanu:
Almost there!
The only reason I'd like to see it again is to see what happened to the tests running on Linux (I see they are only set to run on Windows now).

tests/rc_string_ut/rc_string_ut.c
```diff
@@ -1,18 +1,29 @@
...
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+#include <stdio.h>
 #include <stdlib.h>
 #include <stddef.h>
+<<<<<<< HEAD
+#include <string.h>
+=======
+#include <stdarg.h>
+>>>>>>> d9a2483ae2b1b9043e629706616ebf9e3cb91f39
```
jebrando: You'll need to change this to merge these two together.

dcristoloveanu (approved):
:shipit:

tests/rc_string_ut/rc_string_ut.c
```diff
@@ -218,6 +295,221 @@ TEST_FUNCTION(when_underlying_calls_fail_rc_string_create_also_fails)
...
+TEST_FUNCTION(when_underlying_calls_fail_rc_string_create_with_format_also_fails)
+{
+    // arrange
+    STRICT_EXPECTED_CALL(mocked_vsnprintf(IGNORED_ARG, IGNORED_ARG, IGNORED_ARG, IGNORED_ARG));
+    STRICT_EXPECTED_CALL(malloc_flex(IGNORED_ARG, IGNORED_ARG, sizeof(char)));
+    STRICT_EXPECTED_CALL(mocked_vsnprintf(IGNORED_ARG, IGNORED_ARG, IGNORED_ARG, IGNORED_ARG));
+
+    umock_c_negative_tests_snapshot();
+
+    for (size_t i = 2; i < umock_c_negative_tests_call_count(); i++)
```
dcristoloveanu, line 497: >2 [](http://example.com/codeflow?start=20&length=1)

This needs to be changed back

M-iceberg: done

PR #207, opened 2022-09-12
fix dlist append

jasmineymlo:
Fix an issue where DList_AppendTailList includes the head (contains no data) of the list to be appended in the middle of the final list.  Now the 2nd argument is expected to be the list head of the to be appended list.

### DIFF
--- a/devdoc/doublylinkedlist_requirements.md
+++ b/devdoc/doublylinkedlist_requirements.md
@@ -89,10 +89,10 @@ DList_InsertHeadList shall assume listHead and listEntry non-NULL and pointing t
 
 ### DList_AppendTailList
 ```c
-void DList_AppendTailList(PDLIST_ENTRY listHead, PDLIST_ENTRY ListToAppend);
+void DList_AppendTailList(PDLIST_ENTRY listHead, PDLIST_ENTRY listToBeAppendedHead);
 ```
 
-**SRS_DLIST_06_007: [** DList_AppendTailList shall place the list defined by listToAppend at the end of the list defined by the listHead parameter. **]**
+**SRS_DLIST_06_007: [** DList_AppendTailList shall place the list defined by `listToBeAppendedHead` at the end of the list defined by the `listHead` parameter. **]**
 
 ### DList_RemoveEntryList
 ```c
--- a/src/doublylinkedlist.c
+++ b/src/doublylinkedlist.c
@@ -106,19 +106,24 @@ DList_InsertTailList(
 void
 DList_AppendTailList(
     PDLIST_ENTRY ListHead,
-    PDLIST_ENTRY ListToAppend
+    PDLIST_ENTRY ListToBeAppendedHead
 )
 {
-    /* Codes_SRS_DLIST_06_007: [DList_AppendTailList shall place the list defined by listToAppend at the end of the list defined by the listHead parameter.] */
-    PDLIST_ENTRY ListEnd = ListHead->Blink;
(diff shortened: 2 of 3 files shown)


devdoc/doublylinkedlist_requirements.md
```diff
@@ -89,10 +89,10 @@ DList_InsertHeadList shall assume listHead and listEntry non-NULL and pointing t
 
 ### DList_AppendTailList
 ```c
-void DList_AppendTailList(PDLIST_ENTRY listHead, PDLIST_ENTRY ListToAppend);
+void DList_AppendTailList(PDLIST_ENTRY listHead, PDLIST_ENTRY listToBeAppendedHead);
```
dcristoloveanu, line 92: >listToBeAppendedHead [](http://example.com/codeflow?start=62&length=20)

If we changed the parameter name here, I'd expect the same change in the reals.

src/doublylinkedlist.c
```diff
@@ -106,20 +106,25 @@ DList_InsertTailList(
...
+    {
+        /* Codes_SRS_DLIST_06_007: [DList_AppendTailList shall place the list defined by listToBeAppendedHead at the end of the list defined by the listHead parameter.] */
+        PDLIST_ENTRY ListEnd = ListHead->Blink;
+
+        ListHead->Blink->Flink = ListToBeAppendedHead->Flink;
+        ListHead->Blink = ListToBeAppendedHead->Blink;
+        ListToBeAppendedHead->Blink->Flink = ListHead;
+        ListToBeAppendedHead->Flink->Blink = ListEnd;
+
+        DList_InitializeListHead(ListToBeAppendedHead);
```
dcristoloveanu, line 122: >DList_InitializeListHead [](http://example.com/codeflow?start=8&length=24)

This is calling a public API so this will be weird with regard to mocks.
What we do in these cases is that we have an internal_DList_InitializeListHead that we call from both the publicly exposed DList_InitializeListHead and from DList_AppendTailList

devdoc/doublylinkedlist_requirements.md
```diff
@@ -89,10 +89,10 @@ DList_InsertHeadList shall assume listHead and listEntry non-NULL and pointing t
...
 ```
 
-**SRS_DLIST_06_007: [** DList_AppendTailList shall place the list defined by listToAppend at the end of the list defined by the listHead parameter. **]**
+**SRS_DLIST_06_007: [** DList_AppendTailList shall place the list defined by `listToBeAppendedHead` at the end of the list defined by the `listHead` parameter. **]**
```
dcristoloveanu, line 95: >_DLIST_06_007: [** DList_AppendTailList shall place the list defined by `listToBeAppendedHead` at the end of the list defined by the `listHead` parameter. **]** [](http://example.com/codeflow?start=5&length=160)

Hmmm, the whole headless list vs a list with a head made me wonder again.
I remember that Tony took these from NT.
And lo and behold, here is the docs for one instance of it:

[https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/wdm/nf-wdm-appendtaillist](https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/wdm/nf-wdm-appendtaillist) 

The docs call out explicitly that this is a headless list.
I think I am very curious now how this worked fine for generations and it does not work for us.

dcristoloveanu (changes requested):
:clock1:

jasmineymlo:
Change not needed.  AppendTailList expects a headless list.

PR #221, opened 2022-11-28
make slist add at head

M-iceberg:


dcristoloveanu (approved):
:shipit:

ACKPaul:
The list should have an API that would add at the end of the list and this will be useful where sequential order matters. We can have a pointer that would track the tail of the list and method s_list_add_tail() and time complexity will be O(1). What will be the use cases of slist API's without s_list_add_head()?

ACKPaul (approved):

PR #222, opened 2022-11-30
interlocked_hl_wait_for_value and interlocked_hl_wait_for_not_value  should return timeout error code when wait_on_address fails due to timeout

ipg2013:


inc/c_util/interlocked_hl.h
```diff
@@ -22,6 +22,7 @@ extern "C" {
 #define INTERLOCKED_HL_RESULT_VALUES \
     INTERLOCKED_HL_OK, \
     INTERLOCKED_HL_ERROR, \
+    INTERLOCKED_HL_TIMEOUT, \
     INTERLOCKED_HL_CHANGED
```
dcristoloveanu, line 26: Consider a task to create a separate enum for InterlockedHL_WaitForValue (and likely InterlockedHL_WaitForNotValue)

src/interlocked_hl.c
```diff
@@ -6,6 +6,8 @@
 
 #include "umock_c/umock_c_prod.h"
 
+#include "windows.h"
```
mattdurak, line 9: >windows [](http://example.com/codeflow?start=10&length=7)

I don't think we should introduce windows-specific code in c-util, this won't work on Linux.

It looks like we first need a fix in c-pal so `wait_on_address` returns an enum and we can handle the various error cases there.

mattdurak (changes requested):
:clock1:

src/interlocked_hl.c
```diff
@@ -151,10 +165,20 @@ IMPLEMENT_MOCKABLE_FUNCTION(, INTERLOCKED_HL_RESULT, InterlockedHL_WaitForNotVal
             /* Codes_SRS_INTERLOCKED_HL_42_004: [ When waiting for the value at address to change, the milliseconds argument value shall be used as timeout. ]*/
             if (!wait_on_address(address, current_value, milliseconds))
```
dcristoloveanu, line 166: >milliseconds [](http://example.com/codeflow?start=57&length=12)

There is a general comment here for the old code that this  timeout is not properly respected.

Example:
- Let's say we called wait_on_address with 2000 ms as timeout
- wait_on_address returns as a spurious wake after 1000 ms
- we have 1000 left of the timeout, but we're going to call again wait_on_address with 2000.
- and then wait_on_address times out after these 2000ms

So in total we waited 3000 ms, not the 2000 we wanted.

Probably a separate task also.

tests/interlocked_hl_ut/interlocked_hl_ut.c
```diff
@@ -100,6 +102,8 @@ TEST_SUITE_INITIALIZE(a)
     REGISTER_SYNC_GLOBAL_MOCK_HOOK();
 
     REGISTER_GLOBAL_MOCK_HOOK(wait_on_address, hook_wait_on_address);
+
+    REGISTER_UMOCK_ALIAS_TYPE(DWORD, uint32_t);
```
dcristoloveanu, line 106: >DWORD [](http://example.com/codeflow?start=30&length=5)

Don't we have this in the windows extension of umock?

dcristoloveanu: umocktypes_windows_register_types?

src/interlocked_hl.c
dcristoloveanu: Such a great comment!

tests/interlocked_hl_ut/interlocked_hl_ut.c
dcristoloveanu: Likely this is not needed if we follow Matt's comment.

src/interlocked_hl.c
mattdurak: @dcristoloveanu  This is #7120375 which I mentioned originally for the read-and-wait design. I assigned it to Riddhi and put it in this sprint so we don't forget to also take care of that.

PR #228, opened 2022-12-06
Add RC_STRING_AS_CHARPTR macro

mattdurak:
This is a helper for calling a function like:
```c
int func(const char* str);
```
When you have a `THANDLE(RC_STRING)` which may or may not be `NULL`.
It is shorthand for:
```c
func((rc_str==NULL)?NULL:rc_str->string)
```


darobs (approved):
LGTM

PR #229, opened 2022-12-08
sm_close_begin_with_cb specs and code changes

RohitG28:
Specs and code changes. Added the test changes as well.


mattdurak:
> Close is realized by prohibiting all calls (including competing `sm_close_begin` calls) by setting a bit with `InterlockedOr`. `sm_close_begin` will wait for the state to reach `SM_OPENED` and the number of executing calls to be `0`. This allows an ongoing barrier to finish (and return to `SM_OPENED` state), or the executing APIs to finish.

It seems like there should be a note around here about how there is now a callback to do some work between prohibiting all calls and waiting for the executing calls to be 0. Likely with a note on why we have this (e.g. it is useful if we need to cancel ongoing operations etc.)

---
Refers to: devdoc/sm_requirements.md:58 in e6f2e02. [](commit_id = e6f2e0239dc67865c850655b2a17063984044a78, deletion_comment = False)

devdoc/sm_requirements.md
```diff
@@ -184,6 +186,8 @@ MOCKABLE_FUNCTION(, int, sm_close_begin, SM_HANDLE, sm);
 
 **SRS_SM_02_047: [** If the state is `SM_OPENED` then `sm_close_begin` shall switch it to `SM_OPENED_DRAINING_TO_CLOSE`. **]**
 
+**S_RS_SM_28_008: [** `callback` passed to `sm_close_begin_internal` shall be allowed to be NULL. **]**
```
mattdurak, line 189: >sm_close_begin_internal [](http://example.com/codeflow?start=44&length=23)

likely this spec doesn't belong here. Probably not needed at all because `sm_close_begin` will not have a callback (internal call is always using `NULL` and `sm_close_begin_with_cb` must have a non-`NULL` `callback` per 28_002

devdoc/sm_requirements.md
```diff
@@ -198,6 +202,23 @@ MOCKABLE_FUNCTION(, int, sm_close_begin, SM_HANDLE, sm);
...
+
+`sm_close_begin_with_cb` asks from `sm` permission to exit `SM_OPENED` state (or one of its derived state) and return to `SM_CREATED`. `sm_close_begin_with_cb` invokes the `callback` function with `callback_context` before waiting for pending calls to become 0.
+
+**S_RS_SM_28_001: [** If `sm` is `NULL` then `sm_close_begin_with_cb` shall fail and return `SM_ERROR`. **]**
+
+**S_RS_SM_28_002: [** If `callback` is `NULL` then `sm_close_begin_with_cb` shall fail and return `SM_ERROR`. **]**
+
+**S_RS_SM_28_003: [** `sm_close_begin_with_cb` shall behave as if `sm_close_begin` was called except that `callback` will be invoked with `callback_context` as arguments before waiting for pending calls to become 0. **]**
+
+**S_RS_SM_28_004: [** If `callback` is not `NULL`, `sm_close_begin_internal` shall invoke `callback` function with `callback_context` as argument. **]**
```
mattdurak, line 218: >**S_RS_SM_28_004: [** If `callback` is not `NULL`, `sm_close_begin_internal` shall invoke `callback` function with `callback_context` as argument. **] [](http://example.com/codeflow?start=0&length=150)

This is sort of 28_003, and also the "if" doesn't make sense because it can't be `NULL`. Also the `internal` thing isn't really part of the specs, it's an implementation detail.

devdoc/sm_requirements.md
```diff
@@ -198,6 +202,23 @@ MOCKABLE_FUNCTION(, int, sm_close_begin, SM_HANDLE, sm);
...
+MOCKABLE_FUNCTION(, SM_RESULT, sm_close_begin_with_cb, SM_HANDLE, sm, SM_CLOSING_COMPLETE_CALLBACK, callback, void*, callback_context);
+```
+
+`sm_close_begin_with_cb` asks from `sm` permission to exit `SM_OPENED` state (or one of its derived state) and return to `SM_CREATED`. `sm_close_begin_with_cb` invokes the `callback` function with `callback_context` before waiting for pending calls to become 0.
+
+**S_RS_SM_28_001: [** If `sm` is `NULL` then `sm_close_begin_with_cb` shall fail and return `SM_ERROR`. **]**
+
+**S_RS_SM_28_002: [** If `callback` is `NULL` then `sm_close_begin_with_cb` shall fail and return `SM_ERROR`. **]**
+
+**S_RS_SM_28_003: [** `sm_close_begin_with_cb` shall behave as if `sm_close_begin` was called except that `callback` will be invoked with `callback_context` as arguments before waiting for pending calls to become 0. **]**
```
mattdurak, line 216: >`sm_close_begin_with_cb` shall behave as if `sm_close_begin` was called except that `cal [](http://example.com/codeflow?start=22&length=88)

consider re-writing this like:

first spec: ...shall behave as if sm_close_begin was called except as follows:
second spec: after switching the state to `SM_OPENED_DRAINING_TO_CLOSE` ...shall invoke the callback...

I just think it's 2 specs, first is that it does the same as the normal close, second is that it calls the callback at a specific point in time.

devdoc/sm_requirements.md
```diff
@@ -198,6 +202,23 @@ MOCKABLE_FUNCTION(, int, sm_close_begin, SM_HANDLE, sm);
...
+
+**S_RS_SM_28_004: [** If `callback` is not `NULL`, `sm_close_begin_internal` shall invoke `callback` function with `callback_context` as argument. **]**
+
+**S_RS_SM_28_005: [** `sm_close_begin_with_cb` shall return the returned `SM_RESULT` from `sm_close_begin_internal`. **]**
```
mattdurak, line 220: >sm_close_begin_internal [](http://example.com/codeflow?start=91&length=23)

either this spec should be removed or could be re-phrased that it returns the same result as `sm_close_begin` or something.

devdoc/sm_requirements.md
```diff
@@ -92,6 +93,8 @@ typedef struct SM_HANDLE_DATA_TAG* SM_HANDLE;
 
 MU_DEFINE_ENUM(SM_RESULT, SM_RESULT_VALUES);
 
+typedef void(*SM_OPENED_DRAINING_TO_CLOSE_COMPLETE_CALLBACK)(void* context);
```
anporumb, line 96: >SM_OPENED_DRAINING_TO_CLOSE_COMPLETE_CALLBACK [](http://example.com/codeflow?start=14&length=45)

I will have to say that "DRAINING TO CLOSE" is very rather internal name and never was exposed to the user before.

So as far as discoverability of the API goes, there can be better names. 

How about considering renaming it to "ON_SM_CLOSING" or similar.

devdoc/sm_requirements.md
```diff
@@ -198,6 +201,23 @@ MOCKABLE_FUNCTION(, int, sm_close_begin, SM_HANDLE, sm);
...
+MOCKABLE_FUNCTION(, SM_RESULT, sm_close_begin_with_cb, SM_HANDLE, sm, SM_OPENED_DRAINING_TO_CLOSE_COMPLETE_CALLBACK, callback, void*, callback_context);
+```
+
+`sm_close_begin_with_cb` asks from `sm` permission to exit `SM_OPENED` state (or one of its derived state) and return to `SM_CREATED`. `sm_close_begin_with_cb` invokes the `callback` function with `callback_context` before waiting for pending calls to become 0.
+
+**S_RS_SM_28_001: [** If `sm` is `NULL` then `sm_close_begin_with_cb` shall fail and return `SM_ERROR`. **]**
+
+**S_RS_SM_28_002: [** If `callback` is `NULL` then `sm_close_begin_with_cb` shall fail and return `SM_ERROR`. **]**
+
+**S_RS_SM_28_003: [** `sm_close_begin_with_cb` shall behave as if `sm_close_begin` was called except as follows.  **]**
```
anporumb, line 215: >shall behave as if `sm_close_begin` was called [](http://example.com/codeflow?start=47&length=46)

I would respec both sm_close_being and sm_close_begin_with_cb to call some other internal function (let's say "sm_close_begin_impl"). And that function has 1 liner that says "if CB!=NULL then CB(context)".

sm_close_begin can pass a NULL cb, sm_close_being_with_cb would pass the non-NULL cb.

The above helps with requirements (and also mocks).

anporumb: Seems we have sm_close_begin_internal, so spec that instead.

src/sm.c
```diff
@@ -231,6 +231,16 @@ static SM_RESULT sm_close_begin_internal(SM_HANDLE sm)
...
                 {
+                    if (callback == NULL)
+                    {
+                        // do nothing.
+                    }
+                    else
+                    {
+                        // execute the callback.
+                        callback(callback_context);
+                    }
```
anporumb, line 242: shouldn't here be some SRS... bla bla bla??

404 not found.

anporumb: Also I don't see any changes for UT/INT, so that's reason enough to wait.

anporumb (changes requested):
:clock1:

devdoc/sm_requirements.md
```diff
@@ -198,6 +201,23 @@ MOCKABLE_FUNCTION(, int, sm_close_begin, SM_HANDLE, sm);
...
+MOCKABLE_FUNCTION(, SM_RESULT, sm_close_begin_with_cb, SM_HANDLE, sm, SM_CLOSING_COMPLETE_CALLBACK, callback, void*, callback_context);
+```
+
+`sm_close_begin_with_cb` asks from `sm` permission to exit `SM_OPENED` state (or one of its derived state) and return to `SM_CREATED`. `sm_close_begin_with_cb` invokes the `callback` function with `callback_context` before waiting for pending calls to become 0.
+
+**SRS_SM_28_001: [** If `sm` is `NULL` then `sm_close_begin_with_cb` shall fail and return `SM_ERROR`. **]**
+
+**SRS_SM_28_002: [** If `callback` is `NULL` then `sm_close_begin_with_cb` shall fail and return `SM_ERROR`. **]**
+
+**SRS_SM_28_003: [** `sm_close_begin_with_cb` shall behave as if `sm_close_begin` was called except as follows.  **]**
```
mattdurak: >SRS_SM_28_003 [](http://example.com/codeflow?start=2&length=13)

Now I'm curious, how did traceability pass? You have new specs (SRS and not something like S_R_S that's in code) but no new tests. Is our traceability test broken?

mattdurak: Nevermind, codeflow was just showing me some old iteration I think

devdoc/sm_requirements.md
RohitG28: I was with the understanding that we do not add specs for internal functions/implementations. Do you want me to add separate specs for sm_close_begin_internal?

src/sm.c
RohitG28: Not sure if we add specs for internal implementations. The reason I added a separate comment instead. I will update the patch with test changes.

tests/sm_int/sm_int.c
```diff
@@ -1420,23 +1478,23 @@ static void sm_switches_from_state_to_created(SM_GO_TO_STATE* goToState)
 
 TEST_FUNCTION(STATE_and_API)
```
mattdurak, line 1479: >TEST_FUNCTION [](http://example.com/codeflow?start=0&length=13)

Could we add a test which does what we are trying to fix.

1. Start a thread that will sm_exec_begin and wait for some interlocked variable before calling sm_exec_end (wait forever until cancelled)
2. Call sm_close_begin_with_cb
3. The callback will signal the first thread to complete (this is "cancel")
...

tests/sm_int/sm_int.c
```diff
@@ -1517,4 +1592,57 @@ TEST_FUNCTION(STATE_and_API)
...
+
+    /// act
+    THREAD_HANDLE cancel_wait_thread;
+    ASSERT_ARE_EQUAL(THREADAPI_RESULT, THREADAPI_OK, ThreadAPI_Create(&cancel_wait_thread, waitForCancel, &context));
+
+    // Add delay to make sure the above thread executes till wait.
+    ThreadAPI_Sleep(500);
+
+    /// assert
+    ASSERT_IS_TRUE(sm_close_begin_with_cb(context.sm, cancellingCallback, (void*)(&context.cancel_wait)) == SM_EXEC_GRANTED);
```
mattdurak, line 1641: >ASSERT_IS_TRUE [](http://example.com/codeflow?start=4&length=14)

ASSERT_ARE_EQUAL

tests/sm_int/sm_int.c
```diff
@@ -1517,4 +1592,57 @@ TEST_FUNCTION(STATE_and_API)
...
+    volatile_atomic int32_t* address = context;
+    ASSERT_IS_NOT_NULL(address);
+    wake_by_address_single(address);
+}
+
+TEST_FUNCTION(sm_close_begin_with_cb_triggers_cancel)
+{
+    /// arrange
+    WAIT_FOR_CANCEL context;
+    context.sm = sm_create(NULL);
```
mattdurak, line 1626: >context.sm [](http://example.com/codeflow?start=4&length=10)

ASSERT_IS_NOT_NULL

tests/sm_int/sm_int.c
```diff
@@ -1517,4 +1592,57 @@ TEST_FUNCTION(STATE_and_API)
...
+    context.cancel_wait = 0;
```
mattdurak, line 1627: >cancel_wait [](http://example.com/codeflow?start=12&length=11)

it's volatile so `interlocked_exchange` instead

tests/sm_int/sm_int.c
```diff
@@ -1517,4 +1592,57 @@ TEST_FUNCTION(STATE_and_API)
...
+    volatile_atomic int32_t cancel_wait;
+} WAIT_FOR_CANCEL;
+
+static int waitForCancel(void* context)
+{
+    WAIT_FOR_CANCEL* cancel_wait_context = context;
+    ASSERT_IS_NOT_NULL(cancel_wait_context);
+    int32_t current_value = interlocked_add(&cancel_wait_context->cancel_wait, 0);
+    sm_exec_begin(cancel_wait_context->sm);
+    if (wait_on_address(&cancel_wait_context->cancel_wait, current_value, UINT32_MAX) != WAIT_ON_ADDRESS_OK)
```
mattdurak, line 1607: >wait_on_address [](http://example.com/codeflow?start=8&length=15)

This can have spurious wakes, either should be `InterlockedHL_WaitForValue` or just do the loop here (check the value with `interlocked_add` again in a loop).

tests/sm_int/sm_int.c
```diff
@@ -1517,4 +1592,57 @@ TEST_FUNCTION(STATE_and_API)
...
+    SM_HANDLE sm;
+    volatile_atomic int32_t cancel_wait;
+} WAIT_FOR_CANCEL;
+
+static int waitForCancel(void* context)
+{
+    WAIT_FOR_CANCEL* cancel_wait_context = context;
+    ASSERT_IS_NOT_NULL(cancel_wait_context);
+    int32_t current_value = interlocked_add(&cancel_wait_context->cancel_wait, 0);
+    sm_exec_begin(cancel_wait_context->sm);
```
mattdurak, line 1606: >sm_exec_begin [](http://example.com/codeflow?start=4&length=13)

missing ASSERT_ARE_EQUAL with the result granted

tests/sm_int/sm_int.c
```diff
@@ -1517,4 +1592,57 @@ TEST_FUNCTION(STATE_and_API)
...
+    }
+    sm_exec_end(cancel_wait_context->sm);
+    return 0;
+}
+
+static void cancellingCallback(void* context)
+{
+    volatile_atomic int32_t* address = context;
+    ASSERT_IS_NOT_NULL(address);
+    wake_by_address_single(address);
```
mattdurak, line 1619: >wake_by_address_single [](http://example.com/codeflow?start=4&length=22)

Should set the value first with `interlocked_exchange` or `InterlockedHL_SetAndWake` instead

devdoc/sm_requirements.md
```diff
@@ -169,34 +172,61 @@ MOCKABLE_FUNCTION(, void, sm_open_end, SM_HANDLE, sm, bool, success);
 
 **SRS_SM_02_075: [** If `success` is `false` then `sm_open_end` shall switch the state to `SM_CREATED`. **]**
 
+### sm_close_begin_internal
```
mattdurak, line 175: should add the `static SM_RESULT...` signature here

mattdurak (approved):
:shipit:

anporumb (approved):
:shipit:

PR #254, opened 2023-02-22
Allow list item removal during traversal

ipg2013:


src/doublylinkedlist.c
```diff
@@ -167,7 +168,7 @@ IMPLEMENT_MOCKABLE_FUNCTION(, int, DList_ForEach, PDLIST_ENTRY, listHead, DLIST_
                     goto end;
                 }
             }
-            entry = entry->Flink;
+            entry = next;
```
mattdurak, line 171: > entry = next; [](http://example.com/codeflow?start=11&length=14)

We should have a spec (...shall allow removal blah blah) and a test.

PR #256, opened 2023-03-03
object_lifetime_tracker_register_object detects duplicate objects

parth21999:


devdoc/object_lifetime_tracker.md
```diff
@@ -127,19 +135,19 @@ static int is_same_object(PDLIST_ENTRY listEntry, const void* object_match_conte
...
 
 ```c
-MOCKABLE_FUNCTION(, int, object_lifetime_tracker_register_object, OBJECT_LIFETIME_TRACKER_HANDLE, object_lifetime_tracker, const void*, key, void*, object, DESTROY_OBJECT, destroy_object);
+MOCKABLE_FUNCTION(, OBJECT_LIFETIME_TRACKER_REGISTER_OBJECT_RESULT, object_lifetime_tracker_register_object, OBJECT_LIFETIME_TRACKER_HANDLE, object_lifetime_tracker, const void*, key, void*, object, DESTROY_OBJECT, destroy_object, void*, destroy_context);
 ```
-`object_lifetime_tracker_register_object` registers an object belonging to a key with the object lifetime tracker. Once an object is registered with a given `key`, it can be destroyed using the given `destroy_object` destructor when `object_lifetime_tracker_destroy_all_objects_for_key` is called.
+`object_lifetime_tracker_register_object` registers an object belonging to a key with the object lifetime tracker. Once an object is registered with a given `key`, it can be destroyed using the given `destroy_object` destructor when `object_lifetime_tracker_destroy_all_objects_for_key` is called. `object_lifetime_tracker_register_object` does not allow an `object` to be registered with the same `key` multiple times.
```
anporumb, line 140: >the same [](http://example.com/codeflow?start=389&length=8)

what about different keys?

parth21999: Yes, that's ok.

That is also desirable. In geo_mux_demux_server we would want to allow the same client to register with multiple destinations.

devdoc/object_lifetime_tracker.md
```diff
@@ -151,13 +159,23 @@ MOCKABLE_FUNCTION(, int, object_lifetime_tracker_register_object, OBJECT_LIFETIM
...
 
-**SRS_OBJECT_LIFETIME_TRACKER_43_051: [** `object_lifetime_tracker_register_object` shall release the lock. **]**
+**SRS_OBJECT_LIFETIME_TRACKER_43_077: [** If the `object` is not found in the DList of objects:**]**
+
+  - **SRS_OBJECT_LIFETIME_TRACKER_43_060: [** `object_lifetime_tracker_register_object` shall allocate memory to store data associated with the `object`. **]**
 
-**SRS_OBJECT_LIFETIME_TRACKER_43_020: [** `object_lifetime_tracker_register_object` shall succeed and return zero. **]**
+  - **SRS_OBJECT_LIFETIME_TRACKER_43_019: [** `object_lifetime_tracker_register_object` shall store the given `object` with the given `destroy_object` and `destroy_context` in the DList of objects for given `key` by calling `DList_InsertHeadList`. **]**
+
+  - **SRS_OBJECT_LIFETIME_TRACKER_43_078: [** If the given `key` had not been found, `object_lifetime_tracker_registeer_object` shall return `OBJECT_LIFETIME_TRACKER_REGISTER_OBJECT_NEW_KEY`. **]**
```
anporumb, line 170: >**SRS_OBJECT_LIFETIME_TRACKER_43_078: [** If the given `key` had not been found, [](http://example.com/codeflow?start=4&length=80)

The way this is written created confusion for me.

I don't think it should be indented.

devdoc/object_lifetime_tracker.md
```diff
@@ -151,13 +159,23 @@ MOCKABLE_FUNCTION(, int, object_lifetime_tracker_register_object, OBJECT_LIFETIM
...
+
+  - **SRS_OBJECT_LIFETIME_TRACKER_43_079: [** `object_lifetime_tracker_register_object` shall return `OBJECT_LIFETIME_TRACKER_REGISTER_OBJECT_NEW_OBJECT`. **]**
```
anporumb, line 172: >SRS_OBJECT_LIFETIME_TRACKER_43_079 [](http://example.com/codeflow?start=6&length=34)

Can use some wording as in "If they key has been found, but the object has not been found then... "

devdoc/object_lifetime_tracker.md
```diff
@@ -151,13 +159,23 @@ MOCKABLE_FUNCTION(, int, object_lifetime_tracker_register_object, OBJECT_LIFETIM
...
+**SRS_OBJECT_LIFETIME_TRACKER_43_076: [** `object_lifetime_tracker_register_object` shall find the list entry for the given `object` in the DList of objects for the given `key` by calling `DList_ForEach` with `is_same_object`.**]**
 
-**SRS_OBJECT_LIFETIME_TRACKER_43_051: [** `object_lifetime_tracker_register_object` shall release the lock. **]**
+**SRS_OBJECT_LIFETIME_TRACKER_43_077: [** If the `object` is not found in the DList of objects:**]**
+
+  - **SRS_OBJECT_LIFETIME_TRACKER_43_060: [** `object_lifetime_tracker_register_object` shall allocate memory to store data associated with the `object`. **]**
 
-**SRS_OBJECT_LIFETIME_TRACKER_43_020: [** `object_lifetime_tracker_register_object` shall succeed and return zero. **]**
+  - **SRS_OBJECT_LIFETIME_TRACKER_43_019: [** `object_lifetime_tracker_register_object` shall store the given `object` with the given `destroy_object` and `destroy_context` in the DList of objects for given `key` by calling `DList_InsertHeadList`. **]**
```
anporumb, line 168: >* [](http://example.com/codeflow?start=252&length=1)

You can insert in a new line here sort of "Return values:" so the read would know (and not guess) that the following specs are related to "return value from the function".

anporumb (approved):
:shipit:

PR #264, opened 2023-03-27
Add SMHasherSupport dependency to reals library.

darobs:
Add SMHasherSupport dependency to reals library.

An upstream change I would like to make failed to build, because the c_util_reals library didn't take a dependency on SMHasherSupport like c_util does.


mattdurak (approved):
:shipit:

PR #267, opened 2023-03-30
Make an extra check on index to satisfy Semmle error.

darobs:
Semmle error:

```
##[error]1. Semmle Error SM01954 - File: deps/c-util/src/external_command_helper.c. Line: 90. Column 47. 
  Tool: Semmle: Rule: SM01954 (cpp/unclear-array-index-validation). 
  [String read by fgets](1) flows to here and is used in an array indexing expression, potentially causing an invalid access.
[String read by fgets](1) flows to here and is used in an array indexing expression, potentially causing an invalid access.
```

Added a check to ensure index is not >= array size, even though this should not happen, `strcspn` should always return a value < the array length. I expect this change will satisfy the tool.


mattdurak (approved):
:shipit:

PR #274, opened 2023-04-12
`job_queue` RFC

parth21999:
Here is a spike for the `job_queue` module needed for GeoReplication. I've tried to make the module be as generic as possible. Let me know what you think!

### DIFF
--- a/devdoc/job_queue.md
+++ b/devdoc/job_queue.md
@@ -1,0 +1,59 @@
+# `job_queue` requirements
+
+`job_queue` is a module that allows jobs to be executed asynchronously. It queues jobs when there are no requests and queues requests when there are no jobs. When a job arrives and there are queued requests, it executes the least recently queued request with the job. When a request arrives and there are queued jobs, it executes the request with the least recently queued job.
+
+Consumers of `job_queue` must create a "`session`" before they can queue requests. A session can make two kinds of requests:
+
+- `get` requests
+- `pop` requests
+
+The consumer provides a `JOB_REQUEST_FUNCTION` that is executed when that request is executed. If the `JOB_REQUEST_FUNCTION` returns `true`, that job is considered completed, and it's  `JOB_COMPLETE_CALLBACK` is called. Once a job is completed, any new sessions that will be started will not see the completed job.
+
+Even after a job is completed, it is not discarded. A job is discarded when all the sessions that were active when the job was pushed have popped it. If all the sessions that were active when the job was queued have popped a job without indicating that the job is complete, the `JOB_COMPLETE_CALLBACK` is called with `JOB_QUEUE_JOB_RESULT_ERROR` as the result.
+
+Sessions make progress independent of each other. It is possible that one session has popped all the queued requests, whereas another session has not popped any. A session can be thought of as an independent "view" of the job queue. Once a session has called `pop` on a job, it cannot call `pop` or `get` on the same job again. If a session has called `get` on a job, it can call `pop` on that job but cannot call `get` on that `job`.
+
+A producer of jobs must call `push` to queue jobs. All jobs are queued in a single queue and executed in FIFO order. The producer provides a `JOB_COMPLETE_CALLBACK` to be called when the job is completed. If a job is completed because a request marked it as completed (by returning `true` from `JOB_REQUEST_FUNCTION`), `JOB_COMPLETE_CALLBACK` is called with `JOB_QUEUE_JOB_RESULT_OK` as result. In any other case (session ends before job can be completed, job_queue is closing etc), the `JOB_COMPLETE_CALLBACK` is called with `JOB_QUEUE_JOB_RESULT_ERROR`.
+
+## Exposed API
+
+```c
+#define JOB_QUEUE_JOB_RESULT_VALUES \
+    JOB_QUEUE_JOB_RESULT_OK, \
+    JOB_QUEUE_JOB_RESULT_ERROR
+
+MU_DEFINE_ENUM(JOB_QUEUE_JOB_RESULT, JOB_QUEUE_JOB_RESULT_VALUES);
+
+#define JOB_QUEUE_EXECUTE_JOB_RESULT_VALUES \
+    JOB_QUEUE_EXECUTE_JOB_RESULT_OK, \
+    JOB_QUEUE_EXECUTE_JOB_RESULT_ERROR
+
+MU_DEFINE_ENUM(JOB_QUEUE_EXECUTE_JOB_RESULT, JOB_QUEUE_EXECUTE_JOB_RESULT_VALUES);
+
+typedef struct JOB_QUEUE_TAG* JOB_QUEUE_HANDLE;
+typedef struct JOB_QUEUE_SESSION_TAG* JOB_QUEUE_SESSION_HANDLE;
+
+typedef void(*JOB_COMPLETE_CALLBACK)(void* context, JOB_QUEUE_JOB_RESULT result);
+typedef bool(*JOB_REQUEST_FUNCTION)(const void* creation_context, void* job_request_context, void* job);
+
+#include "umock_c/umock_c_prod.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+    MOCKABLE_FUNCTION(, JOB_QUEUE_HANDLE, job_queue_create, const void*, creation_context);
+    MOCKABLE_FUNCTION(, void, job_queue_destroy, JOB_QUEUE_HANDLE, job_queue);
+    MOCKABLE_FUNCTION(, int, job_queue_push, JOB_QUEUE_HANDLE, job_queue, void*, job, JOB_COMPLETE_CALLBACK, job_complete_callback, void*, job_complete_callback_context);
+
+    MOCKABLE_FUNCTION(, JOB_QUEUE_SESSION_HANDLE, job_queue_session_begin, JOB_QUEUE_HANDLE, job_queue);
+    MOCKABLE_FUNCTION(, void, job_queue_session_end, JOB_QUEUE_HANDLE, job_queue, JOB_QUEUE_SESSION_HANDLE, job_queue_session);
+    MOCKABLE_FUNCTION(, int, job_queue_session_pop, JOB_QUEUE_HANDLE, job_queue, JOB_QUEUE_SESSION_HANDLE, job_queue_session, JOB_REQUEST_FUNCTION, job_request_function, void*, job_request_context);
+    MOCKABLE_FUNCTION(, int, job_queue_session_get, JOB_QUEUE_HANDLE, job_queue, JOB_QUEUE_SESSION_HANDLE, job_queue_session, JOB_REQUEST_FUNCTION, job_request_function, void*, job_request_context);
+
+
+#ifdef __cplusplus
+}
+#endif
+```
+
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -158,6 +158,7 @@ set(c_util_c_files
     ./src/strings.c
     ./src/tarray.c
     ./src/uuid_string.c
+    ./src/job_queue.c
 )
 
 set(c_util_h_files
@@ -185,6 +186,7 @@ set(c_util_h_files
     ./inc/c_util/thandle.h
     ./inc/c_util/thandle_tuple_array.h
     ./inc/c_util/uuid_string.h
+    ./inc/c_util/job_queue.h
 )
 
 if(WIN32)
--- a/inc/c_util/job_queue.h
+++ b/inc/c_util/job_queue.h
@@ -1,0 +1,48 @@
+// Copyright (C) Microsoft Corporation. All rights reserved.
+
+#ifndef JOB_QUEUE_H
+#define JOB_QUEUE_H
+
+#include "macro_utils/macro_utils.h"
+
+#define JOB_QUEUE_JOB_RESULT_VALUES \
+    JOB_QUEUE_JOB_RESULT_OK, \
+    JOB_QUEUE_JOB_RESULT_ERROR
+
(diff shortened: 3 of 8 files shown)


devdoc/job_queue.md
```diff
@@ -0,0 +1,59 @@
+# `job_queue` requirements
+
+`job_queue` is a module that allows jobs to be executed asynchronously. It queues jobs when there are no requests and queues requests when there are no jobs. When a job arrives and there are queued requests, it executes the least recently queued request with the job. When a request arrives and there are queued jobs, it executes the request with the least recently queued job.
```
mattdurak, line 3: >It queues jobs when there are no requests [](http://example.com/codeflow?start=72&length=41)

What does this mean? What if there are requests, what happens to jobs?

mattdurak, line 3: >queues requests when there are no jobs [](http://example.com/codeflow?start=118&length=38)

Maybe my real question is, what is a "job" and what is a "request"?

devdoc/job_queue.md
```diff
@@ -0,0 +1,59 @@
...
+
+Consumers of `job_queue` must create a "`session`" before they can queue requests. A session can make two kinds of requests:
```
mattdurak, line 5: >Consumers [](http://example.com/codeflow?start=0&length=9)

Is this essentially a "multi-producer multi-consumer queue" (since we currently have a multi-producer single-consumer lockless queue)?

mattdurak, line 5: >must create a "`session`" [](http://example.com/codeflow?start=25&length=25)

Why do we need a session for consumers? Or, what really **is** a session?

devdoc/job_queue.md
```diff
@@ -0,0 +1,59 @@
...
+
+- `get` requests
+- `pop` requests
```
mattdurak, line 8: Is "get" like "peek"? Do we actually need "get", or would "pop" alone work?

devdoc/job_queue.md
```diff
@@ -0,0 +1,59 @@
...
+
+The consumer provides a `JOB_REQUEST_FUNCTION` that is executed when that request is executed. If the `JOB_REQUEST_FUNCTION` returns `true`, that job is considered completed, and it's  `JOB_COMPLETE_CALLBACK` is called. Once a job is completed, any new sessions that will be started will not see the completed job.
```
mattdurak, line 10: >new sessions that will be started will not see the completed job [](http://example.com/codeflow?start=249&length=64)

Do current sessions all see the job even though it's completed?

Does every session see every job? Or is a job just processed by one session unless the processing fails?

devdoc/job_queue.md
```diff
@@ -0,0 +1,59 @@
...
+
+Even after a job is completed, it is not discarded. A job is discarded when all the sessions that were active when the job was pushed have popped it. If all the sessions that were active when the job was queued have popped a job without indicating that the job is complete, the `JOB_COMPLETE_CALLBACK` is called with `JOB_QUEUE_JOB_RESULT_ERROR` as the result.
```
mattdurak, line 12: >If all the sessions t [](http://example.com/codeflow?start=150&length=21)

Does this mean at least one session must indicate the job is complete?  What if some have popped it without indicating complete, is the result OK or ERROR or something else?

devdoc/job_queue.md
```diff
@@ -0,0 +1,59 @@
...
+
+Sessions make progress independent of each other. It is possible that one session has popped all the queued requests, whereas another session has not popped any. A session can be thought of as an independent "view" of the job queue. Once a session has called `pop` on a job, it cannot call `pop` or `get` on the same job again. If a session has called `get` on a job, it can call `pop` on that job but cannot call `get` on that `job`.
```
mattdurak, line 14: >Once a session has called `pop` on a job, it cannot call `pop` or `get` on the same job again. If a session has called `get` on a job, it can call `pop` on that job but cannot call `get` on that `job`. [](http://example.com/codeflow?start=233&length=201)

Is this a restriction on the consumer calling this? Meaning you can't call "get" twice in a row? Or is this an implementation detail that the queue will not give the same job multiple times?

It's worded like the former, but it seems more like the latter is true because there is no such thing below as calling pop **on a job**. It looks like "pop" does not take any job argument so which job is popped is handled internally.

Consider re-wording so it's clear:
- What happens when calling "get" multiple times from the same session in a row?
- What happens when calling "get" then "pop"?
etc.

devdoc/job_queue.md
```diff
@@ -0,0 +1,59 @@
...
+
+A producer of jobs must call `push` to queue jobs. All jobs are queued in a single queue and executed in FIFO order. The producer provides a `JOB_COMPLETE_CALLBACK` to be called when the job is completed. If a job is completed because a request marked it as completed (by returning `true` from `JOB_REQUEST_FUNCTION`), `JOB_COMPLETE_CALLBACK` is called with `JOB_QUEUE_JOB_RESULT_OK` as result. In any other case (session ends before job can be completed, job_queue is closing etc), the `JOB_COMPLETE_CALLBACK` is called with `JOB_QUEUE_JOB_RESULT_ERROR`.
```
mattdurak, line 16: >If a job is completed because a request marked it as completed (by returning `true` from `JOB_REQUEST_FUNCTION`), `JOB_COMPLETE_CALLBACK` is called with `JOB_QUEUE_JOB_RESULT_OK` as result [](http://example.com/codeflow?start=205&length=188)

I'm confused again here, earlier it sounded like each consumer would see the same job, but this says that as soon as one consumer completes the job, then the job complete callback is called. I assume the job_complete_callback is called exactly once.

mattdurak, line 16: >JOB_QUEUE_JOB_RESULT_ERROR [](http://example.com/codeflow?start=527&length=26)

Sounds like maybe there should be an "NOT_OPEN" error code to differentiate the queue closing and the consumers actually failing to process

devdoc/job_queue.md
```diff
@@ -0,0 +1,59 @@
+# `job_queue` requirements
```
mattdurak, line 1: I see now you mentioned on teams that this is for parked calls. Is there a larger design that this fits into? I think seeing some example or the big picture of how this module supports parked calls might help, perhaps in a separate design doc.

devdoc/job_queue.md
```diff
@@ -0,0 +1,59 @@
...
+    JOB_QUEUE_EXECUTE_JOB_RESULT_OK, \
+    JOB_QUEUE_EXECUTE_JOB_RESULT_ERROR
+
+MU_DEFINE_ENUM(JOB_QUEUE_EXECUTE_JOB_RESULT, JOB_QUEUE_EXECUTE_JOB_RESULT_VALUES);
+
+typedef struct JOB_QUEUE_TAG* JOB_QUEUE_HANDLE;
+typedef struct JOB_QUEUE_SESSION_TAG* JOB_QUEUE_SESSION_HANDLE;
+
+typedef void(*JOB_COMPLETE_CALLBACK)(void* context, JOB_QUEUE_JOB_RESULT result);
+typedef bool(*JOB_REQUEST_FUNCTION)(const void* creation_context, void* job_request_context, void* job);
```
mattdurak, line 37: >creation_context [](http://example.com/codeflow?start=48&length=16)

I'm curious why this is needed? Supposedly the call to `job_queue_session_pop` has all the context it needs and it can pass it in as `job_request_context`.

But this seems to mean that there is another shared context from queue creation time. If all the sessions really need shared context, they should just maange that themselves and store it as a pointer in their `job_request_context`.

As it's written, it seems a bit split and I'm not sure there's any real benefit...

devdoc/job_queue.md
```diff
@@ -0,0 +1,59 @@
...
+    JOB_QUEUE_JOB_RESULT_OK, \
+    JOB_QUEUE_JOB_RESULT_ERROR
+
+MU_DEFINE_ENUM(JOB_QUEUE_JOB_RESULT, JOB_QUEUE_JOB_RESULT_VALUES);
+
+#define JOB_QUEUE_EXECUTE_JOB_RESULT_VALUES \
+    JOB_QUEUE_EXECUTE_JOB_RESULT_OK, \
+    JOB_QUEUE_EXECUTE_JOB_RESULT_ERROR
+
+MU_DEFINE_ENUM(JOB_QUEUE_EXECUTE_JOB_RESULT, JOB_QUEUE_EXECUTE_JOB_RESULT_VALUES);
```
mattdurak, line 31: >JOB_QUEUE_EXECUTE_JOB_RESULT [](http://example.com/codeflow?start=15&length=28)

Not used...or should `JOB_REQUEST_FUNCTION` return this?

devdoc/job_queue.md
mattdurak: Also, when I did a parked calls implementation, I ran into the following problem and I'm curious if your overall design solves this.

In my case, I have a module that resolves SF client endpoints or returns the client if it's already resolved. When a user calls "get_client", if there is no client yet, it would push the request onto a mpsc queue, parking the call until the client is resolved by another worker thread.

Then a single worker thread would resolve the SF client, and pop all the items from the queue, calling their callbacks with the client. If that client resolution failed, it still popped the items and called the callbacks with error.

The callers had retry logic on error to call "get_client" again, which would push back into the queue. This caused infinite loops because the queue would never fully drain.

The solution I used was to insert a barrier into the queue in the worker thread before draining the queue and to not handle anything queued after the barrier (because the newer requests would again trigger the worker thread to retry when its current iteration was complete).

parth21999: I haven't run into the infinite loop problem yet. I'll keep that in mind.

devdoc/job_queue.md
parth21999: A "job" is what a producer puts in the queue. A "request" is what a consumer puts in the queue. 

In GeoReplication terms, `job_queue_push` is called when the state provider calls `Begin_Replicate`. 

Each replication stream has its own `JOB_QUEUE_SESSION`. Secondary state providers call replication_stream.get_next() which results in a `job_queue_session_get()` call. After receiving the operation, secondary state providers call Acknowledge(), which results in a `job_queue_session_pop()` call. Once enough sessions have popped the operation ("enough" is configurable") then the `JOB_COMPLETE_CALLBACK` is called and the GeoReplicator indicates to the primary state provider that the operation has been replicated.

parth21999: It means that if jobs are pushed when no requests have been submitted, they will be stored in a queue until a session submits requests to consume them.

If a job is submitted when there are pending requests, the `JOB_REQUEST_FUNCTION` for the least recently submitted job for each session is executed with the submitted job as parameter.

devdoc/job_queue.md
parth21999: I'm not very familiar with this terminology, but I would think that "single-producer multi-consumer queue" would be a better description. 

I intend to be pushing jobs (calling `job_queue_push`) from the same thread.

In GeoReplication terms, each job will be a `BeginReplicate` call, so these must be in the order of LSN. Each state provider shall use a separate `job_queue`.

devdoc/job_queue.md
parth21999: "get"  is a little like "peek", but "peek" only allows you to see the head of the queue whereas "get" allows you to iterate over the entire queue.

In GeoReplication terms, each `BeginReplicate` operation needs be gotten (by the secondary state provider calling `get_next`) and acknowledged. So this would require two passes over each job(each "job" is a `BeginReplicate` call by the primary state provider).

Now that makes me think... What if instead of having two passes over each job, we used to `job_queue`s. One for serving `get_next` the other for serving `acknowledge`. 

Hmm...

devdoc/job_queue.md
parth21999: Let's say job X is pushed when there are two sessions active: session A and session B.

Session B calls `job_queue_session_pop` and the `JOB_REQUEST_FUNCTION` from this pop request returns `true` which triggers job X's `JOB_COMPLETE_CALLBACK`. 

Session A will still see the job when it calls `job_queue_session_pop`.

Every session sees every job that is submitted after `job_queue_session_begin` for that session was called. Jobs are processed by multiple sessions.

devdoc/job_queue.md
parth21999: A session is a consumer. `job_queue` guarantees that the requests submitted by a session will be completed in FIFO order of submission. The session identifies the consumer.

devdoc/job_queue.md
parth21999: At least one session must indicate that the job is complete. If all sessions pop the job without indicating complete, it completed with result ERROR.

devdoc/job_queue.md
parth21999: I could have worded this better.

When I say "session calls pop/get on a job" I mean "session submits a pop/get request which gets matched to a job".

Q:What happens when calling "get" multiple times from the same session in a row?

A: The `JOB_REQUEST_FUNCTION` for each get request executes with jobs in the FIFO order of job submission until there are no more jobs left at which point the get requests get queued.

Q: What happens when calling "get" then "pop"?
A: Say two jobs are pushed: job X and job Y in that order. There are 4 active sessions: A, B, C, D.

session A submits get request, receives job X
session A submits get request, receives job Y
session A submits pop request, recieves job X
session A submits pop request, receives job Y

session B submits get request, receives job X
session B submits pop request, recieves job X
session B submits get request, receives job Y
session B submits pop request, receives job Y

session C submits pop request, receives job X
session C submits get request, recieves job Y
session C submits pop request, received job Y
session C submits get request, request is queued

session D submits pop request, receives job X
session D submits pop request, recieves job Y
session D submits get request, job is queued
session D submits get request, request is queued

devdoc/job_queue.md
parth21999: Leftover, will remove.

PR #275, opened 2023-04-17
`waiter_queue` RFC

parth21999:
Following discussions on the [`job_queue RFC`](https://github.com/Azure/c-util/pull/274), here is the `waiter_queue` module that implements just the parked calls aspect.

### DIFF
--- a/devdoc/waiter_queue_requirements.md
+++ b/devdoc/waiter_queue_requirements.md
@@ -2,21 +2,70 @@
 
 ## Overview
 
-`waiter_queue` is a module that allows a consumers to park calls and for producers to execute them when data becomes available. It has two parts: `waiter_queue_ll` that contains the logic for parking and executing calls and `waiter_queue_hl` that contains logic for making `waiter_queue` thread-safe.
+`waiter_queue` is a module that allows consumers to park calls and for producers to signal that the parked calls be executed. A "parked call" consists of a callback function and a context. The consumer provides the callback function and context when it parks a call. The consumer expects the parked call's callback function to be executed with the provided context when the producer signals.
 
+`waiter_queue` has two parts: `waiter_queue_ll` that contains the logic for parking and unblocking calls and `waiter_queue_hl` that contains logic for making `waiter_queue` thread-safe.
+
+Consumers call `waiter_queue_hl_add_waiter` with a `UNBLOCK_CALLBACK` callback function and context to park a call. The parked call is added to a queue and will be executed when the producers signals by calling `waiter_queue_ll_unblock_waiters`.
+
+Producers call `waiter_queue_ll_unblock_waiters` with the data they would like to provide to the parked calls. When a producer calls `waiter_queue_ll_unblock_waiters` and there are parked calls in the queue, the parked calls are executed with the given data and `WAITER_QUEUE_CALL_REASON_UNBLOCKED` as `reason`.
+
+Calls are executed in the FIFO order in which they were parked. If the `UNBLOCK_CALLBACK` for a parked call sets the `remove` out argument to `true`, that call is removed from the queue. Otherwise, the parked call stays in the queue and may be executed again when the producer signals again.
+
+A single call to `waiter_queue_ll_unblock_waiters` may result in executing multiple parked calls. `waiter_queue` executes each `UNBLOCK_CALLBACK` in FIFO order until a `UNBLOCK_CALLBACK` sets the `continue_processing` out argument to `false`.
+
+`waiter_queue_hl_destroy` calls `waiter_queue_ll_abandon` to abandon all parked calls. Each parked call that is in the queue when `waiter_queue_hl_destroy` is called will have its `UNBLOCK_CALLBACK` executed with `NULL` as `data` and `WAITER_QUEUE_CALL_REASON_ABONDONED` as `reason`. 
+
+## Example:
+
+`waiter_queue_hl_create` is called.
+
+Queue state: `head||tail`
+
+Consumer calls `waiter_queue_hl_add_waiter(cb1, ctx1)`
+
+Queue state: `head|(cb1, ctx1)|tail`
+
+Consumer calls `waiter_queue_hl_add_waiter(cb2, ctx2)`
+
+Queue state: `head|(cb1, ctx1)|(cb2, ctx2)|tail`
+
+Producer calls `waiter_queue_hl_unblock_waiters(data1)`
+
+`cb1` is called with `ctx1`, `data1` and `WAITER_QUEUE_CALL_REASON_UNBLOCKED`. `cb1` sets `remove` to `true` and `continue_processing` to `true`.
+
+Queue state: `head|(cb2, ctx2)|tail`
+
+`cb2` is called with `ctx2`, `data1` and `WAITER_QUEUE_CALL_REASON_UNBLOCKED`. `cb2` sets `remove` to `false` and `continue_processing` to `true`.
+
+Queue state: `head|(cb2, ctx2)|tail`
+
+Consumer calls `waiter_queue_hl_add_waiter(cb3, ctx3)`.
+
+Queue state: `head|(cb2, ctx2)|(cb3, ctx3)|tail`
+
+Producer calls `waiter_queue_hl_unblock_waiters(data2)`
+
+`cb2` is called with `ctx2`, `data2` and `WAITER_QUEUE_CALL_REASON_UNBLOCKED`. `cb2` sets `remove` to `true` and `continue_processing` to `false`.
+
+Queue state: `head|(cb3, ctx3)|tail`
+
+`waiter_queue_hl_destroy` is called. `waiter_queue_hl_destroy` calls `waiter_queue_ll_abandon`.
+
+`cb3` is called with `ctx3` as `context`, `NULL` as `data` and `WAITER_QUEUE_CALL_REASON_ABANDONED` as `reason`.
 
 ## Exposed API
 
 ### waiter_queue_ll
 ```
 #define WAITER_QUEUE_CALL_REASON_VALUES \
-    WAITER_QUEUE_CALL_REASON_POPPED, \
+    WAITER_QUEUE_CALL_REASON_UNBLOCKED, \
     WAITER_QUEUE_CALL_REASON_ABANDONED
 
 MU_DEFINE_ENUM(WAITER_QUEUE_CALL_REASON, WAITER_QUEUE_CALL_REASON_VALUES);
 
 typedef struct WAITER_QUEUE_LL_TAG* WAITER_QUEUE_LL_HANDLE;
-typedef bool(*POP_CALLBACK)(void* context, void* data, bool* continue_processing, WAITER_QUEUE_CALL_REASON reason);
+typedef void(*UNBLOCK_CALLBACK)(void* context, void* data, bool* remove, bool* continue_processing, WAITER_QUEUE_CALL_REASON reason);
 
 #ifdef __cplusplus
 extern "C" {
@@ -25,8 +74,8 @@ extern "C" {
     MOCKABLE_FUNCTION(, WAITER_QUEUE_LL_HANDLE, waiter_queue_ll_create);
     MOCKABLE_FUNCTION(, void, waiter_queue_ll_destroy, WAITER_QUEUE_LL_HANDLE, waiter_queue_ll);
 
-    MOCKABLE_FUNCTION(, int, waiter_queue_ll_push, WAITER_QUEUE_LL_HANDLE, waiter_queue_ll, POP_CALLBACK, pop_callback, void*, pop_callback_context);
-    MOCKABLE_FUNCTION(, int, waiter_queue_ll_pop, WAITER_QUEUE_LL_HANDLE, waiter_queue_ll, void*, data);
+    MOCKABLE_FUNCTION(, int, waiter_queue_ll_add_waiter, WAITER_QUEUE_LL_HANDLE, waiter_queue_ll, UNBLOCK_CALLBACK, unblock_callback, void*, unblock_callback_context);
+    MOCKABLE_FUNCTION(, int, waiter_queue_ll_unblock_waiters, WAITER_QUEUE_LL_HANDLE, waiter_queue_ll, void*, data);
     MOCKABLE_FUNCTION(, void, waiter_queue_ll_abandon, WAITER_QUEUE_LL_HANDLE, waiter_queue_ll);
 
 #ifdef __cplusplus
@@ -45,25 +94,10 @@ extern "C" {
     MOCKABLE_FUNCTION(, WAITER_QUEUE_HL_HANDLE, waiter_queue_hl_create);
     MOCKABLE_FUNCTION(, void, waiter_queue_hl_destroy, WAITER_QUEUE_HL_HANDLE, waiter_queue_hl);
 
-    MOCKABLE_FUNCTION(, int, waiter_queue_hl_push, WAITER_QUEUE_HL_HANDLE, waiter_queue_hl, POP_CALLBACK, pop_callback, void*, pop_callback_context);
-    MOCKABLE_FUNCTION(, int, waiter_queue_hl_pop, WAITER_QUEUE_HL_HANDLE, waiter_queue_hl, void*, data);
+    MOCKABLE_FUNCTION(, int, waiter_queue_hl_add_waiter, WAITER_QUEUE_HL_HANDLE, waiter_queue_hl, UNBLOCK_CALLBACK, unblock_callback, void*, unblock_callback_context);
+    MOCKABLE_FUNCTION(, int, waiter_queue_hl_unblock_waiters, WAITER_QUEUE_HL_HANDLE, waiter_queue_hl, void*, data);
 
 #ifdef __cplusplus
 }
 #endif
 ```
-
-Consumers of data call `waiter_queue_hl_push` with a callback function and context to park a call. The call is parked in a queue and will be executed when the producers indicates that data is available by means of calling `waiter_queue_hl_pop`.
-
-Producers of data call `waiter_queue_hl_pop` with the data they would like to provide to the parked calls. If any calls had been parked by consumers by means of calling `waiter_queue_hl_push`, the parked calls are executed with the given data and `WAITER_QUEUE_CALL_REASON` as `WAITER_QUEUE_CALL_REASON_POPPED`.
-
-Calls are executed in the FIFO order in which they were parked. If the `POP_CALLBACK` for a parked call returns `true`, that call is removed from the queue. Otherwise, the parked call stays in the queue and may be executed again when more data becomes available.
-
-A single call to `waiter_queue_hl_pop` may result in executing multiple parked calls. `waiter_queue` executes each `POP_CALLBACK` in FIFO order until a `POP_CALLBACK` sets the `continue_processing` out argument to `false`.
-
-If there are any parked calls when `waiter_queue_hl_destroy` is called, their `POP_CALLBACKS` will be executed with `NULL` as data and `WAITER_QUEUE_CALL_REASON_ABONDONED` as reason.
-
-
-
-
-
--- a/devdoc/waiter_queue_requirements.md
+++ b/devdoc/waiter_queue_requirements.md
@@ -1,0 +1,69 @@
+# `waiter_queue` requirements
+
+## Overview
+
+`waiter_queue` is a module that allows a consumers to park calls and for producers to execute them when data becomes available. It has two parts: `waiter_queue_ll` that contains the logic for parking and executing calls and `waiter_queue_hl` that contains logic for making `waiter_queue` thread-safe.
+
+
+## Exposed API
+
+### waiter_queue_ll
+```
+#define WAITER_QUEUE_CALL_REASON_VALUES \
+    WAITER_QUEUE_CALL_REASON_POPPED, \
+    WAITER_QUEUE_CALL_REASON_ABANDONED
+
+MU_DEFINE_ENUM(WAITER_QUEUE_CALL_REASON, WAITER_QUEUE_CALL_REASON_VALUES);
+
+typedef struct WAITER_QUEUE_LL_TAG* WAITER_QUEUE_LL_HANDLE;
+typedef bool(*POP_CALLBACK)(void* context, void* data, bool* continue_processing, WAITER_QUEUE_CALL_REASON reason);
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+    MOCKABLE_FUNCTION(, WAITER_QUEUE_LL_HANDLE, waiter_queue_ll_create);
+    MOCKABLE_FUNCTION(, void, waiter_queue_ll_destroy, WAITER_QUEUE_LL_HANDLE, waiter_queue_ll);
+
+    MOCKABLE_FUNCTION(, int, waiter_queue_ll_push, WAITER_QUEUE_LL_HANDLE, waiter_queue_ll, POP_CALLBACK, pop_callback, void*, pop_callback_context);
+    MOCKABLE_FUNCTION(, int, waiter_queue_ll_pop, WAITER_QUEUE_LL_HANDLE, waiter_queue_ll, void*, data);
+    MOCKABLE_FUNCTION(, void, waiter_queue_ll_abandon, WAITER_QUEUE_LL_HANDLE, waiter_queue_ll);
+
+#ifdef __cplusplus
+}
+#endif
+```
+
+### waiter_queue_hl
+```
+typedef struct WAITER_QUEUE_HL_TAG* WAITER_QUEUE_HL_HANDLE;
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+    MOCKABLE_FUNCTION(, WAITER_QUEUE_HL_HANDLE, waiter_queue_hl_create);
+    MOCKABLE_FUNCTION(, void, waiter_queue_hl_destroy, WAITER_QUEUE_HL_HANDLE, waiter_queue_hl);
+
+    MOCKABLE_FUNCTION(, int, waiter_queue_hl_push, WAITER_QUEUE_HL_HANDLE, waiter_queue_hl, POP_CALLBACK, pop_callback, void*, pop_callback_context);
+    MOCKABLE_FUNCTION(, int, waiter_queue_hl_pop, WAITER_QUEUE_HL_HANDLE, waiter_queue_hl, void*, data);
+
+#ifdef __cplusplus
+}
+#endif
+```
+
+Consumers of data call `waiter_queue_hl_push` with a callback function and context to park a call. The call is parked in a queue and will be executed when the producers indicates that data is available by means of calling `waiter_queue_hl_pop`.
+
+Producers of data call `waiter_queue_hl_pop` with the data they would like to provide to the parked calls. If any calls had been parked by consumers by means of calling `waiter_queue_hl_push`, the parked calls are executed with the given data and `WAITER_QUEUE_CALL_REASON` as `WAITER_QUEUE_CALL_REASON_POPPED`.
+
+Calls are executed in the FIFO order in which they were parked. If the `POP_CALLBACK` for a parked call returns `true`, that call is removed from the queue. Otherwise, the parked call stays in the queue and may be executed again when more data becomes available.
+
+A single call to `waiter_queue_hl_pop` may result in executing multiple parked calls. `waiter_queue` executes each `POP_CALLBACK` in FIFO order until a `POP_CALLBACK` sets the `continue_processing` out argument to `false`.
+
+If there are any parked calls when `waiter_queue_hl_destroy` is called, their `POP_CALLBACKS` will be executed with `NULL` as data and `WAITER_QUEUE_CALL_REASON_ABONDONED` as reason.
+
+
+
+
+
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -159,6 +159,8 @@ set(c_util_c_files
     ./src/tarray.c
     ./src/uuid_string.c
     ./src/async_op.c
+    ./src/waiter_queue_ll.c
+    ./src/waiter_queue_hl.c
 )
 
 set(c_util_cpp_files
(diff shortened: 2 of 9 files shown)


devdoc/waiter_queue_requirements.md
```diff
@@ -0,0 +1,69 @@
...
+#ifdef __cplusplus
+}
+#endif
+```
+
+Consumers of data call `waiter_queue_hl_push` with a callback function and context to park a call. The call is parked in a queue and will be executed when the producers indicates that data is available by means of calling `waiter_queue_hl_pop`.
+
+Producers of data call `waiter_queue_hl_pop` with the data they would like to provide to the parked calls. If any calls had been parked by consumers by means of calling `waiter_queue_hl_push`, the parked calls are executed with the given data and `WAITER_QUEUE_CALL_REASON` as `WAITER_QUEUE_CALL_REASON_POPPED`.
+
+Calls are executed in the FIFO order in which they were parked. If the `POP_CALLBACK` for a parked call returns `true`, that call is removed from the queue. Otherwise, the parked call stays in the queue and may be executed again when more data becomes available.
```
mattdurak, line 60: >If the `POP_CALLBACK` for a parked call returns `true` [](http://example.com/codeflow?start=64&length=54)

nit: you have `continue_processing` as an out arg and "remove" as a return value. Seems cleaner to just have them both as out args.

devdoc/waiter_queue_requirements.md
```diff
@@ -0,0 +1,69 @@
...
+
+A single call to `waiter_queue_hl_pop` may result in executing multiple parked calls. `waiter_queue` executes each `POP_CALLBACK` in FIFO order until a `POP_CALLBACK` sets the `continue_processing` out argument to `false`.
+
+If there are any parked calls when `waiter_queue_hl_destroy` is called, their `POP_CALLBACKS` will be executed with `NULL` as data and `WAITER_QUEUE_CALL_REASON_ABONDONED` as reason.
```
anporumb, line 64: >reason [](http://example.com/codeflow?start=175&length=6)

I have read the whole md


General observations: 
1)
Traditionally data is "tranferred to", "given to", "made available to", "added", "inserted", "submitted", "posted" with `PUSH`. I think the APIs here are all reversed, because "consumers of data call push". Frankly, what they do is "pop".

Also, if these terms are confusing, it is perfectly fine to name them "produce" and "consume".

2) The term "call" is SO VAGUE here that it is better to be avoided altogether. "call" implies some imperative action - as in "do something" by means of calling a function. I don't see any function pointers or anything else even remotely similar to "call". 

Even parameter in line 49 is called "data". Or in line 19.

So at this moment I understand that there's "data". 

3) "parked" - poetically speaking - there's nothing "parked" here. To me the term "park" sounds like "interrupted" and it would necessitate a "resume", "restart" somehow.

The document doesn't explain what a "call" is conceptually or what the "parked" variety of the "call" is.

devdoc/waiter_queue_requirements.md
anporumb, line 60: >Calls are executed [](http://example.com/codeflow?start=0&length=18)

What's the meaning of "executed" for a call?

devdoc/waiter_queue_requirements.md
```diff
@@ -0,0 +1,69 @@
...
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+    MOCKABLE_FUNCTION(, WAITER_QUEUE_HL_HANDLE, waiter_queue_hl_create);
+    MOCKABLE_FUNCTION(, void, waiter_queue_hl_destroy, WAITER_QUEUE_HL_HANDLE, waiter_queue_hl);
+
+    MOCKABLE_FUNCTION(, int, waiter_queue_hl_push, WAITER_QUEUE_HL_HANDLE, waiter_queue_hl, POP_CALLBACK, pop_callback, void*, pop_callback_context);
+    MOCKABLE_FUNCTION(, int, waiter_queue_hl_pop, WAITER_QUEUE_HL_HANDLE, waiter_queue_hl, void*, data);
```
anporumb, line 49: >void*, data [](http://example.com/codeflow?start=91&length=11)

What's the means by which the queue forgets about this "void* data"?

devdoc/waiter_queue_requirements.md
anporumb, line 64: >reason [](http://example.com/codeflow?start=175&length=6)

I can review this when I'll stop having stack overflows at every phrase which I am trying to decompile.

Also pictures (even when ASCII) would greatly help.

anporumb (changes requested):
:clock1:

devdoc/waiter_queue_requirements.md
```diff
@@ -0,0 +1,69 @@
...
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+    MOCKABLE_FUNCTION(, WAITER_QUEUE_LL_HANDLE, waiter_queue_ll_create);
+    MOCKABLE_FUNCTION(, void, waiter_queue_ll_destroy, WAITER_QUEUE_LL_HANDLE, waiter_queue_ll);
+
+    MOCKABLE_FUNCTION(, int, waiter_queue_ll_push, WAITER_QUEUE_LL_HANDLE, waiter_queue_ll, POP_CALLBACK, pop_callback, void*, pop_callback_context);
+    MOCKABLE_FUNCTION(, int, waiter_queue_ll_pop, WAITER_QUEUE_LL_HANDLE, waiter_queue_ll, void*, data);
+    MOCKABLE_FUNCTION(, void, waiter_queue_ll_abandon, WAITER_QUEUE_LL_HANDLE, waiter_queue_ll);
```
anporumb, line 30: >waiter_queue_ll_abandon [](http://example.com/codeflow?start=30&length=23)

What does this even do? 404 description not found. And why doesn't HL have the same APIs?

devdoc/waiter_queue_requirements.md
```diff
@@ -0,0 +1,69 @@
+# `waiter_queue` requirements
+
+## Overview
+
+`waiter_queue` is a module that allows a consumers to park calls and for producers to execute them when data becomes available. It has two parts: `waiter_queue_ll` that contains the logic for parking and executing calls and `waiter_queue_hl` that contains logic for making `waiter_queue` thread-safe.
```
dcristoloveanu, line 5: >allows a consumers [](http://example.com/codeflow?start=32&length=18)

"allows consumers" or "allows a consumer"

devdoc/waiter_queue_requirements.md
```diff
@@ -0,0 +1,69 @@
...
+typedef bool(*POP_CALLBACK)(void* context, void* data, bool* continue_processing, WAITER_QUEUE_CALL_REASON reason);
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+    MOCKABLE_FUNCTION(, WAITER_QUEUE_LL_HANDLE, waiter_queue_ll_create);
+    MOCKABLE_FUNCTION(, void, waiter_queue_ll_destroy, WAITER_QUEUE_LL_HANDLE, waiter_queue_ll);
+
+    MOCKABLE_FUNCTION(, int, waiter_queue_ll_push, WAITER_QUEUE_LL_HANDLE, waiter_queue_ll, POP_CALLBACK, pop_callback, void*, pop_callback_context);
```
dcristoloveanu, line 28: >push [](http://example.com/codeflow?start=45&length=4)

I think we should walk away from the push/pop poetry. I think we should treat the waiters we wait to have like async/await.

At core, this is what we want: we want execution to be paused or blocked for a certain sequence until a later time when a trigger to unblock it arises.

push and pop will confuse if used in this context.

devdoc/waiter_queue_requirements.md
dcristoloveanu, line 5: >t allows a consumers to park calls and for producers [](http://example.com/codeflow?start=30&length=52)

I think we are referencing in this statement producers/consumers (which relate to data/bytes) and at the same time we mention execution of calls.

I think if we mention both we'll get to a confusing concept. Let's leave the data out and focus on pausing/parking execution until a signal is received.

devdoc/waiter_queue_requirements.md
dcristoloveanu, line 60: >POP [](http://example.com/codeflow?start=72&length=3)

Yeah this pop/push really, REALLY makes things harder to read

devdoc/waiter_queue_requirements.md
```diff
@@ -0,0 +1,69 @@
...
+#endif
+```
+
+Consumers of data call `waiter_queue_hl_push` with a callback function and context to park a call. The call is parked in a queue and will be executed when the producers indicates that data is available by means of calling `waiter_queue_hl_pop`.
+
+Producers of data call `waiter_queue_hl_pop` with the data they would like to provide to the parked calls. If any calls had been parked by consumers by means of calling `waiter_queue_hl_push`, the parked calls are executed with the given data and `WAITER_QUEUE_CALL_REASON` as `WAITER_QUEUE_CALL_REASON_POPPED`.
+
+Calls are executed in the FIFO order in which they were parked. If the `POP_CALLBACK` for a parked call returns `true`, that call is removed from the queue. Otherwise, the parked call stays in the queue and may be executed again when more data becomes available.
+
+A single call to `waiter_queue_hl_pop` may result in executing multiple parked calls. `waiter_queue` executes each `POP_CALLBACK` in FIFO order until a `POP_CALLBACK` sets the `continue_processing` out argument to `false`.
```
dcristoloveanu, line 62: >may [](http://example.com/codeflow?start=39&length=3)

How is it determined which of the waiting calls are unblocked? Or is it all?

dcristoloveanu: Ah I see, it is explained later in the paragraph.

devdoc/waiter_queue_requirements.md
dcristoloveanu, line 30: >waiter_queue_ll_abandon [](http://example.com/codeflow?start=30&length=23)

What does this do? I don't recall discussing it and I don't see it in the doc.

dcristoloveanu:
Let's have a second iteration where we somehow get rid of push/pop. I normally would not mind poetry as "renames are cheap", but in this case I see this used in many places throughout the codebases we have, so I think we should spend a bit of time to polish the names too.

dcristoloveanu (changes requested):
:clock1:

devdoc/waiter_queue_requirements.md
parth21999: Changed wording to remove data.

devdoc/waiter_queue_requirements.md
```diff
@@ -0,0 +1,69 @@
...
+
+    MOCKABLE_FUNCTION(, int, waiter_queue_hl_push, WAITER_QUEUE_HL_HANDLE, waiter_queue_hl, POP_CALLBACK, pop_callback, void*, pop_callback_context);
+    MOCKABLE_FUNCTION(, int, waiter_queue_hl_pop, WAITER_QUEUE_HL_HANDLE, waiter_queue_hl, void*, data);
+
+#ifdef __cplusplus
+}
+#endif
+```
+
+Consumers of data call `waiter_queue_hl_push` with a callback function and context to park a call. The call is parked in a queue and will be executed when the producers indicates that data is available by means of calling `waiter_queue_hl_pop`.
```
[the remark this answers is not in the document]
parth21999: Renaming to "add_waiter" and "unblock_waiters"

devdoc/waiter_queue_requirements.md
parth21999: Renaming to "add_waiter" and "unblock_waiters"

devdoc/waiter_queue_requirements.md
parth21999: This is needed to abandon all parked calls when `waiter_queue_hl_destroy` is called. `waiter_queue_hl` needs some way of telling `waiter_queue_ll` to abandon the calls.

parth21999: HL doesn't need abandon because user can call `waiter_queue_destroy` which will result in the parked calls getting abandoned.

parth21999: This is needed to abandon all parked calls when `waiter_queue_hl_destroy` is called. `waiter_queue_hl` needs some way of telling `waiter_queue_ll` to abandon the calls.

devdoc/waiter_queue_requirements.md
parth21999: Lifetime of data is managed by the user. `waiter_queue` does not store `void* data`.

devdoc/waiter_queue_requirements.md
parth21999: 1)Renamed the APIs: push->add_waiter, pop->unblock_waiters

The thing being queued is parked call, not the data.

2) Line 19 contains the type definition for `POP_CALLBACK`. These callbacks are provided by the consumer to be called when signaled. I've also added a definition for parked calls in the next iteration:

A "parked call" consists of a callback function and a context. The consumer provides the callback function and context when it parks a call. The consumer expects the parked call's callback function to be executed with the provided context when the producer signals.

3) Given the above definition of "parked calls", does that make things clearer? The consumer is interrupted because the data it requires to proceed is not available yet. So it "parks" execution by providing a callback function that is to be executed when signaled (told to resume) by the producer.

devdoc/waiter_queue_requirements.md
parth21999: Changed remove to out arg.

parth21999: The `POP_CALLBACK` is called.

devdoc/waiter_queue_requirements.md
```diff
@@ -0,0 +1,103 @@
+# `waiter_queue` requirements
+
+## Overview
+
+`waiter_queue` is a module that allows consumers to park calls and for producers to signal that the parked calls be executed. A "parked call" consists of a callback function and a context. The consumer provides the callback function and context when it parks a call. The consumer expects the parked call's callback function to be executed with the provided context when the producer signals.
```
anporumb, line 5: > be [](http://example.com/codeflow?start=112&length=3)

can be/should be?

devdoc/waiter_queue_requirements.md
```diff
@@ -0,0 +1,103 @@
...
+
+`waiter_queue` has two parts: `waiter_queue_ll` that contains the logic for parking and unblocking calls and `waiter_queue_hl` that contains logic for making `waiter_queue` thread-safe.
```
anporumb, line 7: >has two parts [](http://example.com/codeflow?start=15&length=13)

it is 2 modules separated. Each with its own md, header everything. These `parts` just invite problems with unittests.

devdoc/waiter_queue_requirements.md
```diff
@@ -0,0 +1,103 @@
...
+
+Consumers call `waiter_queue_hl_add_waiter` with a `UNBLOCK_CALLBACK` callback function and context to park a call. The parked call is added to a queue and will be executed when the producers signals by calling `waiter_queue_ll_unblock_waiters`.
```
anporumb, line 9: >waiter_queue_ll_unblock_waiters [](http://example.com/codeflow?start=212&length=31)

This sounds a bit weird as the consumers exists in the HL layer (waiter_queue_*hl*_add_waiter) and the producer are in LL layer (waiter_queue_*ll*_unblock_waiters)

anporumb, line 9: >signals [](http://example.com/codeflow?start=192&length=7)

signal

devdoc/waiter_queue_requirements.md
```diff
@@ -0,0 +1,103 @@
...
+
+`cb1` is called with `ctx1`, `data1` and `WAITER_QUEUE_CALL_REASON_UNBLOCKED`. `cb1` sets `remove` to `true` and `continue_processing` to `true`.
+
+Queue state: `head|(cb2, ctx2)|tail`
+
+`cb2` is called with `ctx2`, `data1` and `WAITER_QUEUE_CALL_REASON_UNBLOCKED`. `cb2` sets `remove` to `false` and `continue_processing` to `true`.
+
+Queue state: `head|(cb2, ctx2)|tail`
+
+Consumer calls `waiter_queue_hl_add_waiter(cb3, ctx3)`.
```
mattdurak, line 43: I suppose in this example this line happens from a different thread, but if `cb2` actually made the call to `waiter_queue_hl_add_waiter(cb3, ctx3)` then cb3 would execute immediately.

I ask because in the one use case I have, if a waiter is unblocked with data that indicates an error, it will bubble that error up and then the callback handler for that will retry, which would re-queue all on the same thread in the execution context of the waiter callback. And this had infinite loops in my original implementation without a fix to break the cycle.

So I suppose I would still need to handle that unless the `waiter_queue_hl_unblock_waiters` only works off of a snapshot of the queue of waiters when it starts executing.

devdoc/waiter_queue_requirements.md
parth21999: Typo. That should be `waiter_queue_hl_unblock_waiters`

devdoc/waiter_queue_requirements.md
mattdurak, line 7: >`waiter_queue_ll` that contains the logic for parking and unblocking calls and `waiter_queue_hl` [](http://example.com/codeflow?start=30&length=96)

poetry: isn't this `waiter_queue` and `threadsafe_waiter_queue` or something? /consider

devdoc/waiter_queue_requirements.md
parth21999: Fixed

devdoc/waiter_queue_requirements.md
```diff
@@ -0,0 +1,103 @@
...
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+    MOCKABLE_FUNCTION(, WAITER_QUEUE_LL_HANDLE, waiter_queue_ll_create);
+    MOCKABLE_FUNCTION(, void, waiter_queue_ll_destroy, WAITER_QUEUE_LL_HANDLE, waiter_queue_ll);
+
+    MOCKABLE_FUNCTION(, int, waiter_queue_ll_add_waiter, WAITER_QUEUE_LL_HANDLE, waiter_queue_ll, UNBLOCK_CALLBACK, unblock_callback, void*, unblock_callback_context);
+    MOCKABLE_FUNCTION(, int, waiter_queue_ll_unblock_waiters, WAITER_QUEUE_LL_HANDLE, waiter_queue_ll, void*, data);
```
anporumb, line 78: >waiter_queue_ll_unblock_waiters [](http://example.com/codeflow?start=29&length=31)

`waiter_queue_ll_add_waiter` is asynchronous by nature (has callback, context). `waiter_queue_ll_unblock_waiters` is synchronous.

Is there any possible future where this function is async? And it is perfectly fine to say "uncertain now".

devdoc/waiter_queue_requirements.md
```diff
@@ -0,0 +1,103 @@
...
+
+Consumers call `waiter_queue_hl_add_waiter` with a `UNBLOCK_CALLBACK` callback function and context to park a call. The parked call is added to a queue and will be executed when the producers signals by calling `waiter_queue_ll_unblock_waiters`.
+
+Producers call `waiter_queue_ll_unblock_waiters` with the data they would like to provide to the parked calls. When a producer calls `waiter_queue_ll_unblock_waiters` and there are parked calls in the queue, the parked calls are executed with the given data and `WAITER_QUEUE_CALL_REASON_UNBLOCKED` as `reason`.
+
+Calls are executed in the FIFO order in which they were parked. If the `UNBLOCK_CALLBACK` for a parked call sets the `remove` out argument to `true`, that call is removed from the queue. Otherwise, the parked call stays in the queue and may be executed again when the producer signals again.
+
+A single call to `waiter_queue_ll_unblock_waiters` may result in executing multiple parked calls. `waiter_queue` executes each `UNBLOCK_CALLBACK` in FIFO order until a `UNBLOCK_CALLBACK` sets the `continue_processing` out argument to `false`.
```
parth21999, line 15: >ll [](http://example.com/codeflow?start=31&length=2)

typo

devdoc/waiter_queue_requirements.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Overview
+
+`waiter_queue` is a module that allows consumers to park calls and for producers to signal that the parked calls be executed. A "parked call" consists of a callback function and a context. The consumer provides the callback function and context when it parks a call. The consumer expects the parked call's callback function to be executed with the provided context when the producer signals.
+
+`waiter_queue` has two parts: `waiter_queue_ll` that contains the logic for parking and unblocking calls and `waiter_queue_hl` that contains logic for making `waiter_queue` thread-safe.
+
+Consumers call `waiter_queue_hl_add_waiter` with a `UNBLOCK_CALLBACK` callback function and context to park a call. The parked call is added to a queue and will be executed when the producers signals by calling `waiter_queue_ll_unblock_waiters`.
+
+Producers call `waiter_queue_ll_unblock_waiters` with the data they would like to provide to the parked calls. When a producer calls `waiter_queue_ll_unblock_waiters` and there are parked calls in the queue, the parked calls are executed with the given data and `WAITER_QUEUE_CALL_REASON_UNBLOCKED` as `reason`.
```
parth21999, line 11: >ll [](http://example.com/codeflow?start=29&length=2)

typo

devdoc/waiter_queue_requirements.md
parth21999, line 9: >ll [](http://example.com/codeflow?start=225&length=2)

typo.

devdoc/waiter_queue_requirements.md
```diff
@@ -0,0 +1,103 @@
...
+typedef void(*UNBLOCK_CALLBACK)(void* context, void* data, bool* remove, bool* continue_processing, WAITER_QUEUE_CALL_REASON reason);
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+    MOCKABLE_FUNCTION(, WAITER_QUEUE_LL_HANDLE, waiter_queue_ll_create);
+    MOCKABLE_FUNCTION(, void, waiter_queue_ll_destroy, WAITER_QUEUE_LL_HANDLE, waiter_queue_ll);
+
+    MOCKABLE_FUNCTION(, int, waiter_queue_ll_add_waiter, WAITER_QUEUE_LL_HANDLE, waiter_queue_ll, UNBLOCK_CALLBACK, unblock_callback, void*, unblock_callback_context);
```
anporumb, line 77: >waiter_queue_ll_add_waiter [](http://example.com/codeflow?start=29&length=26)

`waiter_queue_ll_add_waiter` and  `continue_processing` != `love`.

An earlier parked waiter can prohibit the later waiters from getting `data*`. So, that implies that the waiters are `ordered` and their behavior does depend on each other as well as the order that they've been enqueued.

So that makes me think that there's some sort of enqueuing waiter `manager` that will intelligently enqueue the waiters. Because `continue_processing` exists.

Another direct consequence of `continue_processing` is that it dictates that all the `UNBLOCK_CALLBACK` are sequenced.

So, we'd NEVER want that `data*` is served in parallel to the 1000 waiters?

devdoc/waiter_queue_requirements.md
mattdurak: I kind of mentioned this to Parth in my two scenarios. I have one scenario where the unblock_call needs to happen on another thread (when BSI unblocks pending read-and-wait it shouldn't delay the write completion thread). And in another scenario I just have a single-threaded worker who solely produces the data needed to unblock the waiters.

So in my mind, the choice of unblocking on another thread or not could be in a layer on top of this...maybe.

devdoc/waiter_queue_requirements.md
parth21999: We can implement it such that `waiter_queue_hl_unblock_waiters` causes only the parked calls that were parked before are executed. Adding verbiage for this behavior in the next iteration.

devdoc/waiter_queue_requirements.md
parth21999: "Another direct consequence of continue_processing is that it dictates that all the UNBLOCK_CALLBACK are sequenced.

So, we'd NEVER want that data* is served in parallel to the 1000 waiters?
"

`UNBLOCK_CALLBACK`s are sequenced anyway because they are added in sequence and are expected to execute in sequence.

devdoc/waiter_queue_requirements.md
anporumb, line 78: >waiter_queue_ll_unblock_waiters [](http://example.com/codeflow?start=29&length=31)

I think execution here might take quite a long time since it depends on the number of waiters and what their callbacks do. 

Would be quite awesome to have this asynchronous. Or if!!! and here's wild idea, "if waiters would indicate their preferred calling environment from which their callbacks should be called". 

... just saying, this only happens because a waiter had to *wait* (it's in the name). If it didn't have to wait, then its execution would have continued immediately in its own execution environment.

Meayh, a bit wild. How about that asynchronous part?

devdoc/waiter_queue_requirements.md
```diff
@@ -0,0 +1,103 @@
...
+
+    MOCKABLE_FUNCTION(, WAITER_QUEUE_HL_HANDLE, waiter_queue_hl_create);
+    MOCKABLE_FUNCTION(, void, waiter_queue_hl_destroy, WAITER_QUEUE_HL_HANDLE, waiter_queue_hl);
+
+    MOCKABLE_FUNCTION(, int, waiter_queue_hl_add_waiter, WAITER_QUEUE_HL_HANDLE, waiter_queue_hl, UNBLOCK_CALLBACK, unblock_callback, void*, unblock_callback_context);
+    MOCKABLE_FUNCTION(, int, waiter_queue_hl_unblock_waiters, WAITER_QUEUE_HL_HANDLE, waiter_queue_hl, void*, data);
+
+#ifdef __cplusplus
+}
+#endif
```
anporumb, line 102: >endif [](http://example.com/codeflow?start=1&length=5)

I need to see where this goes, so "waiting". Ping me vigurously when where's something new to look at.

devdoc/waiter_queue_requirements.md
```diff
@@ -0,0 +1,103 @@
...
+
+Producer calls `waiter_queue_hl_unblock_waiters(data2)`
+
+`cb2` is called with `ctx2`, `data2` and `WAITER_QUEUE_CALL_REASON_UNBLOCKED`. `cb2` sets `remove` to `true` and `continue_processing` to `false`.
+
+Queue state: `head|(cb3, ctx3)|tail`
+
+`waiter_queue_hl_destroy` is called. `waiter_queue_hl_destroy` calls `waiter_queue_ll_abandon`.
+
+`cb3` is called with `ctx3` as `context`, `NULL` as `data` and `WAITER_QUEUE_CALL_REASON_ABANDONED` as `reason`.
```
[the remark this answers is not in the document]
parth21999: Up for discussion:`bool* remove` and `bool* continue_processing` can eith

devdoc/waiter_queue_requirements.md
parth21999: "Is there any possible future where this function is async? And it is perfectly fine to say "uncertain now"."

Leaning towards let this stay synchronous. If the user wants to do any long running task, they should create their own thread.

parth21999: We could definitely implement it such that the callbacks execute on separate threads. But that would mean using threadpools. I think the user should be responsible for creating threads if needed. That way this module stays simple.

Walking over the list and firing the callbacks must be done anyway. Now if the user wants to execute a long-running task in the `UNBLOCK_CALLBACK` and they don't want the producer thread to be blocked, they should use their own threads.

parth21999: It's just a question of how much functionality we want to bake into this module. I say let's keep it simple and let the user deal with the threads. But if there is a compelling argument for it to be in this module, I'm all ears.

devdoc/waiter_queue_requirements.md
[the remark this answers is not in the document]
parth21999: Yes, the user can call `waiter_queue_hl_add_waiter` from an `UNBLOCK_CALLBACK`.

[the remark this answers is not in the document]
mattdurak: So will the real implementation not use srw_lock? Because srw_lock is not reentrant

devdoc/waiter_queue_requirements.md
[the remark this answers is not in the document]
parth21999: Even in GeoReplication, it can happen that the provided data is for LSN `n` but the parked call is waiting for LSN `n+1`. So in that case the parked call will want to ignore the data and keep its position in the queue.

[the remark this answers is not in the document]
parth21999: Going to use ASYNC_OP in next iteration.

devdoc/waiter_queue_requirements.md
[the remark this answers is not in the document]
parth21999: Hmm that's a good point.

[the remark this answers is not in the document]
parth21999: `add_waiter` during an  ongoing `unblock_waiters` shouldn't be a problem, because only parked calls that were parked before the ongoing `unblock_waiters` will be executed.

`unblock_waiters` during an ongoing `unblock_waiters` can create problems because now we have don't have any guarantees that the parked calls will be executed with the data given to the first `unblock_waiters` call before being executed with the data from the second `unblock_waiters` call. But the user cannot expect that anyway if the two `unblock_waiters` calls are being made concurrently.

What are your thoughts?

devdoc/waiter_queue_requirements.md
```diff
@@ -0,0 +1,103 @@
...
+
+Producer calls `waiter_queue_hl_unblock_waiters(data1)`
+
+`cb1` is called with `ctx1`, `data1` and `WAITER_QUEUE_CALL_REASON_UNBLOCKED`. `cb1` sets `remove` to `true` and `continue_processing` to `true`.
+
+Queue state: `head|(cb2, ctx2)|tail`
+
+`cb2` is called with `ctx2`, `data1` and `WAITER_QUEUE_CALL_REASON_UNBLOCKED`. `cb2` sets `remove` to `false` and `continue_processing` to `true`.
+
+Queue state: `head|(cb2, ctx2)|tail`
```
darobs, line 41: >cb2, ctx2) [](http://example.com/codeflow?start=20&length=10)

I don't understand why (cb2, ctz2) isn't removed from the queue once it's called.

darobs: Oh, `remove` is output from the callback function. Now I understand. Was there a use case for this?

PR #279, opened 2023-05-12
`waiter` spike

parth21999:
`waiter` is a module that allows the user to move data asynchronously from producer to consumer without having to block the thread.

### DIFF
--- a/devdoc/rc_ptr.md
+++ b/devdoc/rc_ptr.md
@@ -1,0 +1,34 @@
+# `rc_ptr` requirements
+
+## Overview
+
+`rc_ptr` is a module that encapsulates reference-counted pointers.
+
+## Exposed API
+```c
+typedef void (*RC_PTR_FREE_FUNC)(void* context);
+
+typedef struct RC_PTR_TAG
+{
+    const void* ptr;
+    RC_PTR_FREE_FUNC free_func;
+} RC_PTR;
+
+THANDLE_TYPE_DECLARE(RC_PTR);
+
+#define PRI_RC_PTR "p"
+
+#define RC_PTR_VALUE(rc) ((rc)->ptr)
+
+#define RC_PTR_AS_VOIDPTR(rc) (((rc) == NULL) ? NULL : (rc)->ptr)
+
+MOCKABLE_FUNCTION(, THANDLE(RC_PTR), rc_ptr_create_with_move_memory, const void*, ptr, RC_PTR_FREE_FUNC, free_func);
+```
+
+### rc_ptr_create_with_move_memory
+```c
+MOCKABLE_FUNCTION(, THANDLE(RC_PTR), rc_ptr_create_with_move_memory, const void*, ptr, RC_PTR_FREE_FUNC, free_func);
+```
+
+- `rc_ptr_create_with_move_memory` shall create a `THANDLE(RC_PTR)` that encapsulates the given `ptr`.
+- When the reference count for the returned `THANDLE(RC_PTR)` reaches `0`, the given `free_func` shall be called with the given `ptr`.
--- a/devdoc/rc_ptr.md
+++ b/devdoc/rc_ptr.md
@@ -11,7 +11,7 @@ typedef void (*RC_PTR_FREE_FUNC)(void* context);
 typedef struct RC_PTR_TAG
 {
     const void* ptr;
-    RC_PTR_FREE_FUNC free_func;
+    RC_PTR_FREE_FUNC free_func; // will be removed later
 } RC_PTR;
 
 THANDLE_TYPE_DECLARE(RC_PTR);
@@ -30,5 +30,6 @@ MOCKABLE_FUNCTION(, THANDLE(RC_PTR), rc_ptr_create_with_move_memory, const void*
 MOCKABLE_FUNCTION(, THANDLE(RC_PTR), rc_ptr_create_with_move_memory, const void*, ptr, RC_PTR_FREE_FUNC, free_func);
 ```
 
-- `rc_ptr_create_with_move_memory` shall create a `THANDLE(RC_PTR)` that encapsulates the given `ptr`.
+`rc_ptr_create_with_move_memory` creates a `THANDLE(RC_PTR)` that encapsulates the given `ptr`.
+
 - When the reference count for the returned `THANDLE(RC_PTR)` reaches `0`, the given `free_func` shall be called with the given `ptr`.
--- a/devdoc/waiter.md
+++ b/devdoc/waiter.md
@@ -29,16 +29,16 @@ typedef void(*NOTIFY_COMPLETE_CALLBACK)(void* context, WAITER_RESULT result);
     MOCKABLE_FUNCTION(, WAITER_HANDLE, waiter_create)
 ```
 
-- `waiter_create` creates the waiter and returns it.
+- `waiter_create` shall create the waiter and returns it.
 
 ### waiter_destroy
 ```c
     MOCKABLE_FUNCTION(, void, waiter_destroy, WAITER_HANDLE, waiter)
 ```
 
-- `waiter_destroy` destroys the waiter.
 - If `waiter_register_notification` has been called previously but `waiter_notify` has not been called, `waiter_destroy` shall call the `notification_callback` given to `waiter_register_notification` with `context` as the `context` given to `waiter_register_notification`, `data` as `NULL` and `result` as `WAITER_RESULT_ABANDONED`.
 - If `waiter_notify` has been called previously but `waiter_register_notification` has not been called, `waiter_destroy` shall call the `notify_complete_callback` given to `waiter_notify` with `context` as the `context` given to `waiter_notify`, and `result` as `WAITER_RESULT_ABANDONED`.
+- `waiter_destroy` shall destroy the given `waiter`.
 
 ### waiter_register_notification
 ```c
--- a/devdoc/waiter.md
+++ b/devdoc/waiter.md
@@ -35,12 +35,14 @@ typedef void(*NOTIFY_COMPLETE_CALLBACK)(void* context, WAITER_RESULT result);
     MOCKABLE_FUNCTION(, WAITER_HANDLE, waiter_create)
 ```
 
-- `waiter_create` shall create the waiter and returns it.
+`waiter_create` creates the waiter and returns it.
 
 ### waiter_destroy
 ```c
     MOCKABLE_FUNCTION(, void, waiter_destroy, WAITER_HANDLE, waiter)
 ```
+
+`waiter_destroy` destroys the given `waiter`.
 
 - If `waiter_register_notification` has been called previously but `waiter_notify` has not been called, `waiter_destroy` shall call the `notification_callback` given to `waiter_register_notification` with `context` as the `context` given to `waiter_register_notification`, `data` as `NULL` and `result` as `WAITER_RESULT_ABANDONED`.
 - If `waiter_notify` has been called previously but `waiter_register_notification` has not been called, `waiter_destroy` shall call the `notify_complete_callback` given to `waiter_notify` with `context` as the `context` given to `waiter_notify`, and `result` as `WAITER_RESULT_ABANDONED`.
@@ -50,8 +52,10 @@ typedef void(*NOTIFY_COMPLETE_CALLBACK)(void* context, WAITER_RESULT result);
 ```c
     MOCKABLE_FUNCTION(, THANDLE(ASYNC_OP), waiter_register_notification, WAITER_HANDLE, waiter, NOTIFICATION_CALLBACK, notification_callback, void*, context);
 ```
-- `waiter_register_notification` registers the given `notification_callback` to be called when `waiter_notify` is called.
-- If `waiter_notify` has been called previously, `waiter_register_notification` shall  call:
+
+`waiter_register_notification` registers the given `notification_callback` to be called when `waiter_notify` is called.
+
+- If `waiter_notify` has been called previously, `waiter_register_notification` shall call:
     - the given `notification_callback` with `context` as the given `context`, `data` as the `data` that was given to `waiter_notify` and `result` as `WAITER_RESULT_OK`.
     - the `notify_callback` given to `waiter_notify` with `context` as the `context` given to `waiter_notify` and `result` as `WAITER_RESULT_OK`.
 - `waiter_register_notification` shall return a `THANDLE(ASYNC_OP)` that the user can use to cancel the operation.
@@ -62,7 +66,9 @@ typedef void(*NOTIFY_COMPLETE_CALLBACK)(void* context, WAITER_RESULT result);
 ```c
     MOCKABLE_FUNCTION(, THANDLE(ASYNC_OP), waiter_notify, WAITER_HANDLE, waiter, THANDLE(RC_PTR), data, NOTIFY_COMPLETE_CALLBACK, notify_complete_callback, void*, context);
 ```
-- `waiter_notify` notifies the waiter that there is data available and registers the given `notify_complete_callback` to be called when the given `data` has been consumed.
+
+`waiter_notify` notifies the waiter that there is data available and registers the given `notify_complete_callback` to be called when the given `data` has been consumed.
+
 - If `waiter_register_notification` has been called previously, `waiter_notify` shall call:
     - the given `notify_complete_callback` with `context` as the given `context` and `result` as `WAITER_RESULT_OK`.
     - the `notification_callback` given to `waiter_register_notification` with `context` as the `context` given to `waiter_register_notification`, `data` as the given `data` and `result` as `WAITER_RESULT_OK`.
--- a/devdoc/waiter.md
+++ b/devdoc/waiter.md
@@ -1,0 +1,65 @@
+# `waiter` requirements
+
+## Overview
+
+`waiter` is a module that allows the user to move data asynchronously from producer to consumer without having to block the thread. 
+
+## Exposed API
+```c
+#define WAITER_RESULT_VALUES \
+    WAITER_RESULT_OK, \
+    WAITER_RESULT_CANCELLED, \
+    WAITER_RESULT_ABANDONED
+
+MU_DEFINE_ENUM(WAITER_RESULT, WAITER_RESULT_VALUES);
+
+typedef struct WAITER_TAG* WAITER_HANDLE;
+typedef void(*NOTIFICATION_CALLBACK)(void* context, THANDLE(RC_PTR) data, WAITER_RESULT result);
+typedef void(*NOTIFY_COMPLETE_CALLBACK)(void* context, WAITER_RESULT result);
+
+    MOCKABLE_FUNCTION(, WAITER_HANDLE, waiter_create)
+    MOCKABLE_FUNCTION(, void, waiter_destroy, WAITER_HANDLE, waiter)
+    MOCKABLE_FUNCTION(, THANDLE(ASYNC_OP), waiter_register_notification, WAITER_HANDLE, waiter, NOTIFICATION_CALLBACK, notification_callback, void*, context);
+    MOCKABLE_FUNCTION(, THANDLE(ASYNC_OP), waiter_notify, WAITER_HANDLE, waiter, THANDLE(RC_PTR), data, NOTIFY_COMPLETE_CALLBACK, notify_complete_callback, void*, context);
+
+```
+
+### waiter_create
+```c
+    MOCKABLE_FUNCTION(, WAITER_HANDLE, waiter_create)
+```
+
+- `waiter_create` creates the waiter and returns it.
+
+### waiter_destroy
+```c
+    MOCKABLE_FUNCTION(, void, waiter_destroy, WAITER_HANDLE, waiter)
+```
+
+- `waiter_destroy` destroys the waiter.
+- If `waiter_register_notification` has been called previously but `waiter_notify` has not been called, `waiter_destroy` shall call the `notification_callback` given to `waiter_register_notification` with `context` as the `context` given to `waiter_register_notification`, `data` as `NULL` and `result` as `WAITER_RESULT_ABANDONED`.
+- If `waiter_notify` has been called previously but `waiter_register_notification` has not been called, `waiter_destroy` shall call the `notify_complete_callback` given to `waiter_notify` with `context` as the `context` given to `waiter_notify`, and `result` as `WAITER_RESULT_ABANDONED`.
+
+### waiter_register_notification
+```c
+    MOCKABLE_FUNCTION(, THANDLE(ASYNC_OP), waiter_register_notification, WAITER_HANDLE, waiter, NOTIFICATION_CALLBACK, notification_callback, void*, context);
+```
+- `waiter_register_notification` registers the given `notification_callback` to be called when `waiter_notify` is called.
+- If `waiter_notify` has been called previously, `waiter_register_notification` shall  call:
+    - the given `notification_callback` with `context` as the given `context`, `data` as the `data` that was given to `waiter_notify` and `result` as `WAITER_RESULT_OK`.
+    - the `notify_callback` given to `waiter_notify` with `context` as the `context` given to `waiter_notify` and `result` as `WAITER_RESULT_OK`.
+- `waiter_register_notification` shall return a `THANDLE(ASYNC_OP)` that the user can use to cancel the operation.
+- If the user calls `async_op_cancel` on the returned `THANDLE(ASYNC_OP)` before `waiter_notify` is called, the given `notification_callback` is called with `context` as the given `context`, `data` as `NULL` and `result` as `WAITER_RESULT_CANCELLED`.
+- If `waiter_register_notification` is called more than once, it shall fail and return `NULL`.
+
+### waiter_notify
+```c
+    MOCKABLE_FUNCTION(, THANDLE(ASYNC_OP), waiter_notify, WAITER_HANDLE, waiter, THANDLE(RC_PTR), data, NOTIFY_COMPLETE_CALLBACK, notify_complete_callback, void*, context);
+```
+- `waiter_notify` notifies the waiter that there is data available and registers the given `notify_complete_callback` to be called when the given `data` has been consumed.
+- If `waiter_register_notification` has been called previously, `waiter_notify` shall call:
+    - the given `notify_complete_callback` with `context` as the given `context` and `result` as `WAITER_RESULT_OK`.
+    - the `notification_callback` given to `waiter_register_notification` with `context` as the `context` given to `waiter_register_notification`, `data` as the given `data` and `result` as `WAITER_RESULT_OK`.
+- `waiter_notify` shall return a `THANDLE(ASYNC_OP)` that the user can use to cancel the operation.
+- If the user calls `async_op_cancel` on the returned `THANDLE(ASYNC_OP)` before `waiter_register_notification` is called, the given `notify_complete_callback` is called with `context` as the given `context` and `result` as `WAITER_RESULT_CANCELLED`.
+- If `waiter_notify` is called more than once, it shall fail and return `NULL`.
--- a/devdoc/waiter.md
+++ b/devdoc/waiter.md
@@ -2,7 +2,13 @@
 
 ## Overview
 
-`waiter` is a module that allows the user to move data asynchronously from producer to consumer without having to block the thread. 
+`waiter` is a module that allows the user to move data asynchronously from producer to consumer without having to block the thread.
+
+Consumers call `waiter_register_notification` to register a `NOTIFICATION_CALLBACK` function to be called when data becomes available.
+
+Producers call `waiter_notify` to notify the waiter that data is available.
+
+`waiter_register_notification` and `waiter_notify` can be called from different threads in any order.
 
 ## Exposed API
 ```c
--- a/inc/c_util/rc_ptr.h
+++ b/inc/c_util/rc_ptr.h
@@ -6,20 +6,13 @@
 
 #ifdef __cplusplus
 #include <cstddef>
-#include <cstdarg>
 #else
-#include <stdarg.h>
 #include <stddef.h>
 #endif
 
 #include "c_pal/thandle.h"
 
-#include "umock_c/umock_c_prod.h"
 
-#ifdef __cplusplus
-extern "C"
-{
-#endif
 
 typedef void (*RC_PTR_FREE_FUNC)(void* context);
 
@@ -35,9 +28,15 @@ THANDLE_TYPE_DECLARE(RC_PTR);
 
 #define RC_PTR_VALUE(rc) ((rc)->ptr)
 
-#define RC_PTR_AS_VOIDPTR(rc) (((rc) == NULL) ? NULL : (rc)->ptr)
+#define RC_PTR_VALUE_OR_NULL(rc) (((rc) == NULL) ? NULL : (rc)->ptr)
 
-MOCKABLE_FUNCTION(, THANDLE(RC_PTR), rc_ptr_create_with_move_memory, const void*, ptr, RC_PTR_FREE_FUNC, free_func);
+#include "umock_c/umock_c_prod.h"
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
+    MOCKABLE_FUNCTION(, THANDLE(RC_PTR), rc_ptr_create_with_move_memory, const void*, ptr, RC_PTR_FREE_FUNC, free_func);
 
 #ifdef __cplusplus
 }
--- a/inc/c_util/rc_ptr.h
+++ b/inc/c_util/rc_ptr.h
@@ -1,0 +1,46 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#ifndef RC_PTR_H
+#define RC_PTR_H
+
+#ifdef __cplusplus
+#include <cstddef>
+#include <cstdarg>
+#else
+#include <stdarg.h>
+#include <stddef.h>
+#endif
+
+#include "c_pal/thandle.h"
+
+#include "umock_c/umock_c_prod.h"
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
+typedef void (*RC_PTR_FREE_FUNC)(void* context);
+
+typedef struct RC_PTR_TAG
+{
+    const void* ptr;
+    RC_PTR_FREE_FUNC free_func;
+} RC_PTR;
+
+THANDLE_TYPE_DECLARE(RC_PTR);
+
+#define PRI_RC_PTR "p"
+
+#define RC_PTR_VALUE(rc) ((rc)->ptr)
+
+#define RC_PTR_AS_VOIDPTR(rc) (((rc) == NULL) ? NULL : (rc)->ptr)
+
+MOCKABLE_FUNCTION(, THANDLE(RC_PTR), rc_ptr_create_with_move_memory, const void*, ptr, RC_PTR_FREE_FUNC, free_func);
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif  /* RC_PTR_H */
--- a/src/rc_ptr.c
+++ b/src/rc_ptr.c
@@ -16,7 +16,7 @@
 
 THANDLE_TYPE_DEFINE(RC_PTR);
 
-static void rc_ptr_destroy(RC_PTR* rc_ptr)
+static void dispose(RC_PTR* rc_ptr)
 {
     if (rc_ptr == NULL)
     {
@@ -24,23 +24,26 @@ static void rc_ptr_destroy(RC_PTR* rc_ptr)
     }
     else
     {
-        rc_ptr->free_func((void*)rc_ptr->ptr);
+        if (rc_ptr->free_func != NULL)
+        {
+            rc_ptr->free_func((void*)rc_ptr->ptr);
+        }
     }
 }
 
 THANDLE(RC_PTR) rc_ptr_create_with_move_memory(const void* ptr, RC_PTR_FREE_FUNC free_func)
 {
     THANDLE(RC_PTR) result = NULL;
-    if (ptr == NULL || free_func == NULL)
+    if (ptr == NULL)
     {
         LogError("Invalid arguments: const void* ptr=%p, RC_PTR_FREE_FUNC free_func = %p", ptr, free_func);
     }
     else
     {
-        THANDLE(RC_PTR) thandle = THANDLE_MALLOC(RC_PTR)(rc_ptr_destroy);
+        THANDLE(RC_PTR) thandle = THANDLE_MALLOC(RC_PTR)(dispose);
         if (thandle == NULL)
         {
-            LogError("Failure in THANDLE_MALLOC(RC_PTR)(rc_ptr_destroy=%p)", rc_ptr_destroy);
+            LogError("Failure in THANDLE_MALLOC(RC_PTR)(dispose=%p)", dispose);
         }
         else
         {
--- a/src/rc_ptr.c
+++ b/src/rc_ptr.c
@@ -1,0 +1,54 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#include <stdlib.h>
+
+#include "macro_utils/macro_utils.h"
+
+#include "c_logging/xlogging.h"
+
+#include "c_pal/gballoc_hl.h"
+#include "c_pal/gballoc_hl_redirect.h"
+
+#include "c_pal/thandle.h"
+
+#include "c_util/rc_ptr.h"
+
+THANDLE_TYPE_DEFINE(RC_PTR);
+
+static void rc_ptr_destroy(RC_PTR* rc_ptr)
+{
+    if (rc_ptr == NULL)
+    {
+        LogError("Invalid arguments: RC_PTR* rc_ptr=%p", rc_ptr);
+    }
+    else
+    {
+        rc_ptr->free_func((void*)rc_ptr->ptr);
+    }
+}
+
+THANDLE(RC_PTR) rc_ptr_create_with_move_memory(const void* ptr, RC_PTR_FREE_FUNC free_func)
+{
+    THANDLE(RC_PTR) result = NULL;
+    if (ptr == NULL || free_func == NULL)
+    {
+        LogError("Invalid arguments: const void* ptr=%p, RC_PTR_FREE_FUNC free_func = %p", ptr, free_func);
+    }
+    else
+    {
+        THANDLE(RC_PTR) thandle = THANDLE_MALLOC(RC_PTR)(rc_ptr_destroy);
+        if (thandle == NULL)
+        {
+            LogError("Failure in THANDLE_MALLOC(RC_PTR)(rc_ptr_destroy=%p)", rc_ptr_destroy);
+        }
+        else
+        {
+            RC_PTR* rc_ptr = THANDLE_GET_T(RC_PTR)(thandle);
+            rc_ptr->ptr = ptr;
+            rc_ptr->free_func = free_func;
+            THANDLE_INITIALIZE_MOVE(RC_PTR)(&result, &thandle);
+        }
+    }
+    return result;
+}
--- a/src/waiter.c
+++ b/src/waiter.c
@@ -10,23 +10,20 @@
 
 #include "c_logging/xlogging.h"
 
-#include "c_pal/gballoc_hl.h"
-#include "c_pal/gballoc_hl_redirect.h"
 #include "c_pal/thandle.h"
 #include "c_pal/sm.h"
 #include "c_pal/srw_lock.h"
 
 #include "c_util/async_op.h"
+#include "c_util/rc_ptr.h"
 
 #include "c_util/waiter.h"
-
-THANDLE_TYPE_DEFINE(void_ptr);
 
 typedef struct WAITER_TAG
 {
     SM_HANDLE sm;
     SRW_LOCK_HANDLE lock;
-    THANDLE(void_ptr) data;
+    THANDLE(RC_PTR) data;
     NOTIFICATION_CALLBACK notification_callback;
     void* notification_callback_context;
     NOTIFY_COMPLETE_CALLBACK notify_complete_callback;
@@ -72,7 +69,7 @@ WAITER_HANDLE waiter_create(void)
                 }
                 else
                 {
-                    THANDLE_INITIALIZE(void_ptr)(&waiter->data, NULL);
+                    THANDLE_INITIALIZE(RC_PTR)(&waiter->data, NULL);
                     waiter->notification_callback = NULL;
                     waiter->notification_callback_context = NULL;
                     waiter->notify_complete_callback = NULL;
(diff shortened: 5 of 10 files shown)


devdoc/rc_ptr.md
```diff
@@ -0,0 +1,34 @@
...
+
+THANDLE_TYPE_DECLARE(RC_PTR);
+
+#define PRI_RC_PTR "p"
+
+#define RC_PTR_VALUE(rc) ((rc)->ptr)
+
+#define RC_PTR_AS_VOIDPTR(rc) (((rc) == NULL) ? NULL : (rc)->ptr)
+
+MOCKABLE_FUNCTION(, THANDLE(RC_PTR), rc_ptr_create_with_move_memory, const void*, ptr, RC_PTR_FREE_FUNC, free_func);
```
dcristoloveanu, line 25: >with_move_memory [](http://example.com/codeflow?start=51&length=16)

We're not really moving memory. It's more like create with onwership transfer, which is truly create in this case. with_move_memory made me think that we're copying some memory, which would imply a size too.

devdoc/rc_ptr.md
```diff
@@ -0,0 +1,34 @@
...
+```
+
+### rc_ptr_create_with_move_memory
+```c
+MOCKABLE_FUNCTION(, THANDLE(RC_PTR), rc_ptr_create_with_move_memory, const void*, ptr, RC_PTR_FREE_FUNC, free_func);
+```
+
+- `rc_ptr_create_with_move_memory` shall create a `THANDLE(RC_PTR)` that encapsulates the given `ptr`.
```
dcristoloveanu, line 33: >-  [](http://example.com/codeflow?start=0&length=2)

Likely not needed

devdoc/waiter.md
```diff
@@ -0,0 +1,71 @@
...
+
+Producers call `waiter_notify` to notify the waiter that data is available.
+
+`waiter_register_notification` and `waiter_notify` can be called from different threads in any order.
+
+## Exposed API
+```c
+#define WAITER_RESULT_VALUES \
+    WAITER_RESULT_OK, \
+    WAITER_RESULT_CANCELLED, \
```
dcristoloveanu, line 17: >CANCELLED [](http://example.com/codeflow?start=18&length=9)

I see both CANCELLED and ABANDONED. Do we need both? - it's somewhat confusing.

devdoc/waiter.md
```diff
@@ -0,0 +1,71 @@
...
+```
+
+- `waiter_create` shall create the waiter and returns it.
+
+### waiter_destroy
+```c
+    MOCKABLE_FUNCTION(, void, waiter_destroy, WAITER_HANDLE, waiter)
+```
+
+- If `waiter_register_notification` has been called previously but `waiter_notify` has not been called, `waiter_destroy` shall call the `notification_callback` given to `waiter_register_notification` with `context` as the `context` given to `waiter_register_notification`, `data` as `NULL` and `result` as `WAITER_RESULT_ABANDONED`.
```
dcristoloveanu, line 45: >-  [](http://example.com/codeflow?start=0&length=2)

I notice this doc and the other has the dashes. Do they serve any purpose?

dcristoloveanu: Ah, I see, you are using these instead of CRLF I guess.

devdoc/waiter.md
```diff
@@ -0,0 +1,71 @@
...
+### waiter_create
+```c
+    MOCKABLE_FUNCTION(, WAITER_HANDLE, waiter_create)
+```
+
+- `waiter_create` shall create the waiter and returns it.
+
+### waiter_destroy
+```c
+    MOCKABLE_FUNCTION(, void, waiter_destroy, WAITER_HANDLE, waiter)
```
dcristoloveanu, line 42: >waiter_destroy [](http://example.com/codeflow?start=30&length=14)

Missing synopsis for this function

devdoc/waiter.md
```diff
@@ -0,0 +1,71 @@
...
+    MOCKABLE_FUNCTION(, WAITER_HANDLE, waiter_create)
+    MOCKABLE_FUNCTION(, void, waiter_destroy, WAITER_HANDLE, waiter)
+    MOCKABLE_FUNCTION(, THANDLE(ASYNC_OP), waiter_register_notification, WAITER_HANDLE, waiter, NOTIFICATION_CALLBACK, notification_callback, void*, context);
+    MOCKABLE_FUNCTION(, THANDLE(ASYNC_OP), waiter_notify, WAITER_HANDLE, waiter, THANDLE(RC_PTR), data, NOTIFY_COMPLETE_CALLBACK, notify_complete_callback, void*, context);
+
+```
+
+### waiter_create
+```c
+    MOCKABLE_FUNCTION(, WAITER_HANDLE, waiter_create)
```
dcristoloveanu, line 35: >waiter_create [](http://example.com/codeflow?start=39&length=13)

Missing synopsis for this function

devdoc/waiter.md
dcristoloveanu: And everywhere

devdoc/waiter.md
dcristoloveanu, line 35: >WAITER_HANDLE [](http://example.com/codeflow?start=24&length=13)

Should this be ref counted? Sounds like an object that will be passed between some modules and it would have shared ownership.

devdoc/waiter.md
```diff
@@ -0,0 +1,71 @@
...
+- If `waiter_register_notification` has been called previously but `waiter_notify` has not been called, `waiter_destroy` shall call the `notification_callback` given to `waiter_register_notification` with `context` as the `context` given to `waiter_register_notification`, `data` as `NULL` and `result` as `WAITER_RESULT_ABANDONED`.
+- If `waiter_notify` has been called previously but `waiter_register_notification` has not been called, `waiter_destroy` shall call the `notify_complete_callback` given to `waiter_notify` with `context` as the `context` given to `waiter_notify`, and `result` as `WAITER_RESULT_ABANDONED`.
+- `waiter_destroy` shall destroy the given `waiter`.
+
+### waiter_register_notification
+```c
+    MOCKABLE_FUNCTION(, THANDLE(ASYNC_OP), waiter_register_notification, WAITER_HANDLE, waiter, NOTIFICATION_CALLBACK, notification_callback, void*, context);
+```
+- `waiter_register_notification` registers the given `notification_callback` to be called when `waiter_notify` is called.
+- If `waiter_notify` has been called previously, `waiter_register_notification` shall  call:
```
dcristoloveanu, line 54: >   [](http://example.com/codeflow?start=85&length=2)

Nit: extra space

devdoc/waiter.md
```diff
@@ -0,0 +1,71 @@
...
+    - the given `notification_callback` with `context` as the given `context`, `data` as the `data` that was given to `waiter_notify` and `result` as `WAITER_RESULT_OK`.
+    - the `notify_callback` given to `waiter_notify` with `context` as the `context` given to `waiter_notify` and `result` as `WAITER_RESULT_OK`.
+- `waiter_register_notification` shall return a `THANDLE(ASYNC_OP)` that the user can use to cancel the operation.
+- If the user calls `async_op_cancel` on the returned `THANDLE(ASYNC_OP)` before `waiter_notify` is called, the given `notification_callback` is called with `context` as the given `context`, `data` as `NULL` and `result` as `WAITER_RESULT_CANCELLED`.
```
dcristoloveanu, line 58: >WAITER_RESULT_CANCELLED [](http://example.com/codeflow?start=225&length=23)

I see, interesting

devdoc/waiter.md
```diff
@@ -0,0 +1,71 @@
...
+- If `waiter_register_notification` is called more than once, it shall fail and return `NULL`.
+
+### waiter_notify
+```c
+    MOCKABLE_FUNCTION(, THANDLE(ASYNC_OP), waiter_notify, WAITER_HANDLE, waiter, THANDLE(RC_PTR), data, NOTIFY_COMPLETE_CALLBACK, notify_complete_callback, void*, context);
+```
+- `waiter_notify` notifies the waiter that there is data available and registers the given `notify_complete_callback` to be called when the given `data` has been consumed.
```
dcristoloveanu, line 65: I see, this is the synopsis.
Maybe if you are planning to use dashes, let's have the synopsis separated and with a CRLF after it.

src/rc_ptr.c
```diff
@@ -0,0 +1,54 @@
...
+#include "c_pal/gballoc_hl.h"
+#include "c_pal/gballoc_hl_redirect.h"
+
+#include "c_pal/thandle.h"
+
+#include "c_util/rc_ptr.h"
+
+THANDLE_TYPE_DEFINE(RC_PTR);
+
+static void rc_ptr_destroy(RC_PTR* rc_ptr)
```
dcristoloveanu, line 19: >rc_ptr_destroy [](http://example.com/codeflow?start=12&length=14)

Maybe name this dispose?

src/waiter.c
```diff
@@ -0,0 +1,311 @@
...
+    if (waiter == NULL)
+    {
+        LogError("Failure in malloc(sizeof(WAITER_HANDLE))");
+        result = NULL;
+    }
+    else
+    {
+        waiter->sm = sm_create("waiter");
+        if (waiter->sm == NULL)
+        {
```
dcristoloveanu, line 49: sm seems to be an overshoot, is it really needed?

src/waiter.c
```diff
@@ -0,0 +1,311 @@
...
+    SM_HANDLE sm;
+    SRW_LOCK_HANDLE lock;
+    THANDLE(RC_PTR) data;
+    NOTIFICATION_CALLBACK notification_callback;
+    void* notification_callback_context;
+    NOTIFY_COMPLETE_CALLBACK notify_complete_callback;
+    void* notify_complete_callback_context;
+    bool registered;
+    bool notified;
+    bool callbacks_called;
```
dcristoloveanu, line 33: These don't go very well. These make up the state of the waiter and thus should be absolutely thread safe.
Sounds like they should be one int32_t volatile_atomic that you appropriately change.

src/waiter.c
```diff
@@ -0,0 +1,311 @@
...
+        }
+        free(waiter);
+    }
+all_ok:
+    return result;
+}
+
+static void abandon(void* context)
+{
+    WAITER_HANDLE waiter = (WAITER_HANDLE)context;
```
dcristoloveanu, line 94: >(WAITER_HANDLE) [](http://example.com/codeflow?start=27&length=15)

Cast not needed

src/waiter.c
dcristoloveanu: Oh, I see they are all behind a lock.

src/waiter.c
```diff
@@ -0,0 +1,311 @@
...
+            srw_lock_acquire_exclusive(waiter->lock);
+            {
+                if (waiter->registered)
+                {
+                    LogError("waiter_register_notification called twice on WAITER waiter=%p."
+                    " Previous arguments: NOTIFICATION_CALLBACK notification_callback=%p, void* context=%p"
+                    " Current arguments: NOTIFICATION_CALLBACK notification_callback=%p, void* context=%p",
+                    waiter,
+                    waiter->notification_callback, waiter->notification_callback_context,
+                    notification_callback, context
```
dcristoloveanu, line 189: Can we insert one tab?

dcristoloveanu:
I'd like to see the next iteration..

dcristoloveanu (changes requested):
:clock1:

devdoc/rc_ptr.md
```diff
@@ -0,0 +1,34 @@
...
+#define RC_PTR_VALUE(rc) ((rc)->ptr)
+
+#define RC_PTR_AS_VOIDPTR(rc) (((rc) == NULL) ? NULL : (rc)->ptr)
+
+MOCKABLE_FUNCTION(, THANDLE(RC_PTR), rc_ptr_create_with_move_memory, const void*, ptr, RC_PTR_FREE_FUNC, free_func);
+```
+
+### rc_ptr_create_with_move_memory
+```c
+MOCKABLE_FUNCTION(, THANDLE(RC_PTR), rc_ptr_create_with_move_memory, const void*, ptr, RC_PTR_FREE_FUNC, free_func);
```
anporumb, line 30: >rc_ptr_create_with_move_memory [](http://example.com/codeflow?start=37&length=30)

Errr... so what does this new module bring that `THANDLE` doesn't have already?

Apparently there's "some" memory allocated by "some" means. Now the module takes that memory and will call "free_func".

But THANDLE does all that already (including allocating the memory).

So the question is why is this new module needed...

devdoc/rc_ptr.md
anporumb, line 25: >const [](http://example.com/codeflow?start=69&length=5)

Why does it need to be "const"?

devdoc/rc_ptr.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Exposed API
+```c
+typedef void (*RC_PTR_FREE_FUNC)(void* context);
+
+typedef struct RC_PTR_TAG
+{
+    const void* ptr;
+    RC_PTR_FREE_FUNC free_func;
+} RC_PTR;
```
anporumb, line 15: Why is this struct part of the "exposed API"? Can there be some accessor to the "ptr" API? and then the whole structure can go away? 

It exposes "free_func" which I don't think is ever intended to be accessed (called) by the user.

devdoc/rc_ptr.md
```diff
@@ -0,0 +1,34 @@
...
+
+THANDLE_TYPE_DECLARE(RC_PTR);
+
+#define PRI_RC_PTR "p"
+
+#define RC_PTR_VALUE(rc) ((rc)->ptr)
+
+#define RC_PTR_AS_VOIDPTR(rc) (((rc) == NULL) ? NULL : (rc)->ptr)
```
anporumb, line 23: >RC_PTR_AS_VOIDPTR [](http://example.com/codeflow?start=8&length=17)

Suppose this is RC_PTR_VALUE_OR_NULL? ./poetry

inc/c_util/rc_ptr.h
```diff
@@ -0,0 +1,46 @@
...
+    RC_PTR_FREE_FUNC free_func;
+} RC_PTR;
+
+THANDLE_TYPE_DECLARE(RC_PTR);
+
+#define PRI_RC_PTR "p"
+
+#define RC_PTR_VALUE(rc) ((rc)->ptr)
+
+#define RC_PTR_AS_VOIDPTR(rc) (((rc) == NULL) ? NULL : (rc)->ptr)
```
anporumb, line 38: typedefs, defines can stay out of the umock_c_prod area where ideally only MOCKABLE_FUNCTIUONS should exist.

src/rc_ptr.c
```diff
@@ -0,0 +1,54 @@
...
+    {
+        LogError("Invalid arguments: RC_PTR* rc_ptr=%p", rc_ptr);
+    }
+    else
+    {
+        rc_ptr->free_func((void*)rc_ptr->ptr);
+    }
+}
+
+THANDLE(RC_PTR) rc_ptr_create_with_move_memory(const void* ptr, RC_PTR_FREE_FUNC free_func)
```
anporumb, line 31: >free_func [](http://example.com/codeflow?start=81&length=9)

Frankly, free_func might be `NULL` and still be valid, as a courtesy to the user.

What is the user passes some memory here that's stack allocated and that doesn't have a "free" function?

devdoc/waiter.md
parth21999: `WAITER_RESULT_CANCELLED` is for when the user explicitly cancels the operation. 

`WAITER_RESULT_ABANDONED` is for when the module is wants to close.

parth21999: I think they are slightly different cases, and the user might want to do differentiate between them.

src/waiter.c
parth21999: Yeah I thought about using bit manipulation like how SM does it but that would have taken longer to spike. 

We can do it like that in the final implementation.

inc/c_util/rc_ptr.h
```diff
@@ -0,0 +1,46 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#ifndef RC_PTR_H
+#define RC_PTR_H
+
+#ifdef __cplusplus
+#include <cstddef>
+#include <cstdarg>
+#else
+#include <stdarg.h>
```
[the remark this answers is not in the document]
parth21999: Removing

devdoc/waiter.md
parth21999: Just wanted some bullet points in the `.md`.

devdoc/rc_ptr.md
parth21999: Renamed

devdoc/rc_ptr.md
parth21999: This module is based on `rc_string`. 

```c
MOCKABLE_FUNCTION(, THANDLE(RC_STRING), rc_string_create_with_move_memory, const char*, string);
```

has `const`, so I made it `const`.  I think it makes sense because it tells the user that this module will not modify this memory.

parth21999: I've seen the term `with_move_memory` used when we want to create with ownership transfer. e.g,: `rc_string_create_with_move_memory`, `CONSTBUFFER_CreateWithMoveMemory`. 

So that's why I named it that.

devdoc/rc_ptr.md
parth21999: We need a refcounted `void*` for the `waiter` module. 

Either that or we templatize the entire `waiter` module.

devdoc/rc_ptr.md
parth21999: That's how `rc_string` does it. 

`free_func` will be removed from the exposed struct in the final implementation by use of an internal struct in the `.c` file. 

Just wanted some quick feedback on the API so didn't implement it that way yet.

devdoc/rc_ptr.md
```diff
@@ -0,0 +1,34 @@
...
+
+MOCKABLE_FUNCTION(, THANDLE(RC_PTR), rc_ptr_create_with_move_memory, const void*, ptr, RC_PTR_FREE_FUNC, free_func);
+```
+
+### rc_ptr_create_with_move_memory
```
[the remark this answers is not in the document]
parth21999: I've seen the term with_move_memory used when we want to create with ownership transfer. e.g,: rc_string_create_with_move_memory, CONSTBUFFER_CreateWithMoveMemory. 

So that's why I named it that.

devdoc/waiter.md
```diff
@@ -0,0 +1,71 @@
...
+```c
+    MOCKABLE_FUNCTION(, THANDLE(ASYNC_OP), waiter_register_notification, WAITER_HANDLE, waiter, NOTIFICATION_CALLBACK, notification_callback, void*, context);
+```
+- `waiter_register_notification` registers the given `notification_callback` to be called when `waiter_notify` is called.
+- If `waiter_notify` has been called previously, `waiter_register_notification` shall  call:
+    - the given `notification_callback` with `context` as the given `context`, `data` as the `data` that was given to `waiter_notify` and `result` as `WAITER_RESULT_OK`.
+    - the `notify_callback` given to `waiter_notify` with `context` as the `context` given to `waiter_notify` and `result` as `WAITER_RESULT_OK`.
+- `waiter_register_notification` shall return a `THANDLE(ASYNC_OP)` that the user can use to cancel the operation.
+- If the user calls `async_op_cancel` on the returned `THANDLE(ASYNC_OP)` before `waiter_notify` is called, the given `notification_callback` is called with `context` as the given `context`, `data` as `NULL` and `result` as `WAITER_RESULT_CANCELLED`.
+- If `waiter_register_notification` is called more than once, it shall fail and return `NULL`.
```
anporumb, line 63: >is called more than once [](http://example.com/codeflow?start=36&length=24)

I don't understand this.

This "module" is errr... only ever... handling one piece of data ever and then it shuts down?

src/waiter.c
parth21999: removed it

devdoc/waiter.md
anporumb: >has been called previously [](http://example.com/codeflow?start=21&length=26)

so if data has been produced... and it is sort of in a waiting pattern for `waiter_register_notification` to be called... 

... at the time of `waiter_register_notification` call it looks like the callback is called from the same thread which called `waiter_register_notification`. 

And in that case, ASYNC_OP that is being returned is obsolete by the time when the function returns, as there's no cancellation possible.

devdoc/waiter.md
parth21999: Yes. 1 module for 1 piece of data

devdoc/waiter.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Overview
+
+`waiter` is a module that allows the user to move data asynchronously from producer to consumer without having to block the thread.
+
+Consumers call `waiter_register_notification` to register a `NOTIFICATION_CALLBACK` function to be called when data becomes available.
+
+Producers call `waiter_notify` to notify the waiter that data is available.
+
+`waiter_register_notification` and `waiter_notify` can be called from different threads in any order.
```
anporumb, line 11: >order [](http://example.com/codeflow?start=95&length=5)

Does this mean "no concurrent calls"? What's the threading model here actually?

devdoc/waiter.md
```diff
@@ -0,0 +1,71 @@
...
+MU_DEFINE_ENUM(WAITER_RESULT, WAITER_RESULT_VALUES);
+
+typedef struct WAITER_TAG* WAITER_HANDLE;
+typedef void(*NOTIFICATION_CALLBACK)(void* context, THANDLE(RC_PTR) data, WAITER_RESULT result);
+typedef void(*NOTIFY_COMPLETE_CALLBACK)(void* context, WAITER_RESULT result);
+
+    MOCKABLE_FUNCTION(, WAITER_HANDLE, waiter_create)
+    MOCKABLE_FUNCTION(, void, waiter_destroy, WAITER_HANDLE, waiter)
+    MOCKABLE_FUNCTION(, THANDLE(ASYNC_OP), waiter_register_notification, WAITER_HANDLE, waiter, NOTIFICATION_CALLBACK, notification_callback, void*, context);
+    MOCKABLE_FUNCTION(, THANDLE(ASYNC_OP), waiter_notify, WAITER_HANDLE, waiter, THANDLE(RC_PTR), data, NOTIFY_COMPLETE_CALLBACK, notify_complete_callback, void*, context);
```
anporumb, line 29: >THANDLE(RC_PTR), data [](http://example.com/codeflow?start=81&length=21)

The interface as it is done now is forcing the producer to pack its data as a THANDLE(RC_PTR).

However, in the case where the consumer exists at the time of "notify" the "data" could be passed directly to the consumer without dictating a wrapping format.

So... what would happen if the "data" would be saved/persisted/memorized/held onto in the waiter module only in the cases where the consumer doesn't exist at the time when the data is produced?

... maybe.

devdoc/waiter.md
```diff
@@ -0,0 +1,71 @@
...
+- `waiter_register_notification` shall return a `THANDLE(ASYNC_OP)` that the user can use to cancel the operation.
+- If the user calls `async_op_cancel` on the returned `THANDLE(ASYNC_OP)` before `waiter_notify` is called, the given `notification_callback` is called with `context` as the given `context`, `data` as `NULL` and `result` as `WAITER_RESULT_CANCELLED`.
+- If `waiter_register_notification` is called more than once, it shall fail and return `NULL`.
+
+### waiter_notify
+```c
+    MOCKABLE_FUNCTION(, THANDLE(ASYNC_OP), waiter_notify, WAITER_HANDLE, waiter, THANDLE(RC_PTR), data, NOTIFY_COMPLETE_CALLBACK, notify_complete_callback, void*, context);
+```
+- `waiter_notify` notifies the waiter that there is data available and registers the given `notify_complete_callback` to be called when the given `data` has been consumed.
+- If `waiter_register_notification` has been called previously, `waiter_notify` shall call:
```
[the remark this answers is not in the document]
parth21999: 1) Yes, the `NOTIFICATION_CALLBACK` is called on the producer thread.

2) In that case the returned `ASYNC_OP` doesn't have much use because the operation would complete synchronously. However, if we have a case where `waiter_notify` is called before `waiter_register_notification`, then the user can use the `ASYNC_OP` returned by `waiter_notify` to cancel the operation.

devdoc/waiter.md
```diff
@@ -0,0 +1,71 @@
...
+```c
+#define WAITER_RESULT_VALUES \
+    WAITER_RESULT_OK, \
+    WAITER_RESULT_CANCELLED, \
+    WAITER_RESULT_ABANDONED
+
+MU_DEFINE_ENUM(WAITER_RESULT, WAITER_RESULT_VALUES);
+
+typedef struct WAITER_TAG* WAITER_HANDLE;
+typedef void(*NOTIFICATION_CALLBACK)(void* context, THANDLE(RC_PTR) data, WAITER_RESULT result);
```
anporumb, line 23: >THANDLE(RC_PTR) data [](http://example.com/codeflow?start=52&length=20)

If the producer already has "some" sort of THANDLE like THANDLE(RC_STRING) or THANDLE(blabla) then why would it need to have a wrapper over some void*? 

I believe producers and consumers are always going to use the same data format, and if that is already in THANDLE form, then that's what it should be passed here.

Maybe macros would help with that.

devdoc/waiter.md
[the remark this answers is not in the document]
parth21999: Maybe we can add a `COMPLETED` state to ASYNC_OP?

src/waiter.c
```diff
@@ -0,0 +1,311 @@
...
+
+#include "c_util/async_op.h"
+#include "c_util/rc_ptr.h"
+
+#include "c_util/waiter.h"
+
+typedef struct WAITER_TAG
+{
+    SM_HANDLE sm;
+    SRW_LOCK_HANDLE lock;
```
anporumb, line 25: do we really these 2 both?

devdoc/waiter.md
parth21999: Correct, the operation complete synchronously and the ASYNC_OP that is returned is obsolete.

However if `waiter_notify` is called before `waiter_register_notification`, then the ASYNC_OP can be used to cancel the operation.

devdoc/waiter.md
parth21999: Concurrent calls are ok.

`waiter` is thread-safe. The callbacks are called on the thread which makes the second call to `waiter_register_notification` or `waiter_notify`.

devdoc/waiter.md
parth21999: Since we are passing a THANDLE, it's not very expensive to do some extra refcount increments and decrements. So not much to be gained from the optimization where "the "data" would be saved/persisted/memorized/held onto in the waiter module only in the cases where the consumer doesn't exist at the time when the data is produced"

devdoc/waiter.md
parth21999: That would require the module to be templatized on the type of the data.

I was trying to avoid macros...

src/waiter.c
```diff
@@ -0,0 +1,311 @@
...
+    THANDLE(RC_PTR) data;
+    NOTIFICATION_CALLBACK notification_callback;
+    void* notification_callback_context;
+    NOTIFY_COMPLETE_CALLBACK notify_complete_callback;
+    void* notify_complete_callback_context;
+    bool registered;
```
anporumb, line 31: >registered [](http://example.com/codeflow?start=9&length=10)

some comments would be nice here to say what's the meaning of these variables.

src/waiter.c
```diff
@@ -0,0 +1,311 @@
...
+                    waiter->notification_callback = NULL;
+                    waiter->notification_callback_context = NULL;
+                    waiter->notify_complete_callback = NULL;
+                    waiter->notify_complete_callback_context = NULL;
+                    waiter->registered = false;
+                    waiter->notified = false;
+                    waiter->callbacks_called = false;
+                    result = waiter;
+                    goto all_ok;
+                }
```
anporumb, line 82: >} [](http://example.com/codeflow?start=16&length=1)

would be nice to have an explicit sm_close here, otherwise SM will complain that it needs to close itself before destroy.

Also helps with testing.

src/waiter.c
parth21999: Done

src/waiter.c
```diff
@@ -0,0 +1,311 @@
...
+    bool callbacks_called;
+} WAITER;
+
+WAITER_HANDLE waiter_create(void)
+{
+    WAITER_HANDLE result;
+    WAITER_HANDLE waiter = malloc(sizeof(WAITER));
+    if (waiter == NULL)
+    {
+        LogError("Failure in malloc(sizeof(WAITER_HANDLE))");
```
anporumb, line 42: >malloc [](http://example.com/codeflow?start=29&length=6)

would be nice to have the size logged

src/waiter.c
anporumb: >WAITER_HANDLE [](http://example.com/codeflow?start=28&length=13)

what happens if you remove the cast?

src/waiter.c
```diff
@@ -0,0 +1,311 @@
...
+        free(waiter);
+    }
+all_ok:
+    return result;
+}
+
+static void abandon(void* context)
+{
+    WAITER_HANDLE waiter = (WAITER_HANDLE)context;
+    srw_lock_acquire_exclusive(waiter->lock);
```
anporumb, line 95: >lock [](http://example.com/codeflow?start=39&length=4)

I've taken a brief lookat this lock. Seems to serialize all the calls in the module (and then what's the threading model...)

Then why is SM needed...

src/waiter.c
parth21999: Removed sm altogether

src/waiter.c
parth21999: Done

src/waiter.c
```diff
@@ -0,0 +1,311 @@
...
+    if (!waiter->callbacks_called)
+    {
+        sm_fault(waiter->sm);
+        if (waiter->registered)
+        {
+            waiter->notification_callback(waiter->notification_callback_context, NULL, WAITER_RESULT_ABANDONED);
+            sm_exec_end(waiter->sm);
```
anporumb, line 102: >sm_exec_end [](http://example.com/codeflow?start=12&length=11)

I cannot in good conscience not notice how intentionally we're still using the sm **after** we've faulted in in line 98.

src/waiter.c
anporumb, line 95: >lock [](http://example.com/codeflow?start=39&length=4)

Why does abandon even need this lock since abandon is only called from _destroy, and _destroy is never supposed to be thread-safe in our architecture.

src/waiter.c
```diff
@@ -0,0 +1,311 @@
...
+        THANDLE_ASSIGN(RC_PTR)(&waiter->data, NULL);
+        sm_exec_end(waiter->sm);
+        waiter->notify_complete_callback(waiter->notify_complete_callback_context, WAITER_RESULT_OK);
+    }
+}
+
+THANDLE(ASYNC_OP) waiter_register_notification(WAITER_HANDLE waiter, NOTIFICATION_CALLBACK notification_callback, void* context)
+{
+    THANDLE(ASYNC_OP) result = NULL;
+    if (waiter == NULL || notification_callback == NULL || context == NULL)
```
anporumb, line 165: >waiter == NULL || notification_callback == NULL || context == NULL [](http://example.com/codeflow?start=8&length=66)

usualyl we write these in separate lines cause the SRS_blabla will come before each of them.

anporumb, line 165: >context == NULL [](http://example.com/codeflow?start=59&length=15)

why cannot `context` be `NULL`? It comes from the user, doesn't it?

src/waiter.c
```diff
@@ -0,0 +1,311 @@
...
+    {
+        LogError("Invalid arguments: WAITER_HANDLE waiter = %p, NOTIFICATION_CALLBACK notification_callback = %p, void* context = %p",
+            waiter, notification_callback, context);
+    }
+    else
+    {
+        SM_RESULT sm_exec_begin_result = sm_exec_begin(waiter->sm);
```
anporumb, line 172: >sm_exec_begin [](http://example.com/codeflow?start=41&length=13)

I cannot figure out the relation between SM and the "lock" and why both are needed.

src/waiter.c
parth21999: Removed.

src/waiter.c
```diff
@@ -0,0 +1,311 @@
...
+            LogError("Failure in sm_exec_begin(waiter->sm=%p)", waiter->sm);
+        }
+        else
+        {
+            bool success;
+            srw_lock_acquire_exclusive(waiter->lock);
+            {
+                if (waiter->registered)
+                {
+                    LogError("waiter_register_notification called twice on WAITER waiter=%p."
```
anporumb: >WAITER [](http://example.com/codeflow?start=75&length=6)

WAITER_HANDLE maybe?

src/waiter.c
parth21999: removed sm

src/waiter.c
```diff
@@ -0,0 +1,311 @@
...
+    return result;
+}
+
+static void abandon(void* context)
+{
+    WAITER_HANDLE waiter = (WAITER_HANDLE)context;
+    srw_lock_acquire_exclusive(waiter->lock);
+    if (!waiter->callbacks_called)
+    {
+        sm_fault(waiter->sm);
```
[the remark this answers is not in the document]
parth21999: removed sm

src/waiter.c
```diff
@@ -0,0 +1,311 @@
...
+                    notification_callback, context
+                    );
+                    success = false;
+                }
+                else
+                {
+                    THANDLE(ASYNC_OP) async_op = async_op_create(cancel_register_notification, sizeof(WAITER_HANDLE), alignof(WAITER_HANDLE), dispose);
+                    if (async_op == NULL)
+                    {
+                        LogError("Failure in async_op_create(cancel_register_notification=%p, sizeof(WAITER_HANDLE)=%" PRIu32 ", alignof(WAITER_HANDLE)=%" PRIu32 ", dispose=%p)", cancel_register_notification, (uint32_t)sizeof(WAITER_HANDLE), (uint32_t)alignof(WAITER_HANDLE), dispose);
```
anporumb, line 199: >PRIu32 [](http://example.com/codeflow?start=119&length=6)

zu

anporumb, line 199: >PRIu32 [](http://example.com/codeflow?start=155&length=6)

zu, (i think)

src/waiter.c
parth21999: removing lock from abandon

src/waiter.c
```diff
@@ -0,0 +1,311 @@
...
+                        success = false;
+                    }
+                    else
+                    {
+                        ((ASYNC_OP*)async_op)->context = waiter;
```
anporumb, line 204: >context = waiter; [](http://example.com/codeflow?start=47&length=17)

Every time I see handle=handle my heart skips a beat.

anporumb, line 204: >(ASYNC_OP*) [](http://example.com/codeflow?start=25&length=11)

is this cast necessary?

src/waiter.c
```diff
@@ -0,0 +1,311 @@
...
+}
+
+static void dispose(void* context)
+{
+    (void)context;
+}
+
+static void try_complete(WAITER_HANDLE waiter)
+{
+    if (waiter->registered && waiter->notified && !waiter->callbacks_called)
```
anporumb, line 150: >registered [](http://example.com/codeflow?start=16&length=10)

how does this line execute and `waiter->registered` is `false`?

src/waiter.c
parth21999: Fixed

src/waiter.c
anporumb, line 150: >(waiter->registered [](http://example.com/codeflow?start=7&length=19)

I had a comment here that I suppose needs to be deleted, but I cannot find it anymore.

src/waiter.c
```diff
@@ -0,0 +1,311 @@
...
+    {
+        sm_fault(waiter->sm);
```
anporumb, line 152: >sm_fault [](http://example.com/codeflow?start=8&length=8)

why would we FAULT the SM here? This is not exactly an error/fault/failure case??

src/waiter.c
```diff
@@ -0,0 +1,311 @@
...
+        waiter->callbacks_called = true;
+        sm_exec_end(waiter->sm);
+        waiter->notification_callback(waiter->notification_callback_context, waiter->data, WAITER_RESULT_OK);
```
anporumb, line 155: >notification_callback [](http://example.com/codeflow?start=16&length=21)

can this callback call any of the APIs of the module?

src/waiter.c
```diff
@@ -0,0 +1,311 @@
...
+        THANDLE_ASSIGN(RC_PTR)(&waiter->data, NULL);
+        sm_exec_end(waiter->sm);
+        waiter->notify_complete_callback(waiter->notify_complete_callback_context, WAITER_RESULT_OK);
```
anporumb, line 158: >notify_complete_callback [](http://example.com/codeflow?start=16&length=24)

can this callback call any of the APIs of the module?

src/waiter.c
```diff
@@ -0,0 +1,311 @@
...
+
+static void cancel_register_notification(void* context)
+{
+    WAITER_HANDLE waiter = context;
+    srw_lock_acquire_exclusive(waiter->lock);
+    {
+        waiter->notification_callback(waiter->notification_callback_context, NULL, WAITER_RESULT_CANCELLED);
+        waiter->callbacks_called = true;
+        sm_fault(waiter->sm);
+        sm_exec_end(waiter->sm);
```
anporumb, line 138: >sm_exec_end [](http://example.com/codeflow?start=8&length=11)

again sm_exec_end after fault.

src/waiter.c
```diff
@@ -0,0 +1,311 @@
...
+                    " Current arguments: NOTIFICATION_CALLBACK notification_callback=%p, void* context=%p",
+                    waiter,
+                    waiter->notification_callback, waiter->notification_callback_context,
+                    notification_callback, context
+                    );
+                    success = false;
+                }
+                else
+                {
+                    THANDLE(ASYNC_OP) async_op = async_op_create(cancel_register_notification, sizeof(WAITER_HANDLE), alignof(WAITER_HANDLE), dispose);
```
anporumb, line 196: >dispose [](http://example.com/codeflow?start=142&length=7)

what happens if you pass `NULL` here?

src/waiter.c
parth21999: Yes, I get "error C2166: l-value specifies const object" otherwise

devdoc/waiter.md
anporumb, line 11: >order [](http://example.com/codeflow?start=95&length=5)

what's the reentrancy model here?

src/waiter.c
parth21999: async_op_create takes uint32_t

parth21999: async_op_create takes uint32_t

src/waiter.c
```diff
@@ -0,0 +1,311 @@
...
+    return result;
+}
+
+static void cancel_notify(void* context)
+{
+    WAITER_HANDLE waiter = context;
+    srw_lock_acquire_exclusive(waiter->lock);
+    {
+        waiter->notify_complete_callback(waiter->notify_complete_callback_context, WAITER_RESULT_CANCELLED);
+        waiter->callbacks_called = true;
```
anporumb, line 237: >waiter->callbacks_called = true; [](http://example.com/codeflow?start=8&length=32)

It is funny to see some plural "callback*s*_call" and it is just 1...

src/waiter.c
```diff
@@ -0,0 +1,311 @@
...
+        sm_fault(waiter->sm);
+        sm_exec_end(waiter->sm);
+    }
+    srw_lock_release_exclusive(waiter->lock);
+}
+
+THANDLE(ASYNC_OP) waiter_notify(WAITER_HANDLE waiter, THANDLE(RC_PTR) data, NOTIFY_COMPLETE_CALLBACK notify_complete_callback, void* context)
+{
+    THANDLE(ASYNC_OP) result = NULL;
+    if (waiter == NULL || data == NULL || notify_complete_callback == NULL || context == NULL)
```
anporumb, line 248: >context [](http://example.com/codeflow?start=78&length=7)

why cannot `context` be `NULL`?

src/waiter.c
```diff
@@ -0,0 +1,311 @@
...
+            LogError("Failure in sm_exec_begin(waiter->sm=%p)", waiter->sm);
+        }
+        else
+        {
+            bool success;
+            srw_lock_acquire_exclusive(waiter->lock);
+            {
+                if (waiter->notified)
+                {
+                    LogError("waiter_notify called twice on WAITER waiter=%p."
```
anporumb, line 267: >WAITER [](http://example.com/codeflow?start=60&length=6)

WAITER_HANDLE

src/waiter.c
parth21999: Changing to NULL

src/waiter.c
```diff
@@ -0,0 +1,311 @@
...
+                        "Current arguments: THANDLE(RC_PTR) data=%" PRI_RC_PTR ", NOTIFY_COMPLETE_CALLBACK notify_complete_callback=%p, void* context=%p.",
+                        waiter,
+                        waiter->data, waiter->notify_complete_callback, waiter->notify_complete_callback_context,
+                        data, notify_complete_callback, context
+                    );
+                    success = false;
+                }
+                else
+                {
+                    THANDLE(ASYNC_OP) async_op = async_op_create(cancel_notify, sizeof(WAITER_HANDLE), alignof(WAITER_HANDLE), dispose);
```
anporumb, line 278: >dispose [](http://example.com/codeflow?start=127&length=7)

what happens if this parameter is `NULL`?

src/waiter.c
```diff
@@ -0,0 +1,311 @@
...
+                if (waiter->notified)
+                {
+                    LogError("waiter_notify called twice on WAITER waiter=%p."
+                        "Previous arguments: THANDLE(RC_PTR) data=%" PRI_RC_PTR ", NOTIFY_COMPLETE_CALLBACK notify_complete_callback=%p, void* context=%p."
+                        "Current arguments: THANDLE(RC_PTR) data=%" PRI_RC_PTR ", NOTIFY_COMPLETE_CALLBACK notify_complete_callback=%p, void* context=%p.",
+                        waiter,
+                        waiter->data, waiter->notify_complete_callback, waiter->notify_complete_callback_context,
+                        data, notify_complete_callback, context
+                    );
+                    success = false;
```
anporumb, line 274: >false [](http://example.com/codeflow?start=30&length=5)

can this code be written without "success"?

src/waiter.c
```diff
@@ -0,0 +1,311 @@
...
+                    if (async_op == NULL)
+                    {
+                        LogError("Failure in async_op_create(cancel_notify=%p, sizeof(WAITER_HANDLE)=%" PRIu32 ", alignof(WAITER_HANDLE)=%" PRIu32 ", dispose=%p)", cancel_notify, (uint32_t)sizeof(WAITER_HANDLE), (uint32_t)alignof(WAITER_HANDLE), dispose);
+                        success = false;
+                    }
+                    else
+                    {
+                        ((ASYNC_OP*)async_op)->context = waiter;
+
+                        THANDLE_ASSIGN(RC_PTR)(&waiter->data, data);
```
anporumb, line 288: >THANDLE_ASSIGN [](http://example.com/codeflow?start=24&length=14)

from all the THANDLE APIs, which one should be used to initialize a variable?

Hint: answer is in the question.

src/waiter.c
parth21999: Removed sm. 

Used sm_fault here because don't want module to be reused.

src/waiter.c
parth21999: No. No reentry allowed.

src/waiter.c
parth21999: No. No reentry allowed.

src/waiter.c
parth21999: Removed sm

src/waiter.c
```diff
@@ -0,0 +1,311 @@
...
+                        THANDLE_INITIALIZE_MOVE(ASYNC_OP)(&result, &async_op);
+                        success = true;
+                    }
+                }
+            }
+            srw_lock_release_exclusive(waiter->lock);
+
+            if (!success)
+            {
+                LogError("Failure in waiter_register_notification(waiter=%p, notification_callback=%p, context=%p)", waiter, notification_callback, context);
```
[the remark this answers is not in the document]
parth21999: Removed.

src/waiter.c
parth21999: Yeah it really should be callback(s)_called.

src/waiter.c
parth21999: Yes, done in next iteration.

src/waiter.c
```diff
@@ -0,0 +1,311 @@
...
+            else
+            {
+                goto all_ok;
+            }
+            sm_exec_end(waiter->sm);
+        }
+    }
+all_ok:
+    return result;
+}
```
anporumb, line 312: so after going through this code, I do have to wonder what happens when:

1) _create
2) _notify
3) _cancel_notify
4) _notify again.

As I see it, this is perfectly valid timeout scenario where the user wants its data to be consumed within a time frame OR if that didn't happen, to cancel the said data consumption and re-publish some newer version of the data for consumption.

If any of the ASYNC_OP cancels blocks the lifetime of the module, they might as well not be needed and the user should call _destroy and _create immediately after.

src/waiter.c
parth21999: Removed dispose.

src/waiter.c
parth21999: Done

src/waiter.c
parth21999: `try_complete` is called by `waiter_register_notification` and `waiter_notify`. In case `waiter_notify` is called before `waiter_register_notification`, this line will execute with `waiter->registered` as `false`.

src/waiter.c
anporumb, line 312: waiting because
1) curiosity
2) I think too many ASYNC_OPs are created
3) reentrancy is not fully explained/tested
4) sm+lock != love. So: threading model.

src/waiter.c
parth21999: Fixed.

src/waiter.c
parth21999: fixed

src/waiter.c
parth21999: Fixed.

src/waiter.c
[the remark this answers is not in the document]
parth21999: Yes. Changing

devdoc/waiter.md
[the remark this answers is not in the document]
parth21999: Return async_op as out arg, return enum that states sync complete, async complete, error.

src/waiter.c
parth21999: cancel should undo the effect of the operation (notify/register). Module should not enter terminal state upon cancellation.

devdoc/rc_ptr.md
parth21999: Implementing with THANDLE(RC_PTR) for now, will templatize later

PR #281, opened 2023-05-19
`waiter` header and specs

parth21999:
header and specs for `waiter` module.

Also has a sample implementation and tests for sanity check.


devdoc/waiter.md
```diff
@@ -0,0 +1,231 @@
+# `waiter` requirements
+
+## Overview
+
+`waiter` is a module that allows the user to move data asynchronously from producer to consumer without having to block the thread.
```
anporumb, line 5: >move data asynchronously from producer to consumer [](http://example.com/codeflow?start=45&length=50)

Poetry:

nothing here "waits". So the module being called "waiter" does not convey what "it does". Consider other poetry. Here's some other possible names: "buffer", "broker", "channel"

devdoc/waiter.md
```diff
@@ -0,0 +1,231 @@
...
+    N --> E:  cancel waiter_notify
+    R --> E: cancel waiter_register_notification
+
+    R --> R: waiter_register_notification
+    N --> N: waiter_notify
+```
+
+## Reentrancy
+
+`waiter` is reentrant. This means that `waiter_register_notification` and `waiter_notify` can be called from `NOTIFICATION_CALLBACK` and `NOTIFY_COMPLETE_CALLBACK` functions.
```
anporumb, line 44: >`waiter` is reentrant [](http://example.com/codeflow?start=0&length=21)

nice

anporumb, line 44: >can be called from [](http://example.com/codeflow?start=90&length=18)

I suppose that any API can be called from any callback, so if somehow the threads gets switched and on the producer's callback the consumer API is called and on the consumer's callback the producer API is called - they still work.

devdoc/waiter.md
```diff
@@ -0,0 +1,231 @@
...
+#include "umock_c/umock_c_prod.h"
+#ifdef __cplusplus
+extern "C" {
+#endif /* __cplusplus */
+
+typedef void(*NOTIFICATION_CALLBACK)(void* context, THANDLE(RC_PTR) data, WAITER_CALLBACK_RESULT result);
+typedef void(*NOTIFY_COMPLETE_CALLBACK)(void* context, WAITER_CALLBACK_RESULT result);
+typedef struct WAITER_TAG WAITER;
+
+THANDLE_TYPE_DECLARE(WAITER);
```
anporumb, line 72: this is nice, completely hiding the WAITER's internal fields from the outside world so nobody can peek what's inside.

devdoc/waiter.md
```diff
@@ -0,0 +1,231 @@
...
+
+    MOCKABLE_FUNCTION(, THANDLE(WAITER), waiter_create)
+    MOCKABLE_FUNCTION(, WAITER_RESULT, waiter_register_notification, THANDLE(WAITER), waiter, NOTIFICATION_CALLBACK, notification_callback, void*, context, THANDLE(ASYNC_OP)*, out_op);
```
anporumb, line 75: >out_op [](http://example.com/codeflow?start=176&length=6)

just so they don't get confused, consider maybe calling this "out_op_consumer" and the other one "out_op_producer"

devdoc/waiter.md
```diff
@@ -0,0 +1,231 @@
...
+**SRS_WAITER_43_001: [** `waiter_create` shall call `THANDLE_MALLOC` with `dispose` as `waiter_dispose`. **]**
+
+**SRS_WAITER_43_002: [** If there are any failures, `waiter_create` shall fail and return `NULL`. **]**
+
+### waiter_dispose
+```c
+    void waiter_dispose(WAITER* waiter)
+```
+
+**SRS_WAITER_43_003: [** `waiter_dispose` disposes the given `waiter`. **]**
```
anporumb, line 100: >SRS_WAITER_43_003 [](http://example.com/codeflow?start=2&length=17)

no need for SRS in synopsis.

devdoc/waiter.md
```diff
@@ -0,0 +1,231 @@
...
+- **SRS_WAITER_43_044: [** call the `notify_complete_callback` given to `waiter_notify` with `context` as the `context` given to `waiter_notify`, and `result` as `WAITER_RESULT_ABANDONED`. **]**
+
+- **SRS_WAITER_43_045: [** release the reference to `data` stored in `waiter` by calling `THANDLE_ASSIGN`. **]**
+
+### waiter_register_notification
+```c
+    MOCKABLE_FUNCTION(, WAITER_RESULT, waiter_register_notification, THANDLE(WAITER), waiter, NOTIFICATION_CALLBACK, notification_callback, void*, context, THANDLE(ASYNC_OP)*, out_op);
+```
+
+`waiter_register_notification` registers the given `notification_callback` to be called when `waiter_notify` is called.
```
anporumb, line 117: >when `waiter_notify` is called [](http://example.com/codeflow?start=88&length=30)

hmmmm. Consider adding here some explanatory wording around "when data has been consumed/is preparing to be consumed"...

devdoc/waiter.md
```diff
@@ -0,0 +1,231 @@
...
+
+**SRS_WAITER_43_009: [** If `out_op` is `NULL`, `waiter_register_notification` shall fail and return `WAITER_RESULT_ERROR`. **]**
+
+**SRS_WAITER_43_010: [** `waiter_register_notification` shall call `interlocked_or` to set the `REGISTERED` bit and obtain the current state of the `waiter`. **]**
+
+**SRS_WAITER_43_011: [** If the `REGISTERED` bit was already set in the current state, `waiter_register_notification` shall fail and return `WAITER_RESULT_REFUSED`. **]**
+
+**SRS_WAITER_43_012: [** If the `NOTIFIED` bit is set, `waiter_register_notification` shall : **]**
+
+- **SRS_WAITER_43_042: [** call `interlocked_exchange` to increment the `EPOCH` and unset the `NOTIFIED` bit and `REGISTERED` bit. **]**
```
anporumb, line 131: >** call `interlocked_exchange` to increment the `EPOCH` and unset the `NOTIFIED` bit and `REGISTERED` bit. **]** [](http://example.com/codeflow?start=24&length=112)

So... I see these bits getting cleared (and epoch getting incremented) before the callbacks are called. 

So immediately after this, new callbacks and APIs can be called. What happens with the currently registered callbacks that apparently haven't been called... ?

devdoc/waiter.md
```diff
@@ -0,0 +1,231 @@
...
+- **SRS_WAITER_43_015: [** return `WAITER_RESULT_SYNC`. **]**
+
+
+**SRS_WAITER_43_016: [** If the `NOTIFIED` bit was not already set, `waiter_register_notification` shall: **]**
+
+- **SRS_WAITER_43_017: [** call `async_op_create` with `cancel_register_notification` as the cancel function. **]**
+
+- **SRS_WAITER_43_018: [** store the given `notification_callback` and the given `context` in the `waiter`. **]**
+
+- **SRS_WAITER_43_019: [** store the current state of the `waiter` in the created `THANDLE(ASYNC_OP)`. **]**
```
anporumb, line 148: >current state [](http://example.com/codeflow?start=37&length=13)

when is this "current state" obtained? Is is "obtained just now", or is it the state obtained in 43_010?

devdoc/waiter.md
```diff
@@ -0,0 +1,231 @@
...
+
+- **SRS_WAITER_43_020: [** store the given `waiter` in the created `THANDLE(ASYNC_OP)` by calling `THANDLE_INITIALIZE`. **]**
+
+- **SRS_WAITER_43_021: [** store the created `THANDLE(ASYNC_OP)` in `out_op`by calling `THANDLE_INITIALIZE_MOVE`. **]**
```
anporumb, line 152: >`b [](http://example.com/codeflow?start=75&length=2)

space

devdoc/waiter.md
```diff
@@ -0,0 +1,231 @@
...
+
+- **SRS_WAITER_43_042: [** call `interlocked_exchange` to increment the `EPOCH` and unset the `NOTIFIED` bit and `REGISTERED` bit. **]**
+
+- **SRS_WAITER_43_013: [** call the given `notification_callback` with `context` as the given `context`, `data` as the `data` that was given to `waiter_notify` and `result` as `WAITER_RESULT_OK`. **]**
+
+- **SRS_WAITER_43_053: [** release the reference to `data` stored in `waiter` by calling `THANDLE_ASSIGN`. **]**
+
+- **SRS_WAITER_43_014: [** call the `notify_complete_callback` given to `waiter_notify` with `context` as the `context` given to `waiter_notify` and `result` as `WAITER_RESULT_OK`. **]**
+
+- **SRS_WAITER_43_015: [** return `WAITER_RESULT_SYNC`. **]**
```
anporumb, line 139: >WAITER_RESULT_SYNC [](http://example.com/codeflow?start=35&length=18)

somewhere should say that it sets `out_op` to `NULL`, just for consistency.

devdoc/waiter.md
anporumb: like line 177

devdoc/waiter.md
```diff
@@ -0,0 +1,231 @@
...
+
+**SRS_WAITER_43_010: [** `waiter_register_notification` shall call `interlocked_or` to set the `REGISTERED` bit and obtain the current state of the `waiter`. **]**
+
+**SRS_WAITER_43_011: [** If the `REGISTERED` bit was already set in the current state, `waiter_register_notification` shall fail and return `WAITER_RESULT_REFUSED`. **]**
+
+**SRS_WAITER_43_012: [** If the `NOTIFIED` bit is set, `waiter_register_notification` shall : **]**
+
+- **SRS_WAITER_43_042: [** call `interlocked_exchange` to increment the `EPOCH` and unset the `NOTIFIED` bit and `REGISTERED` bit. **]**
+
+- **SRS_WAITER_43_013: [** call the given `notification_callback` with `context` as the given `context`, `data` as the `data` that was given to `waiter_notify` and `result` as `WAITER_RESULT_OK`. **]**
```
anporumb, line 133: >notification_callback [](http://example.com/codeflow?start=43&length=21)

are these callbacks part of "current state"?

devdoc/waiter.md
```diff
@@ -0,0 +1,231 @@
...
+
+- **SRS_WAITER_43_032: [** return `WAITER_RESULT_SYNC`. **]**
+
+**SRS_WAITER_43_033: [** If the `REGISTERED` bit was not already set, `waiter_notify` shall: **]**
+
+- **SRS_WAITER_43_034: [** call `async_op_create` with `cancel_notify` as the cancel function. **]**
+
+- **SRS_WAITER_43_035: [** store the given `notify_complete_callback` and the given `context` in the `waiter`. **]**
+
+- **SRS_WAITER_43_036: [** store the given `data` in the `waiter` by calling `THANDLE_INITIALIZE`. **]**
```
anporumb, line 205: >store the given `data` in the `waiter` [](http://example.com/codeflow?start=27&length=38)

That's a bit weird to store this data in the waiter, as... I would see the data more related to the async_op more than the module itself. Interesting.

devdoc/waiter.md
```diff
@@ -0,0 +1,231 @@
...
+**SRS_WAITER_43_023: [** If there are any failures, `waiter_register_notification` shall fail and return `WAITER_RESULT_ERROR`. **]**
+
+### cancel_register_notification
+```
+static void cancel_register_notification(void* context)
+```
+
+`cancel_register_notification` is the cancel function given to `async_op_create` when `waiter_register_notification` is called.
+
+**SRS_WAITER_43_046: [** If the current state of the `waiter` is the same as the initial state of the `waiter` when `waiter_register_notification` was called, `cancel_register_notification` shall: **]**
```
anporumb, line 165: >f the current state of the `waiter` is the same [](http://example.com/codeflow?start=26&length=47)

and if it isn't? Suppose it doesn't do anything.

devdoc/rc_ptr.md
```diff
@@ -0,0 +1,35 @@
...
+
+`rc_ptr` is a module that encapsulates reference-counted pointers.
+
+## Exposed API
+```c
+typedef void (*RC_PTR_FREE_FUNC)(void* context);
+
+typedef struct RC_PTR_TAG
+{
+    const void* ptr;
```
anporumb, line 13: >const [](http://example.com/codeflow?start=4&length=5)

likely this is not needed. `void*` can only be used meaningfully after dereference, and at the time of dereferencing it, the pointer type is known - and that pointer type should carry the qualifier rather than here.

src/rc_ptr.c
```diff
@@ -0,0 +1,57 @@
...
+#include "c_util/rc_ptr.h"
+
+THANDLE_TYPE_DEFINE(RC_PTR);
+
+static void dispose(RC_PTR* rc_ptr)
+{
+    if (rc_ptr == NULL)
+    {
+        LogError("Invalid arguments: RC_PTR* rc_ptr=%p", rc_ptr);
+    }
```
anporumb, line 24: was this line of code ever covered by any test/can it be covered?

devdoc/rc_ptr.md
```diff
@@ -0,0 +1,35 @@
...
+
+THANDLE_TYPE_DECLARE(RC_PTR);
+
+#define PRI_RC_PTR "p"
+
+#define RC_PTR_VALUE(rc) ((rc)->ptr)
+
+#define RC_PTR_AS_VOIDPTR(rc) (((rc) == NULL) ? NULL : (rc)->ptr)
+
+MOCKABLE_FUNCTION(, THANDLE(RC_PTR), rc_ptr_create_with_move_memory, const void*, ptr, RC_PTR_FREE_FUNC, free_func);
```
anporumb, line 25: >move_memory [](http://example.com/codeflow?start=56&length=11)

still there's no "memory" moved. It there was ANY memory moved, its size would have been known here.

Consider remove "memory" ./poetry

src/waiter.c
```diff
@@ -0,0 +1,263 @@
...
+    {
+        LogError("Invalid arguments: WAITER_HANDLE waiter = %p, NOTIFICATION_CALLBACK notification_callback = %p, void* context = %p, THANDLE(ASYNC_OP)* out_op = %p",
+            waiter, notification_callback, context, out_op);
+        result = WAITER_RESULT_ERROR;
+    }
+    else
+    {
+        WAITER* waiter_ptr = THANDLE_GET_T(WAITER)(waiter);
+        int32_t state = interlocked_or(&waiter_ptr->state, WAITER_REGISTERED);
+        if (state & WAITER_REGISTERED)
```
anporumb, line 129: so, in general when talking about interlocked and "state" *all* decision and inspection of the "state" should ensure that the state doesn't change in between 2 such reads.

Now in this case between lines 128 and 129 , nothing is prohibiting `waiter_notify` to be called. And I don't know what happens in that case, but I suppose somethign along inconsistency.

Also title of code review says "header and md" and those have been reviewed so... Iw ill stop here looking in the code.

Also for the next review (the one with code/tests/reals/UT/int) don't forget to add "chaos".

anporumb: And remove code in this iterations.

anporumb: I am changing my state to "waiting" because I think my comments might require some more work.

anporumb (changes requested):
:clock1:

devdoc/waiter.md
parth21999: Fixed.

devdoc/waiter.md
parth21999: Current callback and context will be stored in local variables before bits are cleared.

devdoc/waiter.md
parth21999: Yes, these callbacks are stored in the waiter and are saved locally before bits are reset.

devdoc/waiter.md
parth21999: In 43_010

inc/c_util/rc_ptr.h
```diff
@@ -0,0 +1,45 @@
...
+#ifdef __cplusplus
+#include <cstddef>
+#else
+#include <stddef.h>
+#endif
+
+#include "c_pal/thandle.h"
+
+
+
```
dcristoloveanu, line 16: Nitpick: some extra newlines.

devdoc/waiter.md
```diff
@@ -0,0 +1,231 @@
...
+
+**SRS_WAITER_43_006: [** If the `NOTIFIED` bit is set, `waiter_dispose` shall: **]**
+
+- **SRS_WAITER_43_044: [** call the `notify_complete_callback` given to `waiter_notify` with `context` as the `context` given to `waiter_notify`, and `result` as `WAITER_RESULT_ABANDONED`. **]**
+
+- **SRS_WAITER_43_045: [** release the reference to `data` stored in `waiter` by calling `THANDLE_ASSIGN`. **]**
+
+### waiter_register_notification
+```c
+    MOCKABLE_FUNCTION(, WAITER_RESULT, waiter_register_notification, THANDLE(WAITER), waiter, NOTIFICATION_CALLBACK, notification_callback, void*, context, THANDLE(ASYNC_OP)*, out_op);
```
dcristoloveanu, line 114: >out_op [](http://example.com/codeflow?start=176&length=6)

I like the out_ prefix. It makes me better aware that this is an out argument.

devdoc/waiter.md
```diff
@@ -0,0 +1,231 @@
...
+**SRS_WAITER_43_008: [** If `notification_callback` is `NULL`, `waiter_register_notification` shall fail and return `WAITER_RESULT_ERROR`. **]**
+
+**SRS_WAITER_43_009: [** If `out_op` is `NULL`, `waiter_register_notification` shall fail and return `WAITER_RESULT_ERROR`. **]**
+
+**SRS_WAITER_43_010: [** `waiter_register_notification` shall call `interlocked_or` to set the `REGISTERED` bit and obtain the current state of the `waiter`. **]**
+
+**SRS_WAITER_43_011: [** If the `REGISTERED` bit was already set in the current state, `waiter_register_notification` shall fail and return `WAITER_RESULT_REFUSED`. **]**
+
+**SRS_WAITER_43_012: [** If the `NOTIFIED` bit is set, `waiter_register_notification` shall : **]**
+
```
dcristoloveanu, line 130: What if both bits are set? Can that happen? I guess I shall see.

devdoc/waiter.md
dcristoloveanu, line 131: >interlocked_exchange [](http://example.com/codeflow?start=33&length=20)

This blind interlocked exchange here tells me that this is a race condition waiting to happen.
You can only do an interlocked_exchange in the case that you know 100% that there is no way that this value will be changed by other threads. Otherwise, you will overwrite what the other threads are changing.

dcristoloveanu, line 131: >interlocked_exchange [](http://example.com/codeflow?start=33&length=20)

interlocked_compare_exchange

dcristoloveanu, line 131: >rlocked_exchange` to increment the `EPOCH` and unset the `NOTIFIED` bit and `REGISTERED` bit.  [](http://example.com/codeflow?start=37&length=94)

Aha, so both bits set means barrier - as in no APIs coming in would actually go on with incrementing the epoch.

Maybe this should be spelled out as a note for example.

devdoc/waiter.md
```diff
@@ -0,0 +1,231 @@
...
+- **SRS_WAITER_43_042: [** call `interlocked_exchange` to increment the `EPOCH` and unset the `NOTIFIED` bit and `REGISTERED` bit. **]**
+
+- **SRS_WAITER_43_013: [** call the given `notification_callback` with `context` as the given `context`, `data` as the `data` that was given to `waiter_notify` and `result` as `WAITER_RESULT_OK`. **]**
+
```
dcristoloveanu, line 134: Here is the problem:
once the epoch was changed and the bits were reset, the data that is stored is not valid anymore (the next microsecond after the .SRS_WAITER_43_042 spec was executed, another call to notify comes in and stores the data overwriting your existing data.

The only way I believe you can avoid this is by having barriers for putting and taking the data out of the waiter.

Think the following states:
- NO_DATA
- NOTIFYING - stores notify data (the data, etc.)
- NOTIFIED
- REGISTERING - stores register_notification data (callback)
- REGISTERED
- NOTIFIED_REGISTERING - takes out data preparing to call  notification_callback
- REGISTER_NOTIFYING - takes out data preparing to call notification_callback

The *ING states should be short and always be treated as a lock.

--------

An alternative would be to use a mutex/lock

----------

Another alternative solution (maybe smarter but more complex) would be to bundle everything in a structure and use a pointer to make sure that you atomically swap the state and the data.

On Windows (and others), due to malloc alignment you can use the least significant bits to store some state (pretty sure you can use 2 bits there to store the states you want.

dcristoloveanu:
I don't think this works as is, so shall wait for the next iteration

dcristoloveanu (changes requested):
:clock1:

devdoc/waiter.md
parth21999: The `waiter` needs to be able to access the data to complete the operation. `waiter` cannot reference the async_op because the async_op references the `waiter` (otherwise there will be a circular reference and memory leak). So the data must be stored in the `waiter`.

devdoc/waiter.md
parth21999: If the state has changed then nothing needs to be done.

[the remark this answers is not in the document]
parth21999: EPOCH and state, added clarification.

devdoc/waiter.md
parth21999: Changed to use enum as state in next iteration

devdoc/waiter.md
```diff
@@ -0,0 +1,231 @@
...
+
+### cancel_notify
+```c
+static void cancel_notify(void* context)
+```
+
+`cancel_notify` is the cancel function given to `async_op_create` when `waiter_notify` is called.
+
+**SRS_WAITER_43_049: [** If the current state of the `waiter` is the same as the initial state of the `waiter` when `waiter_notify` was called, `cancel_notify` shall: **]**
```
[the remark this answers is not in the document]
parth21999: Complete state, added clarification

devdoc/waiter.md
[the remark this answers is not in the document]
parth21999: complete state, added clarification.

devdoc/waiter.md
parth21999: I like channel. Reminds me of Golang channels

devdoc/rc_ptr.md
parth21999: Removed

devdoc/rc_ptr.md
parth21999: removed

devdoc/channel.md
```diff
@@ -0,0 +1,267 @@
+# `channel` requirements
+
+## Overview
+
+`channel` is a module that allows the user to move data asynchronously from producer to consumer without having to block the thread.
+
+Consumers of data call `channel_pull` to register a `PULL_CALLBACK` function to be called when data becomes available.
+
+Producers of data call `channel_push` to notify the channel that data is available.
+
```
anporumb, line 10: i absolutely love the new vocabulary here.

devdoc/channel.md
```diff
@@ -0,0 +1,267 @@
...
+`channel_pull` and `channel_push` can be called from different threads in any order.
+
+## `channel` state
+
+The state of the `channel` module consists of two components:
+- EPOCH: This is a monotonically increasing integer that is incremented every time an operation is completed or cancelled.
```
anporumb, line 16: >operation [](http://example.com/codeflow?start=86&length=9)

Now, if this "operation" is somehow linked to the completion of an API... then it might be fine. If it is the complete cycle of going through a push AND a pull, then I doubt this is fine.

Is the state incrementing at every API call? I guess the rest of the document will have answers.

And is this "operations" the same as the THANDLE(ASYNC_OP)?

devdoc/channel.md
```diff
@@ -0,0 +1,267 @@
...
+  - `CHANNEL_PUSHED`: This is the state when `channel_push` has completed execution. A subsequent call to `channel_pull` will result in the callbacks being called synchronously.
+  - `CHANNEL_PULLING`: This is a temporary state when `channel_pull` is executing. It causes concurrent calls to `channel_push` to wait.
+  - `CHANNEL_PULLED`: This is the state when `channel_pull` has completed execution. A subsequent call to `channel_push` will result in the callback being called synchronously.
+
+Both components are stored in a single `volatile_atomic` variable that is manipulated using `interlocked` APIs.
+
+```mermaid
+stateDiagram-v2
+    B: EPOCH=x,STATE=CHANNEL_READY
+    N: EPOCH=x,STATE=CHANNEL_PUSHED
```
anporumb, line 29: >x [](http://example.com/codeflow?start=13&length=1)

seeing in this diagram as the state doesn't increase until it returns to "CHANNEL_READY" I am starting to think that the "operation" from line 16 is a complete cycle. 

oO

devdoc/channel.md
```diff
@@ -0,0 +1,267 @@
...
+### channel_dispose
+```c
+    void channel_dispose(CHANNEL* channel)
+```
+
+`channel_dispose` disposes the given `channel`.
+
+**SRS_CHANNEL_43_004: [** `channel_dispose` shall obtain the current state of the channel by calling `interlocked_add`. **]**
+
+**SRS_CHANNEL_43_057: [** If the current `state` is `CHANNEL_PULLING` or `CHANNEL_PUSHING`, `channel_dispose` shall wait for the state to change by calling `wait_on_address`. **]**
```
anporumb, line 115: >wait_on_address [](http://example.com/codeflow?start=157&length=15)

This API is always wrong to be called for any waits, and the reason is that `wait_on_address` is vulnerable in regards to spurious wakes.

I would use InterlockedHL... something. I think we have both, waitForValue and waitForNotValue.

I think this goes "here and everywhere".

anporumb, line 115: >CHANNEL_PULLING [](http://example.com/codeflow?start=53&length=15)

I don't think this can ever happen, can it? 

For `channel_dispose` to be called (seeing as this is a THANDLE) would mean that there are no other references to it. 

And if there are no references... how come the state is somewhere "in between", as in "somebody **is calling**" pull or push.

Wondering if a unit test can even be written for this.

devdoc/channel.md
```diff
@@ -0,0 +1,267 @@
...
+
+**SRS_CHANNEL_43_008: [** If `pull_callback` is `NULL`, `channel_pull` shall fail and return `CHANNEL_RESULT_INVALID_ARGS`. **]**
+
+**SRS_CHANNEL_43_009: [** If `out_op_pull` is `NULL`, `channel_pull` shall fail and return `CHANNEL_RESULT_INVALID_ARGS`. **]**
+
+**SRS_CHANNEL_43_010: [** `channel_pull` shall call `interlocked_add` to obtain the current `state` of the `channel`. **]**
+
+**SRS_CHANNEL_43_011: [** If the current `state` is `CHANNEL_PULLING` or `CHANNEL_PULLED`, `channel_pull` shall fail and return `CHANNEL_RESULT_REFUSED`. **]**
+
+**SRS_CHANNEL_43_059: [** If the current `state` is `CHANNEL_PUSHING`, `channel_pull` shall wait for the `state` to change by calling `wait_on_address`. **]**
```
anporumb, line 142: >wait [](http://example.com/codeflow?start=92&length=4)

I remember how in iteration 11?? where I had my last comments, one of them was "nothing waits here" (and then made the argument about the name of the module).

Fate loves irony.

devdoc/channel.md
```diff
@@ -0,0 +1,267 @@
...
+
+**SRS_CHANNEL_43_012: [** If the current `state` is `CHANNEL_PUSHED`, `channel_pull` shall : **]**
+
+- **SRS_CHANNEL_43_060: [** store `push_callback`, `push_context` and `data` given to `channel_push`. **]**
```
anporumb, line 146: >SRS_CHANNEL_43_060 [](http://example.com/codeflow?start=4&length=18)

suppose between 012_ and 060 the state should change to "pulling"?

anporumb, line 146: >`data` [](http://example.com/codeflow?start=70&length=6)

This data is moved with THANDLE_MOVE?

devdoc/channel.md
```diff
@@ -0,0 +1,267 @@
...
+
+- **SRS_CHANNEL_43_014: [** call the stored `push_callback` with the stored `push_context` and `result` as `CHANNEL_RESULT_OK`. **]**
+
+- **SRS_CHANNEL_43_054: [** set `out_op_pull` to `NULL` by calling `THANDLE_ASSIGN`. **]**
+
+- **SRS_CHANNEL_43_015: [** return `CHANNEL_RESULT_SYNC`. **]**
+
+**SRS_CHANNEL_43_016: [** If the current state is `CHANNEL_READY`, `channel_pull` shall: **]**
+
+**SRS_CHANNEL_43_061: [** set the state to `CHANNEL_PULLING` by calling `interlocked_compare_exchange`. **]**
```
anporumb, line 162: >interlocked_compare_exchange [](http://example.com/codeflow?start=73&length=28)

`compare` with what? And what happens if the `compare` is equal/not equal?

devdoc/channel.md
```diff
@@ -0,0 +1,267 @@
...
+
+- **SRS_CHANNEL_43_018: [** store the given `pull_callback` and the given `pull_context` in the `channel`. **]**
+
+- **SRS_CHANNEL_43_019: [** store `state` in the created `THANDLE(ASYNC_OP)`. **]**
+
+- **SRS_CHANNEL_43_020: [** store the given `channel` in the created `THANDLE(ASYNC_OP)` by calling `THANDLE_INITIALIZE`. **]**
+
+- **SRS_CHANNEL_43_021: [** store the created `THANDLE(ASYNC_OP)` in `out_op_pull` by calling `THANDLE_INITIALIZE_MOVE`. **]**
+
+- **SRS_CHANNEL_43_062: [** set the `state` of the given `channel` to `CHANNEL_PULLED` by calling `interlocked_exchange`. **]**
```
anporumb, line 174: >interlocked_exchange [](http://example.com/codeflow?start=99&length=20)

hmmmm.

devdoc/channel.md
```diff
@@ -0,0 +1,267 @@
...
+
+- **SRS_CHANNEL_43_063: [** signal waiting threads by calling `wake_by_address_all`. **]**
```
anporumb, line 176: >threads [](http://example.com/codeflow?start=43&length=7)

So there can be more than 1 thread waiting? So in the context of "pull" how does that happen?

What are these threads that are waiting doing and why?

devdoc/channel.md
```diff
@@ -0,0 +1,267 @@
...
+**SRS_CHANNEL_43_023: [** If there are any failures, `channel_pull` shall fail and return `CHANNEL_RESULT_ERROR`. **]**
+
+### cancel_pull
+```
+static void cancel_pull(void* context)
+```
+
+`cancel_pull` is the cancel function given to `async_op_create` when `channel_pull` is called.
+
+**SRS_CHANNEL_43_046: [** If the current `EPOCH` and `state` of the `channel` is the same as the initial `EPOCH` and `state` of the `channel` when `channel_pull` was called, `cancel_pull` shall: **]**
```
anporumb, line 189: >current `EPOCH` and `state` of the `channel` [](http://example.com/codeflow?start=33&length=44)

Should mention somewhere how these are obtained, seeing as in the other places of the document there's ample mentions of interlocked operations.

anporumb, line 189: >initial [](http://example.com/codeflow?start=97&length=7)

I think "recorded"/"stored" is the term here. (looking at SRS_43_019)

devdoc/channel.md
```diff
@@ -0,0 +1,267 @@
...
+
+- **SRS_CHANNEL_43_047: [** increment the `EPOCH` and set the `state` of the `channel` stored in the given `context` to `CHANNEL_READY` by calling `interlocked_compare_exchange`. **]**
```
anporumb, line 191: >interlocked_compare_exchange [](http://example.com/codeflow?start=148&length=28)

This is the last time I am making this comment, but interlocked_`compare`_exchange is essentially an "if" + "exchange"  instruction. Not sure what the "if" is here and how it is used.

Here and everywhere, I have a feeling this comment might apply to other places in this document.

devdoc/channel.md
anporumb, line 189: >`state` of the `channel` [](http://example.com/codeflow?start=53&length=24)

why does the state of the channel matter for this cancel?

I can almost understand `epoch`, but ... state?

devdoc/channel.md
```diff
@@ -0,0 +1,267 @@
...
+
+- **SRS_CHANNEL_43_048: [** call the `pull_callback` given to `channel_pull` with `pull_context` as the `pull_context` given to `channel_pull`, `data` as `NULL` and `result` as `CHANNEL_RESULT_CANCELLED`. **]**
+
+
+### channel_push
+```c
+    MOCKABLE_FUNCTION(, CHANNEL_RESULT, channel_push, THANDLE(CHANNEL), channel, THANDLE(RC_PTR), data, PUSH_CALLBACK, push_callback, void*, push_context, THANDLE(ASYNC_OP)*, out_op_push);
+```
+
+`channel_push` notifies the channel that there is data available and registers the given `push_callback` to be called when the given `data` has been consumed.
```
anporumb, line 201: this is SO CLEAR to me! I understood everything!

Cheers!

devdoc/channel.md
```diff
@@ -0,0 +1,267 @@
...
+
+**SRS_CHANNEL_43_029: [** If the current `state` is `CHANNEL_PULLED`, `channel_push` shall: **]**
+
+- **SRS_WAITER_43_001: [** store `pull_callback` and `pull_context` given to `channel_pull`. **]**
+
+- **SRS_CHANNEL_43_043: [** call `interlocked_compare_exchange` to increment the `EPOCH` and change the state to `CHANNEL_READY`. **]**
+
+- **SRS_CHANNEL_43_030: [** call the stored `pull_callback` with the stored `pull_context`, `data` as the given `data` and `result` as `CHANNEL_RESULT_OK`. **]**
+
+- **SRS_CHANNEL_43_031: [** call the given `push_callback` with `push_context` as the given `push_context` and `result` as `CHANNEL_RESULT_OK`. **]**
```
anporumb, line 223: >call the given `push_callback` [](http://example.com/codeflow?start=28&length=30)

I understand the re-entrancy here, and maybe that's a concern for later, but... the document should spell out the mitigations against stack overflows. A producer that pushes to an "always ready" consumer will stack overflow itself.

devdoc/channel.md
```diff
@@ -0,0 +1,267 @@
...
+
+- **SRS_CHANNEL_43_055: [** set `out_op_push` to `NULL` by calling `THANDLE_ASSIGN`. **]**
+
+- **SRS_CHANNEL_43_032: [** return `CHANNEL_RESULT_SYNC`. **]**
+
+**SRS_CHANNEL_43_033: [** If the current state is `CHANNEL_READY`, `channel_push` shall: **]**
+
+- **SRS_CHANNEL_43_065: [** set the state to `CHANNEL_PUSHING` by calling `interlocked_compare_exchange`. **]**
```
anporumb, line 231: >CHANNEL_PUSHING [](http://example.com/codeflow?start=46&length=15)

I think this state should be set in all cases, not only when READY. I might have had a similar comment for pull.

devdoc/channel.md
```diff
@@ -0,0 +1,267 @@
...
+
+- **SRS_CHANNEL_43_036: [** store the given `data` in the `channel` by calling `THANDLE_INITIALIZE`. **]**
+
+- **SRS_CHANNEL_43_037: [** store `state` in the created `THANDLE(ASYNC_OP)`. **]**
+
+- **SRS_CHANNEL_43_038: [** store the given `channel` in the created `THANDLE(ASYNC_OP)` by calling `THANDLE_INITIALIZE`. **]**
+
+- **SRS_CHANNEL_43_039: [** store the created `THANDLE(ASYNC_OP)` in `out_op_push`by calling `THANDLE_INITIALIZE_MOVE`. **]**
+
+set the `state` of the given `channel` to `CHANNEL_PUSHED` by calling `interlocked_exchange`.
```
anporumb, line 245: >CHANNEL_PUSHED [](http://example.com/codeflow?start=43&length=14)

missing ... SRS on this line? And the next one

devdoc/channel.md
```diff
@@ -0,0 +1,267 @@
...
+
+`cancel_push` is the cancel function given to `async_op_create` when `channel_push` is called.
+
+**SRS_CHANNEL_43_049: [** If the current `EPOCH` and `state` of the `channel` is the same as the initial `EPOCH` and `state` of the `channel` when `channel_push` was called, `cancel_push` shall: **]**
+
+- **SRS_CHANNEL_43_050: [** increment the `EPOCH` and set the `state` of the `channel` stored in the given `context` to `CHANNEL_READY` by calling `interlocked_compare_exchange`. **]**
+
+- **SRS_CHANNEL_43_051: [** call the `push_callback` given to `channel_push` with `push_context` as the `push_context` given to `channel_push` and `result` as `CHANNEL_RESULT_CANCELLED`. **]**
+
+- **SRS_CHANNEL_43_052: [** release the reference to `data` stored in `channel` by calling `THANDLE_ASSIGN`. **]**
```
anporumb, line 267: >SRS_CHANNEL_43_052 [](http://example.com/codeflow?start=4&length=18)

I will be waiting because the interlocked comments and the state management in general.

devdoc/channel.md
parth21999: "operation" is "complete cycle of going through a push AND a pull"

Only need to increment epoch at the completion of the cycle

devdoc/channel.md
```diff
@@ -0,0 +1,267 @@
...
+  - `CHANNEL_READY`: This is the default state. In this state, the module accepts calls to `channel_pull` and `channel_push`
+  - `CHANNEL_PUSHING`: This is a temporary state when `channel_push` is executing. It causes concurrent calls to `channel_pull` to wait.
+  - `CHANNEL_PUSHED`: This is the state when `channel_push` has completed execution. A subsequent call to `channel_pull` will result in the callbacks being called synchronously.
+  - `CHANNEL_PULLING`: This is a temporary state when `channel_pull` is executing. It causes concurrent calls to `channel_push` to wait.
+  - `CHANNEL_PULLED`: This is the state when `channel_pull` has completed execution. A subsequent call to `channel_push` will result in the callback being called synchronously.
+
+Both components are stored in a single `volatile_atomic` variable that is manipulated using `interlocked` APIs.
+
+```mermaid
+stateDiagram-v2
```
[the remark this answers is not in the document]
parth21999: Added them

devdoc/channel.md
parth21999: Changing to InterlockedHL_WaitForNotValue

parth21999: Hmm yeah this won't happen

devdoc/channel.md
parth21999: Even sm claims to not wait:

"""
there's no "parking", "queueing", "sleeping", "waiting", "buffering", "saving" or otherwise postponing or delaying the execution of an API: it either starts to execute (for barriers execution includes waiting for all the previous APIs to finish executing; for close it means draining everything - barriers or calls alike) or it fails.
"""

4 hits for `InterlockedHL_WaitForValue`
1 hit for `ThreadAPI_Sleep`

I think what "not waiting" means is that the module doesn't wait for the user, can wait on its own APIs.

devdoc/channel.md
parth21999: THANDLE_INITIALIZE_MOVE

parth21999: No. State changes are only as per the state diagram.

parth21999: Now that I think about it, we shouldn't modify the channel (_MOVE releases the reference that channel holds), because the current thread might not be the one that completes the operation.

So this should be THANDLE_INITIALIZE.

devdoc/channel.md
parth21999: State should not be changed by a different thread once the module is in PULLING/PUSHING state. Only the thread that put the module in PULLING/PUSHING should change the state.

devdoc/channel.md
parth21999: State is READY.
Thread 1 calls pull. State is PULLING.
Thread 2 calls push, Thread 3 calls push. Both threads wait.
Thread 1 changes state to PULLED, both threads should be signaled. Only one will succeed, the other will be REFUSED.

devdoc/channel.md
parth21999: Only epoch really needs to be compared. 
If epoch matches, state must match because once channel is in PULLED/PUSHED states, the only way to change the state is push/pull or cancel, both of which increment epoch.

Epoch must be compared because channel could cycle back to the PULLED state for a later operation, which means the current operation has already been completed or cancelled and nothing more needs to be done in the current call to cancel.

devdoc/channel.md
parth21999: I added the PULLING/PUSHING states to solve the problem of concurrent pull/push calls on READY state trying to change the data in the channel (the callbacks, context, THANDLE(RC_PTR) data)). PULLING/PUSHING are temporary states that cause a concurrent push/pull to wait until the current pull/push is done updating the data in the channel.

If the channel is already in PULLED/PUSHED, all pull/push calls will be REFUSED and only a single push/pull will proceed.

devdoc/channel.md
parth21999: 43_046 says : If ...., `cancel_pull` shall:

This is the if condition.

Indented requirements (47-48), are the requirements that execute if the condition is true. 

For the "else" part, do nothing is implicit.

parth21999: Indenting conditional requirements for better readability

devdoc/channel.md
parth21999: Added note in Reentrancy section

src/channel.c
```diff
@@ -0,0 +1,70 @@
...
+
+IMPLEMENT_MOCKABLE_FUNCTION(, CHANNEL_RESULT, channel_push, THANDLE(CHANNEL), channel, THANDLE(RC_PTR), data, PUSH_CALLBACK, push_callback, void*, push_context, THANDLE(ASYNC_OP)*, out_op_push)
+{
+    (void)channel;
+    (void)data;
+    (void)push_callback;
+    (void)push_context;
+    (void)out_op_push;
+    return CHANNEL_RESULT_ERROR;
+}
```
dcristoloveanu, line 70: Nitpick: we should have a new line at the end of file

inc/c_util/channel.h
```diff
@@ -0,0 +1,51 @@
...
+
+#include "c_pal/gballoc_hl.h"
+#include "c_pal/gballoc_hl_redirect.h"
+#include "c_pal/thandle.h"
+
+#include "c_util/async_op.h"
+#include "c_util/rc_ptr.h"
+
+#define CHANNEL_RESULT_VALUES \
+    CHANNEL_RESULT_SYNC, \
```
dcristoloveanu, line 17: >SYNC [](http://example.com/codeflow?start=19&length=4)

What does this mean? Does this mean COMPLETED_SYNC?

inc/c_util/channel.h
```diff
@@ -0,0 +1,51 @@
...
+    CHANNEL_RESULT_ASYNC, \
```
dcristoloveanu, line 18: >ASYNC [](http://example.com/codeflow?start=19&length=5)

Is this OK (as in WILL_COMPLETE_ASYNC)?

dcristoloveanu: /consider poetry, but not critical.

inc/c_util/channel.h
parth21999: Yes

devdoc/rc_ptr.md
```diff
@@ -0,0 +1,41 @@
...
+`rc_ptr` is a module that encapsulates reference-counted pointers.
+
+## Exposed API
+```c
+typedef void (*RC_PTR_FREE_FUNC)(void* context);
+
+typedef struct RC_PTR_TAG
+{
+    const void* ptr;
+    RC_PTR_FREE_FUNC free_func; // will be removed later
```
dcristoloveanu, line 14: >// will be removed later [](http://example.com/codeflow?start=32&length=24)

huh?

devdoc/channel.md
```diff
@@ -0,0 +1,274 @@
...
+
+**SRS_CHANNEL_43_010: [** `channel_pull` shall call `interlocked_add` to obtain the `state` of the `channel`. **]**
+
+**SRS_CHANNEL_43_011: [** If the obtained `state` is `CHANNEL_PULLING` or `CHANNEL_PULLED`, `channel_pull` shall fail and return `CHANNEL_RESULT_REFUSED`. **]**
+
+**SRS_CHANNEL_43_059: [** If the obtained `state` is `CHANNEL_PUSHING`, `channel_pull` shall wait for the `state` to change by calling `InterlockedHL_WaitForNotValue`. **]**
+
+**SRS_CHANNEL_43_012: [** If the obtained `state` is `CHANNEL_PUSHED`, `channel_pull` shall : **]**
+
+  - **SRS_CHANNEL_43_060: [** store `push_callback`, `push_context` and `data` given to `channel_push`. **]**
```
dcristoloveanu, line 149: >store  [](http://example.com/codeflow?start=30&length=6)

Not clean, we should say that we're taking a snapshot on the stack of these values

devdoc/channel.md
```diff
@@ -0,0 +1,274 @@
...
+
+  - **SRS_CHANNEL_43_053: [** release the reference to the stored `data` by calling `THANDLE_ASSIGN`. **]**
+
+  - **SRS_CHANNEL_43_014: [** call the stored `push_callback` with the stored `push_context` and `result` as `CHANNEL_RESULT_OK`. **]**
+
+  - **SRS_CHANNEL_43_054: [** set `out_op_pull` to `NULL` by calling `THANDLE_ASSIGN`. **]**
+
+  - **SRS_CHANNEL_43_015: [** return `CHANNEL_RESULT_SYNC`. **]**
+
+**SRS_CHANNEL_43_016: [** If the obtained `state` is `CHANNEL_READY`, `channel_pull` shall: **]**
```
dcristoloveanu, line 165: >If the obtained `state` is `CHANNEL_READ [](http://example.com/codeflow?start=26&length=40)

I could not help but thinking about how we would use an event loop for this rather than baking interlocked operations. But just food for thought.

dcristoloveanu (approved):
:shipit:

devdoc/rc_ptr.md
parth21999: Removed

devdoc/channel.md
```diff
@@ -0,0 +1,274 @@
...
+
+**SRS_CHANNEL_43_024: [** If `channel` is `NULL`, `channel_push` shall fail and return `CHANNEL_RESULT_INVALID_ARGS`. **]**
+
+**SRS_CHANNEL_43_025: [** If `push_callback` is `NULL`, `channel_push` shall fail and return `CHANNEL_RESULT_INVALID_ARGS`. **]**
+
+**SRS_CHANNEL_43_026: [** If `out_op_push` is `NULL`, `channel_push` shall fail and return `CHANNEL_RESULT_INVALID_ARGS`. **]**
+
+**SRS_CHANNEL_43_027: [** `channel_push` shall call `interlocked_add` to obtain the current `state` of the `channel`. **]**
+
+**SRS_CHANNEL_43_028: [** If the obtained `state` is `CHANNEL_PUSHING` or `CHANNEL_PUSHED`, `channel_push` shall fail and return `CHANNEL_RESULT_REFUSED`. **]**
```
anporumb, line 216: >If the obtained `state` is `CHANNEL_PUSHING` or `CHANNEL_PUSHED`, [](http://example.com/codeflow?start=26&length=65)

So I am going to have a comment here that nobody will like (self included), but.

On one hand we are doing all the fancy interlocked thingies and smart bitness and everything. However, the result is this:

If the producer has 100 items, it can still only hand over these items ONE by ONE, back to back. That is, the next item, even when available, cannot be passed to `push` until the previous item has been consumed. The problem with this being of 2 types:

1) performance - back to back is not very quick... sigh

2) ease of use - as a producer looks like I would have to manage the state of this component myself - and make sure that I do not call "push" unless the state is somehow detected as "ready to receive a push".

====

I can hear arguments that say: "but we're going to have 222 of these in a <<queue>>" and I still believe that everything above still applies.

Maybe we should reconsider this design. (that's how back we'd have to go).

./waiting

PR #291, opened 2023-07-10
`channel` header and specs

parth21999:


devdoc/rc_ptr.md
```diff
@@ -0,0 +1,41 @@
...
+
+`rc_ptr` is a module that encapsulates reference-counted pointers.
+
+## Exposed API
+```c
+typedef void (*RC_PTR_FREE_FUNC)(void* context);
+
+typedef struct RC_PTR_TAG
+{
+    const void* ptr;
```
anporumb, line 13: >const [](http://example.com/codeflow?start=4&length=5)

why is const needed?

inc/c_util/rc_ptr.h
```diff
@@ -0,0 +1,43 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#ifndef RC_PTR_H
+#define RC_PTR_H
+
+#ifdef __cplusplus
+#include <cstddef>
+#else
+#include <stddef.h>
```
anporumb, line 10: >stddef [](http://example.com/codeflow?start=10&length=6)

where is this header needed?

devdoc/rc_ptr.md
anporumb, line 13: >const [](http://example.com/codeflow?start=4&length=5)

header doesn't have "const" here.

devdoc/channel.md
```diff
@@ -0,0 +1,297 @@
...
+Consumers of data call `channel_pull` to register a `PULL_CALLBACK` function to be called when data becomes available.
+
+Producers of data call `channel_push` to notify the channel that data is available and provide a `PUSH_CALLBACK` to be called when the data has been consumed.
+
+`channel_pull` and `channel_push` can be called from different threads in any order. The callbacks are executed on threadpool threads.
+
+## `channel` internal state
+
+The internal state of the `channel` module can have the following values:
+  - `CHANNEL_EMPTY`: This is the initial state of the `channel`. It means that there is no data available to be pulled. Calling `channel_pull` or `channel_push` will cause the state to transition to `CHANNEL_PUSHING` or `CHANNEL_PULLING` respectively.
```
anporumb, line 16: does this internal state ever return back to "EMPTY"?

What happens after push+pull+callbacks executed? What's the state now?

devdoc/channel.md
```diff
@@ -0,0 +1,297 @@
...
+
+    R --> R: channel_pull
+    N --> N: channel_push
+
+    E --> [*]: channel_destroy
+```
+
+## Reentrancy
+
+`channel_pull` and `channel_push` can be called from callbacks of this module. Since the callbacks are executed on threadpool threads, there is no risk of stack overflow.
```
anporumb, line 47: >there is no risk of stack overflow. [](http://example.com/codeflow?start=135&length=35)

famous last words

devdoc/channel.md
```diff
@@ -0,0 +1,297 @@
...
+
+**SRS_CHANNEL_43_078: [** `channel_create` shall create a `CHANNEL_INTERNAL` object by calling `THANDLE_MALLOC`.**]**
+
+**SRS_CHANNEL_43_079: [** `channel_create` shall store the created `THANDLE(CHANNEL_INTERNAL)` in the `THANDLE(CHANNEL)`. **]**
+
+**SRS_CHANNEL_43_080: [** `channel_create` shall initialize the `CHANNEL_INTERNAL` with the given `threadpool`. **]**
+
+**SRS_CHANNEL_43_098: [** `channel_create` shall call `srw_lock_create`. **]**
+
+**SRS_CHANNEL_43_081: [** `channel_create` shall call `sm_create`. **]**
```
anporumb, line 109: every time I see "srw_lock" and "sm" my heart skips a beat

devdoc/channel.md
```diff
@@ -0,0 +1,297 @@
...
+
+`channel_pull` registers the given `pull_callback` to be called when there is data to be consumed.
+
+**SRS_CHANNEL_43_007: [** If `channel` is `NULL`, `channel_pull` shall fail and return `CHANNEL_RESULT_INVALID_ARGS`. **]**
+
+**SRS_CHANNEL_43_008: [** If `pull_callback` is `NULL`, `channel_pull` shall fail and return `CHANNEL_RESULT_INVALID_ARGS`. **]**
+
+**SRS_CHANNEL_43_009: [** If `out_op_pull` is `NULL`, `channel_pull` shall fail and return `CHANNEL_RESULT_INVALID_ARGS`. **]**
+
+**SRS_CHANNEL_43_010: [** `channel_pull` shall call `srw_lock_acquire_exclusive`. **]**
```
anporumb, line 177: so if the first thing that happens here is "acquire_exclusive" I don't understand why SM is then needed.

Consider removing SM and maybe using a simple counter variable?

devdoc/channel.md
```diff
@@ -0,0 +1,297 @@
...
+
+ - **SRS_CHANNEL_43_107: [** set `*out_op_pull` to the created `THANDLE(ASYNC_OP)`. **]**
+
+**SRS_CHANNEL_43_108: [** If the `state` of the `channel` is `CHANNEL_PUSHED`: **]**
+
+ - **SRS_CHANNEL_43_109: [** `channel_pull` shall call `DList_RemoveHeadList` on the list of pending operations to obtain the `operation`. **]**
+
+ - **SRS_CHANNEL_43_110: [** If the list of pending operations is now empty, `channel_pull` shall set the `state` to `CHANNEL_EMPTY`. **]**
+
+ - **SRS_CHANNEL_43_111: [** `channel_pull` shall set the `state` of the obtained `operation` to `CHANNEL_OP_SCHEDULED`. **]**
```
anporumb, line 199: >CHANNEL_OP_SCHEDULED [](http://example.com/codeflow?start=98&length=20)

this must be some internal enum. Doesn't seem documented.

devdoc/channel.md
```diff
@@ -0,0 +1,297 @@
...
+
+
+### cancel_channel_op
+```c
+    static void cancel_channel_op(void* channel_op_context);
+```
+
+**SRS_CHANNEL_43_133: [** `cancel_channel_op` is the cancel callback that is passed to `async_op_create` when creating a `THANDLE(ASYNC_OP)` for a `channel_push` or `channel_pull` operation. **]**
+
+**SRS_CHANNEL_43_134: [** `cancel_channel_op` shall call `srw_lock_acquire_exclusive`. **]**
```
anporumb, line 271: >srw_lock_acquire_exclusive [](http://example.com/codeflow?start=58&length=26)

up until now I would have believed that the SRW applies to the "channel" (seeing as it is taken at push/pull APIs).

Is there any way in which we can have this `cancel` that applies to a single operation not acquire the whole channel?

devdoc/channel.md
```diff
@@ -0,0 +1,297 @@
...
+
+ - **SRS_CHANNEL_43_110: [** If the list of pending operations is now empty, `channel_pull` shall set the `state` to `CHANNEL_EMPTY`. **]**
+
+ - **SRS_CHANNEL_43_111: [** `channel_pull` shall set the `state` of the obtained `operation` to `CHANNEL_OP_SCHEDULED`. **]**
+
+ - **SRS_CHANNEL_43_112: [** `channel_pull` shall store the `pull_callback` and `pull_context` in the obtained `operation`. **]**
+
+ - **SRS_CHANNEL_43_113: [** `channel_pull` shall call `threadpool_schedule_work` with `execute_callbacks` as `work_function` and the obtained `operation` as `work_function_context`. **]**
+
+ - **SRS_CHANNEL_43_114: [** `channel_pull` shall set `*out_op_pull` to the `THANDLE(ASYNC_OP)` of the obtained `operation`. **]**
```
anporumb, line 205: >`THANDLE(ASYNC_OP) [](http://example.com/codeflow?start=76&length=18)

so when push and pull are "paired" because one of them already exists, they'll share the THANDLE(ASYNC_OP)? Meaning the user will see the same ASYNC_OP on both the push and the pull callback?

devdoc/channel.md
```diff
@@ -0,0 +1,297 @@
...
+
+**SRS_CHANNEL_43_134: [** `cancel_channel_op` shall call `srw_lock_acquire_exclusive`. **]**
+
+**SRS_CHANNEL_43_135: [** If the `state` of `operation` is `CHANNEL_OP_CREATED`: **]**
+
+ - **SRS_CHANNEL_43_136: [** `cancel_channel_op` shall set the `state` to `CHANNEL_OP_CANCELLED` **]**
+
+ - **SRS_CHANNEL_43_137: [** `cancel_channel_op` shall call `DList_RemoveEntryList` to remove the `operation` from the list of pending operations. **]**
+
+ - **SRS_CHANNEL_43_138: [** `cancel_channel_op` shall call `threadpool_schedule_work` with `execute_callbacks` as `work_function` and the `operation` as `work_function_context`. **]**
```
anporumb, line 279: >`operation` as `work_function_context` [](http://example.com/codeflow?start=139&length=38)

whenever I see a `void*` pointer getting assigned some value from a THANDLE my heart skips a beat.

what happens if work has been scheduled... and work did not execute YET, but meanwhile the user decides to THANDLE_ASSIGN(ASYNC_OP) = NULL for both the pull and the push ASYNC_OPs?

Is there anybody still holding on to the ASYNC_OP? Or when the work is executed, it will find some freed memory?

devdoc/channel.md
```diff
@@ -0,0 +1,297 @@
...
+
+**SRS_CHANNEL_43_041: [** If there are any failures, `channel_push` shall fail and return `CHANNEL_RESULT_ERROR`. **]**
+
+
+### cancel_channel_op
+```c
+    static void cancel_channel_op(void* channel_op_context);
+```
+
+**SRS_CHANNEL_43_133: [** `cancel_channel_op` is the cancel callback that is passed to `async_op_create` when creating a `THANDLE(ASYNC_OP)` for a `channel_push` or `channel_pull` operation. **]**
```
anporumb, line 269: >SRS_CHANNEL_43_133 [](http://example.com/codeflow?start=2&length=18)

words following this requirement seem to be more of a description rather than actions.

I think this doesn't need the SRS...

devdoc/channel.md
anporumb: Seems to happen in other requirements too (such as SRS_CHANNEL_43_113)

devdoc/channel.md
```diff
@@ -0,0 +1,297 @@
...
+
+**SRS_CHANNEL_43_081: [** `channel_create` shall call `sm_create`. **]**
+
+**SRS_CHANNEL_43_082: [** `channel_create` shall call `sm_open_begin`. **]**
+
+**SRS_CHANNEL_43_083: [** `channel_create` shall call `sm_open_end`. **]**
+
+**SRS_CHANNEL_43_084: [** `channel_create` shall call `DList_InitializeListHead`. **]**
+
+**SRS_CHANNEL_43_085: [** `channel_create` shall call `interlocked_exchange` to set the `state` to `CHANNEL_EMPTY`. **]**
```
anporumb, line 117: >interlocked_exchange [](http://example.com/codeflow?start=55&length=20)

so there's just 1 mention of "interlocked" in this whole document (here).

Either:
1) there's other interlocked that are missing
2) this is un-needed.

THere cannot be a single interlocked mention in a document.

devdoc/channel.md
anporumb, line 47: >channel_pull` and `channel_push` can be called from callbacks of this module [](http://example.com/codeflow?start=1&length=76)

Can the callbacks assign NULL to the last reference of the THANDLE(CHANNEL) that is returned from create?

devdoc/channel.md
```diff
@@ -0,0 +1,297 @@
...
+
+**SRS_CHANNEL_43_077: [** If `threadpool` is `NULL`, `channel_create` shall fail and return `NULL`. **]**
+
+**SRS_CHANNEL_43_001: [** `channel_create` shall create a `CHANNEL` object by calling `THANDLE_MALLOC` with `channel_dispose` as `dispose`. **]**
+
+**SRS_CHANNEL_43_078: [** `channel_create` shall create a `CHANNEL_INTERNAL` object by calling `THANDLE_MALLOC`.**]**
+
+**SRS_CHANNEL_43_079: [** `channel_create` shall store the created `THANDLE(CHANNEL_INTERNAL)` in the `THANDLE(CHANNEL)`. **]**
+
+**SRS_CHANNEL_43_080: [** `channel_create` shall initialize the `CHANNEL_INTERNAL` with the given `threadpool`. **]**
```
anporumb, line 105: >CHANNEL_INTERNAL [](http://example.com/codeflow?start=65&length=16)

I've looked everywhere and this CHANNEL_INTERNAL seems to be holding `threadpool`. But threadpool is itself a THANDLE.

It is really weird to have a THANDLE on top of a THANDLE.

devdoc/channel.md
```diff
@@ -0,0 +1,297 @@
...
+
+**SRS_CHANNEL_43_133: [** `cancel_channel_op` is the cancel callback that is passed to `async_op_create` when creating a `THANDLE(ASYNC_OP)` for a `channel_push` or `channel_pull` operation. **]**
+
+**SRS_CHANNEL_43_134: [** `cancel_channel_op` shall call `srw_lock_acquire_exclusive`. **]**
+
+**SRS_CHANNEL_43_135: [** If the `state` of `operation` is `CHANNEL_OP_CREATED`: **]**
+
+ - **SRS_CHANNEL_43_136: [** `cancel_channel_op` shall set the `state` to `CHANNEL_OP_CANCELLED` **]**
+
+ - **SRS_CHANNEL_43_137: [** `cancel_channel_op` shall call `DList_RemoveEntryList` to remove the `operation` from the list of pending operations. **]**
```
anporumb, line 277: >DList_RemoveEntryList [](http://example.com/codeflow?start=61&length=21)

There's something which is not right here.

Here's a sequence of events.
1) push (so state is PUSHED)
2) pull 
Note: inside `pull` the op is given to the execution queue work. `threadpool_schedule_work`

nothing happens for a long time, because "insert reasons" but the user has the THANDLE(ASYNC_OP) for both push and pull.

The user gets bored so the user calls "cancel".

At this moment he is unable to cancel because the state has been switched to "SCHEDULED".

Consider having the possibility to cancel SCHEDULED operations.

devdoc/channel.md
```diff
@@ -0,0 +1,297 @@
...
+
+ - **SRS_CHANNEL_43_126: [** If the list of pending operations is now empty, `channel_push` shall set the `state` to `CHANNEL_EMPTY`. **]**
+
+ - **SRS_CHANNEL_43_127: [** `channel_push` shall set the `state` of the obtained `operation` to `CHANNEL_OP_SCHEDULED`. **]**
+
+ - **SRS_CHANNEL_43_128: [** `channel_push` shall store the `push_callback`, `push_context` and `data` in the obtained `operation`. **]**
+
+ - **SRS_CHANNEL_43_129: [** `channel_push` shall call `threadpool_schedule_work` with `execute_callbacks` as `work_function` and the obtained `operation` as `work_function_context`. **]**
+
+ - **SRS_CHANNEL_43_130: [** `channel_push` shall set `*out_op_push` to the `THANDLE(ASYNC_OP)` of the obtained `operation`. **]**
```
anporumb, line 255: >`*out_op_push` to the `THANDLE(ASYNC_OP)` of the obtained `operation`. [](http://example.com/codeflow?start=54&length=70)

I like this idea, buuuuut.... what happens if the item is now scheduled AND
1) pusher calls cancel. The pushed should expect that HIS callback will be called with "cancel". What should be the value that is received by the puller in the puller callback?
2) reverse for the puller... if the puller calls "cancel", then what's the callback value received by the pusher in the pusher's callback? Is it ABANDONED?

devdoc/channel.md
```diff
@@ -0,0 +1,297 @@
...
+
+**SRS_CHANNEL_43_131: [** `channel_push` shall call `srw_lock_release_exclusive`. **]**
+
+**SRS_CHANNEL_43_132: [** `channel_push` shall succeeds and return `CHANNEL_RESULT_OK`. **]**
```
anporumb, line 259: >s [](http://example.com/codeflow?start=54&length=1)

extra "s"

devdoc/channel.md
```diff
@@ -0,0 +1,297 @@
...
+    N: STATE=CHANNEL_PUSHED
+    R: STATE=CHANNEL_PULLED
+    E: STATE=CHANNEL_EMPTY
+
+    [*] --> B: channel_create
+
+    B --> N: channel_push
+    B --> R: channel_pull
+
+    N --> E:  channel_pull
```
anporumb, line 32: >  [](http://example.com/codeflow?start=13&length=1)

extra space. here and in line 36

devdoc/channel.md
```diff
@@ -0,0 +1,297 @@
...
+
+**SRS_CHANNEL_43_092: [** `channel_dispose` shall release the reference to `THANDLE(CHANNEL_INTERNAL)`. **]**
+
+
+### channel_close
+```c
+    static void channel_close(void* context);
+```
+
+`channel_close` is the callback passed to `sm_close_begin_with_cb` by `channel_dispose`. It abandons all pending operations
```
anporumb, line 151: >It abandons all pending operations [](http://example.com/codeflow?start=89&length=34)

How does it abandon operations for which both the pull and the push have been called, and the OP was given to threadpool_execute_work and the "execute work" didn't happen yet?

anporumb: Seeing as "43_096" says it will iterate over something (it doesn't have it, was given to threadpool_schedule_work).

devdoc/channel.md
```diff
@@ -0,0 +1,297 @@
...
+
+ - **SRS_CHANNEL_43_109: [** `channel_pull` shall call `DList_RemoveHeadList` on the list of pending operations to obtain the `operation`. **]**
+
+ - **SRS_CHANNEL_43_110: [** If the list of pending operations is now empty, `channel_pull` shall set the `state` to `CHANNEL_EMPTY`. **]**
+
+ - **SRS_CHANNEL_43_111: [** `channel_pull` shall set the `state` of the obtained `operation` to `CHANNEL_OP_SCHEDULED`. **]**
+
+ - **SRS_CHANNEL_43_112: [** `channel_pull` shall store the `pull_callback` and `pull_context` in the obtained `operation`. **]**
+
+ - **SRS_CHANNEL_43_113: [** `channel_pull` shall call `threadpool_schedule_work` with `execute_callbacks` as `work_function` and the obtained `operation` as `work_function_context`. **]**
```
anporumb, line 203: >threadpool_schedule_work [](http://example.com/codeflow?start=56&length=24)

can this lead to some... inversions here?

Assume the following:
1) push1
2) pull1 => schedule work1

The threadpool doesn't want to execute it yet.

3) push2
4) pull2 => schedule work2

The threadpool is free to execute work1 and work2 in whichever order it likes, including indicating first "work2" (so pull2 and push2 get completed) and then work1 (so pull1 and push1 callbacks get called).

Is using a threadpool even the solution here?

anporumb: Only noticing that everything is VERY ordered about the CHANNEL (not a bad thing) due to the exclusive lock taking at the begin of every API.

Then callbacks can be called into user space in a very random way, just because threadpool wants so.

devdoc/channel.md
```diff
@@ -0,0 +1,297 @@
...
+
+**SRS_CHANNEL_43_025: [** If `push_callback` is `NULL`, `channel_push` shall fail and return `CHANNEL_RESULT_INVALID_ARGS`. **]**
+
+**SRS_CHANNEL_43_026: [** If `out_op_push` is `NULL`, `channel_push` shall fail and return `CHANNEL_RESULT_INVALID_ARGS`. **]**
+
+**SRS_CHANNEL_43_116: [** `channel_push` shall call `srw_lock_acquire_exclusive`. **]**
+
+**SRS_CHANNEL_43_117: [** If the `state` of the `channel` is `CHANNEL_EMPTY` or `CHANNEL_PUSHED`: **]**
+
+ - **SRS_CHANNEL_43_118: [** `channel_push` shall call `sm_exec_begin`. **]**
```
anporumb, line 231: >sm_exec_begin [](http://example.com/codeflow?start=56&length=13)

the sm_exec_end that completes this sm_exec_begin is in line 297.

Meanwhile... a lot of things need to happen:
1) this function needs to exit (so it releases SRW_LOCK)
2) pull needs to be called (so it acquires SRW_LOCK) and pull must have advanced to line 

At this moment, we're inside `pull` function and pull advanced to line 203.

Note: at this moment we have sm_exec_begin and SRW_LOCK still taken.

Depending on the wits of the day of the threadpool, the SRW_LOCK might get released before or after the sm_exec_end is called.

Is that intended that sm's operation and the SRW_LOCK release can be completed in any order? Would that ever lead to a deadlock?

inc/c_util/rc_ptr.h
```diff
@@ -0,0 +1,43 @@
...
+    RC_PTR_FREE_FUNC free_func;
+} RC_PTR;
+
+THANDLE_TYPE_DECLARE(RC_PTR);
+
+#define PRI_RC_PTR "p"
+
+#define RC_PTR_VALUE(rc) ((rc)->ptr)
+
+#define RC_PTR_VALUE_OR_NULL(rc) (((rc) == NULL) ? NULL : (rc)->ptr)
```
anporumb, line 29: >RC_PTR_VALUE_OR_NULL [](http://example.com/codeflow?start=8&length=20)

:) seems .md is outdated

devdoc/channel.md
```diff
@@ -0,0 +1,297 @@
...
+
+`execute_callbacks` is the work function that is passed to `threadpool_schedule_work` when scheduling the execution of the callbacks for a `channel_push` or `channel_pull` operation.
+
+**SRS_CHANNEL_43_140: [** If the `state` of the `operation` is `CHANNEL_OP_SCHEDULED`, `execute_callbacks` shall call the `pull_callback` and the `push_callback` with `CHANNEL_CALLBACK_RESULT_OK`. **]**
+
+**SRS_CHANNEL_43_141: [** If the `state` of the `operation` is `CHANNEL_OP_CANCELLED`, `execute_callbacks` shall call whichever callback has been registered with `CHANNEL_CALLBACK_RESULT_CANCELLED`. **]**
+
+**SRS_CHANNEL_43_142: [** If the `state` of the `operation` is `CHANNEL_OP_ABANDONED`, `execute_callbacks` shall call whichever callback has been registered with `CHANNEL_CALLBACK_RESULT_ABANDONED`. **]**
+
+**SRS_CHANNEL_43_143: [** `execute_callbacks` shall call `sm_exec_end`. **]**
```
anporumb, line 297: >**]** [](http://example.com/codeflow?start=72&length=5)

all in all I need to revisit this review at the next iteration. Please ping me when the next iteration is available.

anporumb (changes requested):
:clock1:

devdoc/channel.md
parth21999: Yes it returns back to EMPTY. Added explanation.

devdoc/channel.md
parth21999: That would deadlock

devdoc/channel.md
parth21999: Interlocked is not needed. Spec was out of date.

devdoc/channel.md
parth21999: The section on Internal Reference Counting explains this.

devdoc/channel.md
parth21999: We need both:

SM : To make sure that THANDLE_ASSIGN(&channel, NULL)/dispose/close waits for all the callbacks to execute before returning.

SRW LOCK: To make sure only one thread is changing the state and the list at any time.

devdoc/channel.md
parth21999: Those operations will not be abandoned. Only the ones still sitting in the queue will be abandoned. Once an OP has been given to threadpool_schedule_work, it cannot be abandoned or cancelled.

devdoc/channel.md
parth21999: Yes that can happen. `channel` does not guarantee the order of the callbacks. The guarantee is that the data from the `n`th call to `channel_push` will be matched with the `n`th call to `channel_pull`.

devdoc/channel.md
parth21999: Yes.

devdoc/channel.md
parth21999: Tagged it by mistake.

devdoc/channel.md
parth21999: The lock is needed because `cancel_channel_op` must remove manipulate the list of pending operations.

devdoc/channel.md
parth21999: I considered it. Allowing this increases complexity without giving much benefit. Cancelling a SCHEDULED operation does not make the callbacks execute any sooner.

devdoc/channel.md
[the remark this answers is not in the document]
parth21999: After the callbacks are called

devdoc/channel.md
parth21999: I think that should be fine. Releasing the lock does not wait for sm_exec_end so the lock will always release.

devdoc/channel.md
parth21999: Added section explaining this.

devdoc/rc_ptr.md
parth21999: Fixed.

inc/c_util/channel.h
```diff
@@ -0,0 +1,50 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#ifndef CHANNEL_H
+#define CHANNEL_H
+
+#include "macro_utils/macro_utils.h"
+
+#include "c_pal/gballoc_hl.h"
+#include "c_pal/gballoc_hl_redirect.h"
```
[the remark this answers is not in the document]
parth21999: Removed.

inc/c_util/rc_ptr.h
parth21999: I thought we include this if we use NULL

devdoc/channel.md
parth21999: Removed SM

devdoc/channel.md
parth21999: Removed SM

devdoc/rc_ptr.md
```diff
@@ -0,0 +1,41 @@
...
+
+THANDLE_TYPE_DECLARE(RC_PTR);
+
+#define PRI_RC_PTR "p"
+
+#define RC_PTR_VALUE(rc) ((rc)->ptr)
+
+#define RC_PTR_AS_VOIDPTR(rc) (((rc) == NULL) ? NULL : (rc)->ptr)
+
+MOCKABLE_FUNCTION(, THANDLE(RC_PTR), rc_ptr_create_with_move_pointer, void*, ptr, RC_PTR_FREE_FUNC, free_func);
```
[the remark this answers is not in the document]
parth21999: Because we don't know the type of the data. 

Doing this would require templatizing the entire module on the type of data: `THANDLE(CHANNEL(DATA))` 

Can be done, but I haven't written a lot of macro code and don't want to spend time on it right now. Priorities.

devdoc/channel.md
parth21999: Allowing cancel right up to when the callbacks are called

devdoc/channel.md
```diff
@@ -0,0 +1,294 @@
...
+### cancel_channel_op
+```c
+    static void cancel_channel_op(void* channel_op_context);
+```
+
+`cancel_channel_op` is the cancel callback that is passed to `async_op_create` when creating a `THANDLE(ASYNC_OP)` for a `channel_push` or `channel_pull` operation.
+
+**SRS_CHANNEL_43_134: [** `cancel_channel_op` shall call `srw_lock_acquire_exclusive`. **]**
+
+**SRS_CHANNEL_43_136: [** `cancel_channel_op` shall call `interlocked_exchange` to obtain and set the `state` of the `operation` to `CHANNEL_OP_CANCELLED`. **]**
```
anporumb, line 270: >CHANNEL_OP_CANCELLED [](http://example.com/codeflow?start=133&length=20)

what happens if the OP had its state changed to "abandoned" already? still change it?

I think abandoned trumps cancelled.

devdoc/channel.md
```diff
@@ -0,0 +1,294 @@
...
+
+ - `CHANNEL_OP_CREATED`: This is the initial state of the `operation` after only one of `channel_pull`/`channel_push` have been called. It means that the `operation` has been created and added to the `list of pending operations`. A subsequent call to `channel_push`/`channel_pull` will cause the `operation` to transition to `CHANNEL_OP_SCHEDULED`.
+
+ - `CHANNEL_OP_SCHEDULED`: This is the state after the a pair of `channel_pull`/`channel_push` have been called. It means that the `operation` has been removed from the `list of pending operations` and is scheduled to have its callbacks called on the threadpool with result `CHANNEL_CALLBACK_RESULT_OK`. 
+
+ - `CHANNEL_OP_CANCELLED`: This is the state after the `operation` has been cancelled. It means that the `operation` has been removed from the `list of pending operations` and is scheduled to have a callback (either `PULL_CALLBACK` XOR `PUSH_CALLBACK`, depending on which of `channel_pull`/`channel_push` created it) called with result `CHANNEL_CALLBACK_RESULT_CANCELLED`.
```
anporumb, line 53: >XOR [](http://example.com/codeflow?start=232&length=3)

Why XOR? I believe we've discussed that operations can be cancelled even after both the callbacks exist and the operation has been transferred to the threadpool.

Reason to wait, because either me doesn't understand what XOR means or this is not what we have discussed yesterday.

I believe the "to have **a** callback" means only one which sort of reinforces that I understand what XOR means in this context AND this is not what I expect the specs to say.

devdoc/channel.md
```diff
@@ -0,0 +1,294 @@
...
+
+`channel_pull` and `channel_push` can be called from different threads in any order. The callbacks are executed on threadpool threads.
+
+## `channel` internal state
+
+The internal state of the `channel` module can have the following values:
+  - `CHANNEL_EMPTY`: This is the initial state of the `channel`. It means that there is no data available to be pulled. Calling `channel_pull` or `channel_push` will cause the state to transition to `CHANNEL_PUSHED` or `CHANNEL_PULLED` respectively. After all the `channel_pull` and `channel_push` calls have been paired, the state will transition back to `CHANNEL_EMPTY`.
+  - `CHANNEL_PUSHED`: This is the state after `channel_push` has been called. A subsequent call to `channel_pull` will schedule the callbacks on the threadpool.
+  - `CHANNEL_PULLED`: This is the state after `channel_pull` has been called. A subsequent call to `channel_push` will the callbacks on the threadpool.
+
```
anporumb, line 19: Consider removing these states.

PULLED is equivalent with "data has been produced and is not consumed"

PUSHED is equivalent to "consumer of data exists but data is not produced yet".

Both of these imply existance of a previous call to _pull or to _push which implies "tail" of the queue not NULL.

tail!=NULL => if it has _pull callback=> PULLED state, otherwise it is PUSHED state.

devdoc/channel.md
```diff
@@ -0,0 +1,294 @@
...
+    R --> R: channel_pull
+    N --> N: channel_push
+
+    E --> [*]: channel_destroy
+```
+
+## Operations
+
+The first call to `channel_pull`/`channel_push` either creates an `operation` and adds it to a `list of pending operations`. When `channel_push`/`channel_pull` is called subsequently, the least-recently added `operation` in the `list of pending operations` is scheduled to have its callbacks called on the threadpool. Each operation maintains an internal `state`. The `state` of each `operation` can have the following values:
```
anporumb, line 47: >either [](http://example.com/codeflow?start=48&length=6)

404 "or" not found.

parth21999: removed either

devdoc/channel.md
parth21999: Removing XOR

devdoc/channel.md
parth21999: Removed

devdoc/channel.md
```diff
@@ -0,0 +1,294 @@
...
+
+**SRS_CHANNEL_43_010: [** `channel_pull` shall call `srw_lock_acquire_exclusive`. **]**
+
+**SRS_CHANNEL_43_101: [** If the `state` of the `channel` is `CHANNEL_EMPTY` or `CHANNEL_PULLED`: **]**
+
+ - **SRS_CHANNEL_43_103: [** `channel_pull` shall create a `THANDLE(ASYNC_OP)` by calling `async_op_create` with `cancel_channel_op` as `cancel`. **]**
+
+ - **SRS_CHANNEL_43_104: [** `channel_pull` shall store the `pull_callback` and `pull_context` in the `THANDLE(ASYNC_OP)`. **]**
+
+ - **SRS_CHANNEL_43_144: [** `channel_pull`  shall call `interlocked_exchange` to set the `state` of the created `operation` to `CHANNEL_OP_CREATED`. **]**
```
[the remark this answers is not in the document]
parth21999: Removed

devdoc/channel.md
```diff
@@ -0,0 +1,294 @@
...
+
+
+### execute_callbacks
+```c
+    static void execute_callbacks(void* channel_op_context);
+```
+
+`execute_callbacks` is the work function that is passed to `threadpool_schedule_work` when scheduling the execution of the callbacks for an operation.
+
+**SRS_CHANNEL_43_145: [** `execute_callbacks` shall call `interlocked_exchange` to obtain the `state` of the `operation`.  **]**
```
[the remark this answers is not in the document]
parth21999: Removed interlocked

devdoc/channel.md
```diff
@@ -0,0 +1,278 @@
...
+
+The first call to `channel_pull`/`channel_push` creates an `operation` and adds it to a `list of pending operations`. When `channel_push`/`channel_pull` is called subsequently, the least-recently added `operation` in the `list of pending operations` is scheduled to have its callbacks called on the threadpool. Each operation stores the `CHANNEL_CALLBACK_RESULT` that the callback should be called with.
+
+`CHANNEL_CALLBACK_RESULT_ABANDONED`: The callbacks are called with this result when the when the module is disposing.
+
+`CHANNEL_CALLBACK_RESULT_CANCELLED`: The callbacks are called with this result when the `operation` has been cancelled. 
+
+`CHANNEL_CALLBACK_RESULT_OK`: The callbacks are called with this result in the success case.
+
+## `CHANNEL_CALLBACK_RESULT_ABANDONED` trumps `CHANNEL_CALLBACK_RESULT_CANCELLED` trumps `CHANNEL_CALLBACK_RESULT_OK`
```
anporumb, line 48: >## [](http://example.com/codeflow?start=0&length=2)

octothorpes not needed.

devdoc/channel.md
```diff
@@ -0,0 +1,278 @@
...
+### cancel_channel_op
+```c
+    static void cancel_channel_op(void* channel_op_context);
+```
+
+`cancel_channel_op` is the cancel callback that is passed to `async_op_create` when creating a `THANDLE(ASYNC_OP)` for a `channel_push` or `channel_pull` operation.
+
+**SRS_CHANNEL_43_134: [** `cancel_channel_op` shall call `srw_lock_acquire_exclusive`. **]**
+
+**SRS_CHANNEL_43_135: [** If the`operation` is in the list of pending `operations`: **]**
```
anporumb, line 258: >is in the list of pending [](http://example.com/codeflow?start=44&length=25)

This is an interesting determination. Does it piggyback on something or how is this determination made?

anporumb (approved):
:shipit:

devdoc/channel.md
```diff
@@ -0,0 +1,278 @@
...
+    static void channel_dispose(CHANNEL* channel);
+```
+
+`channel_dispose` disposes the given `channel` and schedules all pending operations to be abandoned.
+
+**SRS_CHANNEL_43_094: [** `channel_dispose` shall call `srw_lock_acquire_exclusive`. **]**
+
+**SRS_CHANNEL_43_095: [** `channel_dispose` shall iterate over the list of pending operations and do the following: **]**
+
+ - **SRS_CHANNEL_43_096: [** set the `result` of the `operation` to `CHANNEL_CALLBACK_RESULT_ABANDONED`. **]**
```
anporumb, line 141: >CHANNEL_CALLBACK_RESULT_ABANDONED [](http://example.com/codeflow?start=69&length=33)

Will the operations that were given to the threadpool equally set to `ABANDONED`?

parth21999: If they are in the threadpool, channel module does not hold a reference to them anymore, so they cannot be set to abandoned.

PR #292, opened 2023-07-20
Add linux compiler checks

jebrando:


inc/c_util/tarray_ll.h
```diff
@@ -18,7 +18,15 @@
 
 /*TARRAY is backed by a THANDLE build on the structure below*/
 #define TARRAY_STRUCT_TYPE_NAME_TAG(T) MU_C2(TARRAY_TYPEDEF_NAME(T), _TAG)
+// For gcc build that causes a false negative
+#ifdef __GNUC__
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wmaybe-uninitialized"
+#endif
```
dcristoloveanu, line 25: Hmmmmm, can we not ignore this?

tests/flags_to_string_ut/flags_to_string_ut.c
```diff
@@ -107,6 +107,12 @@ TEST_FUNCTION_CLEANUP(method_cleanup)
     umock_c_negative_tests_deinit();
 }
 
+// For gcc build that causes a false negative
+#ifdef __GNUC__
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wstringop-overflow"
+#endif
```
dcristoloveanu, line 114: What is the error for this one?
Not sure we should suppress ....

dcristoloveanu (changes requested):
:clock1:

dcristoloveanu (approved):
:shipit:

tests/flags_to_string_ut/flags_to_string_ut.c
jebrando: I believe this one will need to be suppressed.  I'll add information I get from the interwebs for the reasons.

PR #294, opened 2023-07-27
`channel` code and tests

parth21999:


src/channel.c
```diff
@@ -0,0 +1,131 @@
...
+        LogError("Invalid arguments: THANDLE(THREADPOOL) threadpool=%p", threadpool);
+    }
+    else
+    {
+        /*Codes_SRS_CHANNEL_43_078: [ channel_create shall create a CHANNEL_INTERNAL object by calling THANDLE_MALLOC with channel_internal_dispose as dispose.]*/
+        THANDLE(CHANNEL_INTERNAL) channel_internal = channel_internal_create_and_open(threadpool);
+        if (channel_internal == NULL)
+        {
+            /*Codes_SRS_CHANNEL_43_002: [ If there are any failures, channel_create shall fail and return NULL. ]*/
+            LogError("Failure in THANDLE_MALLOC(CHANNEL_INTERNAL)(channel_internal_dispose)");
```
jebrando, line 51: >THANDLE_MALLOC(CHANNEL_INTERNAL)(channel_internal_dispose [](http://example.com/codeflow?start=33&length=57)

I think this is the wrong log, should say channel_internal_create_and_open

devdoc/channel_internal.md
```diff
@@ -63,17 +63,15 @@ MOCKABLE_FUNCTION(, CHANNEL_RESULT, channel_internal_push, THANDLE(CHANNEL_INTER
...
+**SRS_CHANNEL_INTERNAL_43_095: [** `channel_internal_close` shall iterate over the list of pending operations and do the following: **]**
 
  - **SRS_CHANNEL_INTERNAL_43_096: [** set the `result` of the `operation` to `CHANNEL_CALLBACK_RESULT_ABANDONED`. **]**
 
  - **SRS_CHANNEL_INTERNAL_43_097: [** call `threadpool_schedule_work` with `execute_callbacks` as `work_function`. **]**
 
-**SRS_CHANNEL_INTERNAL_43_100: [** `channel_internal_dispose` shall call `srw_lock_release_exclusive`. **]**
-
-**SRS_CHANNEL_INTERNAL_43_092: [** `channel_internal_dispose` shall release the reference to `THANDLE(CHANNEL_INTERNAL)`. **]**
+**SRS_CHANNEL_INTERNAL_43_100: [** `channel_internal_close` shall call `srw_lock_release_exclusive`. **]**
```
anporumb, line 74: >srw_lock_release_exclusive [](http://example.com/codeflow?start=72&length=26)

43_097+43_100 means that the threadpool miiiight execute its callbacks while the lock on the operations is taken, isn't it?

Now, what happens if the user (in one of its callbacks decides to call `channel_internal_pull` for example?

anporumb:
> `channel_internal` is a module that contains the implementation for the `channel` module.

should specify somewhere in this file the threading model and reentrancy model.

---
Refers to: devdoc/channel_internal.md:5 in b011019. [](commit_id = b0110194e55ec17265f55712712be8ca2a3be55f, deletion_comment = False)

anporumb:
> MOCKABLE_FUNCTION(, void, channel_internal_close, THANDLE(CHANNEL_INTERNAL), channel_internal);

This function name is misleading.

What's its semantics?

After the user calls `channel_internal_create_and_open` followed by "channel_internal_close" how does the user "re-open"?



---
Refers to: devdoc/channel_internal.md:34 in b011019. [](commit_id = b0110194e55ec17265f55712712be8ca2a3be55f, deletion_comment = False)

src/channel_internal.c
```diff
@@ -0,0 +1,395 @@
...
+
+#include "../inc/c_util/channel_internal.h"
+
+typedef struct CHANNEL_INTERNAL_TAG
+{
+    THANDLE(THREADPOOL) threadpool;
+    SRW_LOCK_HANDLE lock;
+    DLIST_ENTRY op_list;
+}
+CHANNEL_INTERNAL;
```
anporumb, line 31: extra line breaks detected.

src/channel_internal.c
```diff
@@ -0,0 +1,395 @@
...
+        {
+            LogError("Failure in threadpool_schedule_work(execute_callbacks, channel_op)");
+        }
+    }
+
+    /*Codes_SRS_CHANNEL_INTERNAL_43_100: [ channel_internal_close shall call srw_lock_release_exclusive. ]*/
+    srw_lock_release_exclusive(channel_internal_ptr->lock);
+}
+
+IMPLEMENT_MOCKABLE_FUNCTION(, THANDLE(CHANNEL_INTERNAL), channel_internal_create_and_open, THANDLE(THREADPOOL), threadpool)
```
anporumb, line 88: >threadpool [](http://example.com/codeflow?start=112&length=10)

is threadpool `NULL` a valid argument here?

src/channel_internal.c
```diff
@@ -0,0 +1,395 @@
...
+
+        /*Codes_SRS_CHANNEL_INTERNAL_43_145: [ execute_callbacks shall call the stored callback(s) with the result of the operation. ]*/
+        if (channel_op->pull_callback)
+        {
+            channel_op->pull_callback(channel_op->pull_context, channel_op->result, channel_op->data);
+        }
+        if (channel_op->push_callback)
+        {
+            channel_op->push_callback(channel_op->push_context, channel_op->result);
+        }
```
anporumb, line 149: why are both callbacks served from the same thread? 

Wouldn't it make more sense that that pull callback and the push callbacks can be served at the same time?

src/channel_internal.c
```diff
@@ -0,0 +1,395 @@
...
+
+        /*Codes_SRS_CHANNEL_INTERNAL_43_147: [ execute_callbacks shall perform cleanup of the operation. ]*/
+        THANDLE_ASSIGN(RC_PTR)(&channel_op->data, NULL);
+        THANDLE_ASSIGN(CHANNEL_INTERNAL)(&channel_op->channel_internal, NULL);
+
+        // copy to local reference to avoid cutting the branch you are sitting on
+        THANDLE(ASYNC_OP) temp = NULL;
+        THANDLE_INITIALIZE_MOVE(ASYNC_OP)(&temp, &channel_op->async_op);
+        THANDLE_ASSIGN(ASYNC_OP)(&temp, NULL);
```
anporumb, line 158: I do not understand why these lines of code are needed.

They seem to do 

```
THANDLE_ASSIGN(ASYNC_OP)(&channel_op->async_op NULL);
```

src/channel_internal.c
```diff
@@ -0,0 +1,395 @@
...
+    }
+    else
+    {
+        result = 0;
+    }
+    return result;
+}
+
+
+IMPLEMENT_MOCKABLE_FUNCTION(, CHANNEL_RESULT, channel_internal_pull, THANDLE(CHANNEL_INTERNAL), channel_internal, PULL_CALLBACK, pull_callback, void*, pull_context, THANDLE(ASYNC_OP)*, out_op_pull)
```
jebrando, line 288: >CHANNEL_RESULT, channel_internal_pull [](http://example.com/codeflow?start=30&length=37)

This function doesn't check it's parameters, bad things can happen.

src/channel_internal.c
```diff
@@ -0,0 +1,395 @@
...
+            result = CHANNEL_RESULT_ERROR;
+        }
+    }
+    /*Codes_SRS_CHANNEL_INTERNAL_43_115: [ channel_internal_pull shall call srw_lock_release_exclusive. ]*/
+    srw_lock_release_exclusive(channel_internal_ptr->lock);
+
+    return result;
+}
+
+IMPLEMENT_MOCKABLE_FUNCTION(, CHANNEL_RESULT, channel_internal_push, THANDLE(CHANNEL_INTERNAL), channel_internal, THANDLE(RC_PTR), data, PUSH_CALLBACK, push_callback, void*, push_context, THANDLE(ASYNC_OP)*, out_op_push)
```
jebrando, line 342: >channel_internal_push [](http://example.com/codeflow?start=46&length=21)

Parameters need to be checked here.

src/channel_internal.c
```diff
@@ -0,0 +1,395 @@
...
+static void cancel_op(void* context)
+{
+    CHANNEL_OP* channel_op = context;
+    CHANNEL_INTERNAL* channel_internal_ptr = THANDLE_GET_T(CHANNEL_INTERNAL)(channel_op->channel_internal);
+
+    /*Codes_SRS_CHANNEL_INTERNAL_43_134: [ cancel_op shall call srw_lock_acquire_exclusive. ]*/
+    srw_lock_acquire_exclusive(channel_internal_ptr->lock);
+    {
+        /*Codes_SRS_CHANNEL_INTERNAL_43_135: [ If the operation is in the list of pending operations: ]*/
+        if (channel_op->anchor.Flink != &channel_op->anchor && &channel_op->anchor == channel_op->anchor.Flink->Blink)
```
anporumb, line 171: >channel_op->anchor.Flink != &channel_op->anchor && &channel_op->anchor == channel_op->anchor.Flink->Blink [](http://example.com/codeflow?start=12&length=105)

why are both of these conditions necessary?

src/channel_internal.c
```diff
@@ -0,0 +1,395 @@
...
+            THANDLE_INITIALIZE_MOVE(CHANNEL_INTERNAL)(&result, &channel_internal);
+            goto all_ok;
+        }
+        srw_lock_destroy(lock);
+    }
+all_ok:
+    return result;
+}
+
+static void execute_callbacks(void* context)
```
anporumb, line 131: >execute_callbacks [](http://example.com/codeflow?start=12&length=17)

I don't think execute_callbacks can access any of the fields of channel_op in the absence of locks.

This is (at least) because in cancel_op the field "result" is accessed under a lock.

src/channel_internal.c
```diff
@@ -0,0 +1,395 @@
...
+    srw_lock_release_exclusive(channel_internal_ptr->lock);
+}
+
+static void dispose_channel_op(void* context)
+{
+    /* don't need to do anything here because actual cleanup happens in execute_callbacks */
+    (void)context;
+}
+
+static int enqueue_operation(CHANNEL_INTERNAL* channel_internal, THANDLE(ASYNC_OP)* out_op, PULL_CALLBACK pull_callback, void* pull_context, PUSH_CALLBACK push_callback, void* push_context, THANDLE(RC_PTR) data)
```
anporumb, line 201: >CHANNEL_INTERNAL [](http://example.com/codeflow?start=29&length=16)

this needs to be a THANDLE(CHANNEL_INTERNAL) because in line 225 is used to initializechannel_op->channel_internal.

anporumb, line 201: >enqueue_operation [](http://example.com/codeflow?start=11&length=17)

I don't like this function. It is misleading.

I believe it is misleading because it populates the fields and enqueues. However, it only ever builds HALF of the operations (either the pull half or the push half).

At any time when this function is called, either PUSH or PULL callback are going to be NULL.

src/channel_internal.c
```diff
@@ -0,0 +1,395 @@
...
+
+        /*Codes_SRS_CHANNEL_INTERNAL_43_107: [ channel_internal_pull shall set *out_op_pull to the created THANDLE(ASYNC_OP). ]*/
+        /*Codes_SRS_CHANNEL_INTERNAL_43_123: [ channel_internal_push shall set *out_op_push to the created THANDLE(ASYNC_OP). ]*/
+        THANDLE_INITIALIZE_MOVE(ASYNC_OP)(out_op, &async_op);
+        result = 0;
+    }
+    return result;
+}
+
+static int dequeue_operation(CHANNEL_INTERNAL* channel_internal, THANDLE(ASYNC_OP)* out_op, PULL_CALLBACK pull_callback, void* pull_context, PUSH_CALLBACK push_callback, void* push_context, THANDLE(RC_PTR) data)
```
anporumb, line 245: >dequeue_operation [](http://example.com/codeflow?start=11&length=17)

Is this function ever called with both PUSH/PULL callbacks set to non-NULL values?

src/channel_internal.c
```diff
@@ -0,0 +1,395 @@
...
+        result = 0;
+    }
+    return result;
+}
+
+
+IMPLEMENT_MOCKABLE_FUNCTION(, CHANNEL_RESULT, channel_internal_pull, THANDLE(CHANNEL_INTERNAL), channel_internal, PULL_CALLBACK, pull_callback, void*, pull_context, THANDLE(ASYNC_OP)*, out_op_pull)
+{
+    CHANNEL_RESULT result;
+    CHANNEL_INTERNAL* channel_internal_ptr = THANDLE_GET_T(CHANNEL_INTERNAL)(channel_internal);
```
anporumb, line 291: >channel_internal [](http://example.com/codeflow?start=77&length=16)

404 parameter validation not found.

Here and everywhere

src/channel_internal.c
```diff
@@ -0,0 +1,395 @@
...
+            else
+            {
+                /* Codes_SRS_CHANNEL_INTERNAL_43_011: [ channel_internal_pull shall succeeds and return CHANNEL_RESULT_OK. ]*/
+                result = CHANNEL_RESULT_OK;
+            }
+        }
+        else
+        {
+            /*Codes_SRS_CHANNEL_INTERNAL_43_023: [ If there are any failures, channel_internal_pull shall fail and return CHANNEL_RESULT_ERROR. ]*/
+            LogError("THANDLE(CHANNEL) channel = %p is in unexpected state", channel_internal);
```
anporumb, line 332: >LogError("THANDLE(CHANNEL) channel = %p is in unexpected state", channel_internal); [](http://example.com/codeflow?start=12&length=83)

how did you write unit tests for these lines of code? Just curious.

tests/channel_internal_ut/channel_internal_ut.c
```diff
@@ -0,0 +1,774 @@
...
+    ASSERT_ARE_EQUAL(char_ptr, umock_c_get_expected_calls(), umock_c_get_actual_calls());
+
+    //cleanup
+    THANDLE_ASSIGN(ASYNC_OP)(&push_op, NULL);
+    THANDLE_ASSIGN(ASYNC_OP)(&pull_op, NULL);
+    THANDLE_ASSIGN(RC_PTR)(&data, NULL);
+    THANDLE_ASSIGN(CHANNEL_INTERNAL)(&channel_internal, NULL);
+}
+
+END_TEST_SUITE(TEST_SUITE_NAME_FROM_CMAKE)
```
anporumb, line 774: >TEST_SUITE_NAME_FROM_CMAKE [](http://example.com/codeflow?start=15&length=26)

I've briefly ran the code coverage for these and they are not touching all the lines of code.

Here's SOME lines that are not exercised from channel_internal.c: 387, 322, 276, 80.

tests/reals_ut/c_util_reals_ut.c
```diff
@@ -46,12 +52,15 @@ TEST_FUNCTION(check_all_c_util_reals)
     // arrange
 
     // act
+    REGISTER_ASYNC_OP_GLOBAL_MOCK_HOOKS();
+    REGISTER_CHANNEL_GLOBAL_MOCK_HOOKS();
```
anporumb, line 56: >REGISTER_CHANNEL_GLOBAL_MOCK_HOOKS [](http://example.com/codeflow?start=4&length=34)

missing channel_internal

tests/channel_int/channel_int.c
```diff
@@ -0,0 +1,513 @@
...
+}
+
+static void pull_callback_order_checker(void* context, CHANNEL_CALLBACK_RESULT result, THANDLE(RC_PTR) data)
+{
+    ASSERT_IS_NOT_NULL(context);
+    ASSERT_ARE_EQUAL(CHANNEL_CALLBACK_RESULT, CHANNEL_CALLBACK_RESULT_OK, result);
+    ASSERT_IS_NOT_NULL(data);
+
+    //assert that n-th pull is matched with n-th push
+     ASSERT_ARE_EQUAL(int32_t, *((int32_t*)context), (int32_t)(int64_t)((void*)data->ptr)); //casts needed to suppress pointer truncation warning
```
anporumb, line 123: >(int32_t)(int64_t) [](http://example.com/codeflow?start=53&length=18)

also known as uint_ptr_t?

tests/channel_int/channel_int.c
```diff
@@ -0,0 +1,513 @@
...
+
+BEGIN_TEST_SUITE(TEST_SUITE_NAME_FROM_CMAKE)
+
+TEST_SUITE_INITIALIZE(suite_init)
+{
+    g_execution_engine = execution_engine_create(NULL);
+    ASSERT_IS_NOT_NULL(g_execution_engine);
+    THANDLE(THREADPOOL) temp = threadpool_create(g_execution_engine);
+    ASSERT_IS_NOT_NULL(temp);
+    THANDLE_INITIALIZE_MOVE(THREADPOOL)(&g.g_threadpool, &temp);
```
anporumb, line 186: >g_threadpool [](http://example.com/codeflow?start=43&length=12)

all these tests should run with the default config of the treadpool and with a config of "1" thread only in the threadpool.

tests/channel_int/channel_int.c
```diff
@@ -0,0 +1,513 @@
...
+    volatile_atomic int32_t context;
+    (void)interlocked_exchange(&context, 0);
+
+    /// act
+    THANDLE(ASYNC_OP) async_op = NULL;
+    CHANNEL_RESULT result = channel_pull(channel, test_on_pull_callback_cancelled, (void*)&context, &async_op);
+    async_op_cancel(async_op);
+
+    //wait for callback to execute
+    wait_on_address(&context, 0, UINT32_MAX);
```
anporumb, line 233: >wait_on_address [](http://example.com/codeflow?start=4&length=15)

wait_on_address is always unstable and always can have spurious wakeups.

Use interlocked_hl_Wait...

tests/channel_int/channel_int.c
```diff
@@ -0,0 +1,513 @@
...
+
+}
+
+TEST_FUNCTION(test_push_and_cancel)
+{
+    /// arrange
+    THANDLE(CHANNEL) channel = channel_create(g.g_threadpool);
+    volatile_atomic int32_t context;
+    (void)interlocked_exchange(&context, 0);
+    THANDLE(RC_PTR) channel_data = rc_ptr_create_with_move_pointer(test_data, test_free_channel_data);
```
anporumb, line 252: >channel_data [](http://example.com/codeflow?start=20&length=12)

can this channel_data be shipped under the globals?

tests/channel_int/channel_int.c
```diff
@@ -0,0 +1,513 @@
...
+static int32_t pull_cancelled = 0x0002;
+static int32_t push_cancelled = 0x0003;
+static int32_t pull_success = 0x0004;
+static int32_t push_success = 0x0005;
+static int32_t pull_abandoned = 0x0006;
+static int32_t push_abandoned = 0x0007;
+static void* test_data2 = (void*)0x0008;
+
+
+static void test_on_pull_callback_cancelled(void* context, CHANNEL_CALLBACK_RESULT result, THANDLE(RC_PTR) data)
```
anporumb, line 44: >test_on_pull_callback_cancelled [](http://example.com/codeflow?start=12&length=31)

should assert that context is not NULL.

anporumb, line 44: >test_on_pull_callback_cancelled [](http://example.com/codeflow?start=12&length=31)

these "test" callbacks would be so much more easier to read if they'd have a int32_t* original = context, just so that the reader would understand what the original context used to be at the time when the function was called.

tests/channel_int/channel_int.c
```diff
@@ -0,0 +1,513 @@
...
+    /// act
+    THANDLE(ASYNC_OP) async_op = NULL;
+    CHANNEL_RESULT result = channel_pull(channel, test_on_pull_callback_cancelled, (void*)&context, &async_op);
+    async_op_cancel(async_op);
+
+    //wait for callback to execute
+    wait_on_address(&context, 0, UINT32_MAX);
+
+    /// assert
+    ASSERT_IS_NOT_NULL(async_op);
```
anporumb, line 236: >ASSERT_IS_NOT_NULL(async_op); [](http://example.com/codeflow?start=4&length=29)

should be next line after channel_pull.

tests/channel_int/channel_int.c
```diff
@@ -0,0 +1,513 @@
...
+    // cleanup
+    THANDLE_ASSIGN(CHANNEL)(&channel, NULL);
+}
+
+TEST_FUNCTION(test_pull_and_cancel)
+{
+    /// arrange
+    THANDLE(CHANNEL) channel = channel_create(g.g_threadpool);
+    volatile_atomic int32_t context;
+    (void)interlocked_exchange(&context, 0);
```
anporumb, line 225: >0 [](http://example.com/codeflow?start=41&length=1)

0 here is not a very good value, specially since it is tested for value in the callback.

Please use a random value.

tests/channel_int/channel_int.c
```diff
@@ -0,0 +1,513 @@
...
+    /// assert
+    ASSERT_IS_NOT_NULL(async_op);
+    ASSERT_ARE_EQUAL(CHANNEL_RESULT, CHANNEL_RESULT_OK, result);
+    ASSERT_ARE_EQUAL(int32_t, pull_cancelled, interlocked_add(&context, 0));
+
+    // cleanup
+    THANDLE_ASSIGN(ASYNC_OP)(&async_op, NULL);
+    THANDLE_ASSIGN(CHANNEL)(&channel, NULL);
+
+}
```
anporumb, line 244: please apply all comments from this TEST to all the tests in this unit.

anporumb (approved):
:shipit:

parth21999:
> `channel_internal` is a module that contains the implementation for the `channel` module.

channel.md contains that


---
In reply to: [1664321347](https://github.com/Azure/c-util/pull/294#issuecomment-1664321347) [](http://example.com/codeflow?ancestors=1664321347)

---
Refers to: devdoc/channel_internal.md:5 in 3b0a0b0. [](commit_id = 3b0a0b0f7d049211f4e34ad14458d4ea81cc24cd, deletion_comment = False)

parth21999:
> MOCKABLE_FUNCTION(, void, channel_internal_close, THANDLE(CHANNEL_INTERNAL), channel_internal);

There is no need for reopen. `channel_internal_close` is called when the user of `CHANNEL` releases all references to `THANDLE(CHANNEL)`. At that time, the channel is considered disposed by the user and the user cannot reuse that channel.


---
In reply to: [1664326018](https://github.com/Azure/c-util/pull/294#issuecomment-1664326018) [](http://example.com/codeflow?ancestors=1664326018)

---
Refers to: devdoc/channel_internal.md:34 in 3b0a0b0. [](commit_id = 3b0a0b0f7d049211f4e34ad14458d4ea81cc24cd, deletion_comment = False)

devdoc/channel_internal.md
parth21999: The "user" of `channel_internal` is just `channel`. `channel` calls `channel_internal_close` when the user of `channel` releases their last reference to `THANDLE(CHANNEL)`. Since the user of `channel` has released their last reference, they can no longer call `channel_pull`, so `channel_internal_pull` can no longer be called either.

src/channel_internal.c
parth21999: argument validation is performed at the `channel` module level, no need to perform again at `channel_internal` level. `channel_internal` is only meant to be used by `channel`.

src/channel_internal.c
parth21999: `channel_op` is the same as `async_op->context`.

`channel_op` also holds a `THANDLE(ASYNC_OP)` reference. So in a way, the async_op holds a reference to itself. 

If we did:

```c
THANDLE_ASSIGN(ASYNC_OP)(&channel_op->async_op NULL);
```

Then the memory for the async_op would get freed before the NULL assignment would happen. The NULL assignment would be an invalid access.

src/channel_internal.c
parth21999: The lock is actually just needed while removing the operation from the list, does not need to be held while accessing result.

src/channel_internal.c
parth21999: Each pull/push is only half an operation. `enqueue_operation` does half the work. What you have described is precisely what I intend it to do. 

The alternative would be to write the same code for pull  and push, with the only difference being which fields are set.

src/channel_internal.c
parth21999: No

parth21999: `dequeue_operation` does the symmetric "other half" of work as `enqueue_operation`

src/channel_internal.c
parth21999: Guaranteed to be non-NULL because it is given by `channel` module.

src/channel_internal.c
parth21999: Cannot really. One of the above cases should always be true. Added this check in case something goes horribly wrong in production.

src/channel_internal.c
parth21999: Parameter validation is done by `channel`

src/channel_internal.c
parth21999: Moving release so that only DList operation is under lock.

src/channel_internal.c
parth21999: Let the first one by C1, second one be C2.

Truth table:

C1T, C2T: Flink points to something else, and that something points back to it. Part of a non-empty list. 

C1T, C2F: Flink points to something else, that something doesn't point back to it. Cannot be part of a list.

C1F, C2TL: Flink points to itself. Not part of list.

C1F, C2F: Cannot happen.

Both conditions are needed.

src/channel_internal.c
parth21999: arg validation is done by `channel`.

src/channel_internal.c
parth21999: We want push callback to execute after pull callback has succeeded.

tests/channel_int/channel_int.c
parth21999: its not unsigned

tests/channel_int/channel_int.c
parth21999: done

PR #297, opened 2023-08-25
task: make worker thread OS agnostic

jebrando:


anporumb:
>     (the number of executed WORKER_FUNC calls is <= number of worker_thread_schedule_process calls made ) */

I do not understand this note... 

---
Refers to: devdoc/worker_thread_requirements.md:13 in b044c02. [](commit_id = b044c0297dd29e5da79598b0f89075ffe4cdefed, deletion_comment = False)

anporumb:
> **SRS_WORKER_THREAD_01_028: [** `worker_thread_open` shall start a `worker_thread` thread that will call `worker_func` by calling `ThreadAPI_Create`. **]**

We could/should use here a THREADPOOL. Otherwise we create a thread for... waiting idle most of the time.

./consider

---
Refers to: devdoc/worker_thread_requirements.md:80 in b044c02. [](commit_id = b044c0297dd29e5da79598b0f89075ffe4cdefed, deletion_comment = False)

anporumb:
> ## Overview

404 threading model not found (seems to be a case of threaded module, but should be explicitly stated so).

---
Refers to: devdoc/worker_thread_requirements.md:3 in b044c02. [](commit_id = b044c0297dd29e5da79598b0f89075ffe4cdefed, deletion_comment = False)

anporumb:
> ## Overview

404 reentrancy model not found.

can `worker_func` from line 25 call `worker_thread_schedule_process`? would that result in 
1) execution
2) stack overflow
3) none of the above

can `worker_func` from line 25 call `worker_thread_close` or `worker_thread_destroy`?

---
Refers to: devdoc/worker_thread_requirements.md:3 in b044c02. [](commit_id = b044c0297dd29e5da79598b0f89075ffe4cdefed, deletion_comment = False)

devdoc/worker_thread_requirements.md
```diff
@@ -47,9 +47,7 @@ MOCKABLE_FUNCTION(, WORKER_THREAD_HANDLE, worker_thread_create, WORKER_FUNC, wor
 
 - **SRS_WORKER_THREAD_01_037: [** `worker_thread_create` shall create a state manager object by calling `sm_create` with the name `worker_thread`. **]**
 
-- **SRS_WORKER_THREAD_01_006: [** `worker_thread_create` shall create an auto reset unnamed event used for signaling the thread to shutdown by calling `CreateEvent`. **]**
-
-- **SRS_WORKER_THREAD_01_007: [** `worker_thread_create` shall create an auto reset unnamed event used for signaling that a new execution of the worker function should be done by calling `CreateEvent`. **]**
+- **SRS_WORKER_THREAD_01_006: [** `worker_thread_create` shall initialize the internal objects. **]**
```
anporumb, line 50: > internal objects [](http://example.com/codeflow?start=77&length=17)

cannot write unit tests that test "internal objects" without those being specified what they are. Here it is needed a more concise phrasing.

anporumb:
> **SRS_WORKER_THREAD_01_039: [** If the worker thread is open, `worker_thread_destroy` shall perform a close. **]**

If the thread is created with ThreadAPI_Create (line 89) then we should say explicitly what's the function that "closes".

---
Refers to: devdoc/worker_thread_requirements.md:64 in b044c02. [](commit_id = b044c0297dd29e5da79598b0f89075ffe4cdefed, deletion_comment = False)

anporumb:
> **SRS_WORKER_THREAD_01_039: [** If the worker thread is open, `worker_thread_destroy` shall perform a close. **]**

aaaaahm, here you mean `worker_thread_close`, got it.

---
In reply to: [1696362231](https://github.com/Azure/c-util/pull/297#issuecomment-1696362231) [](http://example.com/codeflow?ancestors=1696362231)

---
Refers to: devdoc/worker_thread_requirements.md:64 in b044c02. [](commit_id = b044c0297dd29e5da79598b0f89075ffe4cdefed, deletion_comment = False)

anporumb:
>     typedef void(*WORKER_FUNC)(void* worker_func_context);

why is this not `THREAD_START_FUNC`?

---
Refers to: devdoc/worker_thread_requirements.md:14 in b044c02. [](commit_id = b044c0297dd29e5da79598b0f89075ffe4cdefed, deletion_comment = False)

anporumb:
>     MOCKABLE_FUNCTION(, WORKER_THREAD_HANDLE, worker_thread_create, WORKER_FUNC, worker_func, void*, worker_func_context);

How does WORKER_FUNC return if it failed somehow? How does the user that calls `worker_thread_schedule_process` know if the function was executed (and with what result?) or the design here is that "result" is passed through `worker_func_context` back to the caller?

---
Refers to: devdoc/worker_thread_requirements.md:25 in b044c02. [](commit_id = b044c0297dd29e5da79598b0f89075ffe4cdefed, deletion_comment = False)

jebrando:
>     (the number of executed WORKER_FUNC calls is <= number of worker_thread_schedule_process calls made ) */

I believe this is stating that if you call worker_thread_schedule_process while WORKER_FUNC is running, it may not schedule that thread due to the signal not being sent.  That is no longer an issue, I will remove.

---
In reply to: [1696355813](https://github.com/Azure/c-util/pull/297#issuecomment-1696355813) [](http://example.com/codeflow?ancestors=1696355813)

---
Refers to: devdoc/worker_thread_requirements.md:13 in f1fcb0d. [](commit_id = f1fcb0dc84d22ebe1e6b81802844b1a25d39b1b1, deletion_comment = False)

devdoc/worker_thread_requirements.md
```diff
@@ -125,21 +119,20 @@ MOCKABLE_FUNCTION(, WORKER_THREAD_SCHEDULE_PROCESS_RESULT, worker_thread_schedul
...
 ### worker_thread
 
-**SRS_WORKER_THREAD_01_019: [** The worker thread started by `worker_thread_create` shall wait for the 2 events: thread shutdown event and execute worker function event. **]**
+**SRS_WORKER_THREAD_01_019: [** The worker thread started by `worker_thread_create` shall get the module state. **]**
 
-**SRS_WORKER_THREAD_01_020: [** When the shutdown event is signaled, the worker thread shall exit. **]**
+**SRS_WORKER_THREAD_01_020: [** If the module state is `WORKER_THREAD_STATE_CLOSE`, the worker thread shall exit. **]**
 
-**SRS_WORKER_THREAD_01_021: [** When the execute worker function event is signaled, the worker thread shall call the `worker_func` function passed to `worker_thread_create` and it shall pass `worker_func_context` as argument. **]**
+**SRS_WORKER_THREAD_01_021: [** If the module state is `WORKER_THREAD_STATE_PROCESS_ITEM`, the worker thread shall call the `worker_func` function passed to `worker_thread_create` and it shall pass `worker_func_context` as argument... **]**
```
anporumb, line 134: >... [](http://example.com/codeflow?start=231&length=3)

?

jebrando:
>     typedef void(*WORKER_FUNC)(void* worker_func_context);

These question are out of scope for this PR, but I would guess that because THREAD_START_FUNC is for threading, this is slightly different since this is a worker thread and you want to differentiate this a little, hence giving it a different name.

---
In reply to: [1696364040](https://github.com/Azure/c-util/pull/297#issuecomment-1696364040) [](http://example.com/codeflow?ancestors=1696364040)

---
Refers to: devdoc/worker_thread_requirements.md:14 in f1fcb0d. [](commit_id = f1fcb0dc84d22ebe1e6b81802844b1a25d39b1b1, deletion_comment = False)

jebrando:
> **SRS_WORKER_THREAD_01_028: [** `worker_thread_open` shall start a `worker_thread` thread that will call `worker_func` by calling `ThreadAPI_Create`. **]**

I like that change.

---
In reply to: [1696357552](https://github.com/Azure/c-util/pull/297#issuecomment-1696357552) [](http://example.com/codeflow?ancestors=1696357552)

---
Refers to: devdoc/worker_thread_requirements.md:80 in f1fcb0d. [](commit_id = f1fcb0dc84d22ebe1e6b81802844b1a25d39b1b1, deletion_comment = False)

jebrando:
> **SRS_WORKER_THREAD_01_039: [** If the worker thread is open, `worker_thread_destroy` shall perform a close. **]**

I'm not sure what meant here, I'm guessing you are correct.

---
In reply to: [1696363056](https://github.com/Azure/c-util/pull/297#issuecomment-1696363056) [](http://example.com/codeflow?ancestors=1696363056,1696362231)

---
Refers to: devdoc/worker_thread_requirements.md:64 in f1fcb0d. [](commit_id = f1fcb0dc84d22ebe1e6b81802844b1a25d39b1b1, deletion_comment = False)

anporumb:
> **SRS_WORKER_THREAD_01_034: [** `worker_thread_close` shall signal the thread shutdown event in order to indicate that the thread shall shutdown. **]**

But... this is not event driven anymore. Seems to look at the mysterious "module state"

---
Refers to: devdoc/worker_thread_requirements.md:102 in b044c02. [](commit_id = b044c0297dd29e5da79598b0f89075ffe4cdefed, deletion_comment = False)

jebrando:
>     MOCKABLE_FUNCTION(, WORKER_THREAD_HANDLE, worker_thread_create, WORKER_FUNC, worker_func, void*, worker_func_context);

The only way for WORKER_FUNC to fail is for it not to get scheduled and that case worker_thread_schedule_process shall fail.  Besides that there is nothing that can fail in WORKER_FUNC and I'm guessing that is by design.

---
In reply to: [1696366670](https://github.com/Azure/c-util/pull/297#issuecomment-1696366670) [](http://example.com/codeflow?ancestors=1696366670)

---
Refers to: devdoc/worker_thread_requirements.md:25 in f1fcb0d. [](commit_id = f1fcb0dc84d22ebe1e6b81802844b1a25d39b1b1, deletion_comment = False)

devdoc/worker_thread_requirements.md
```diff
@@ -125,21 +119,20 @@ MOCKABLE_FUNCTION(, WORKER_THREAD_SCHEDULE_PROCESS_RESULT, worker_thread_schedul
...
 
-**SRS_WORKER_THREAD_01_018: [** If any other error occurs, `worker_thread_schedule_process` shall fail and return `WORKER_THREAD_SCHEDULE_PROCESS_ERROR`. **]**
-
 ### worker_thread
 
-**SRS_WORKER_THREAD_01_019: [** The worker thread started by `worker_thread_create` shall wait for the 2 events: thread shutdown event and execute worker function event. **]**
+**SRS_WORKER_THREAD_01_019: [** The worker thread started by `worker_thread_create` shall get the module state. **]**
 
-**SRS_WORKER_THREAD_01_020: [** When the shutdown event is signaled, the worker thread shall exit. **]**
+**SRS_WORKER_THREAD_01_020: [** If the module state is `WORKER_THREAD_STATE_CLOSE`, the worker thread shall exit. **]**
```
anporumb, line 132: >module state [](http://example.com/codeflow?start=39&length=12)

do we have "module state" and SM at the same time? This was not a very succesfull recipe in the past...

anporumb, line 132: >WORKER_THREAD_STATE_CLOSE [](http://example.com/codeflow?start=56&length=25)

likely should be explained in some enum or something.

devdoc/worker_thread_requirements.md
```diff
@@ -125,21 +119,20 @@ MOCKABLE_FUNCTION(, WORKER_THREAD_SCHEDULE_PROCESS_RESULT, worker_thread_schedul
...
 
-**SRS_WORKER_THREAD_01_021: [** When the execute worker function event is signaled, the worker thread shall call the `worker_func` function passed to `worker_thread_create` and it shall pass `worker_func_context` as argument. **]**
+**SRS_WORKER_THREAD_01_021: [** If the module state is `WORKER_THREAD_STATE_PROCESS_ITEM`, the worker thread shall call the `worker_func` function passed to `worker_thread_create` and it shall pass `worker_func_context` as argument... **]**
 
-**SRS_WORKER_THREAD_01_025: [** In case of any error, the worker thread shall exit. **]**
+**SRS_WORKER_THREAD_11_001: [** ... and set the module state to idle if it has not been changed. **]**
```
anporumb, line 136: >idle [](http://example.com/codeflow?start=64&length=4)

is "idle" here some enum value in the same enum as `WORKER_THREAD_STATE_CLOSE` and `WORKER_THREAD_STATE_PROCESS_ITEM`?

devdoc/worker_thread_requirements.md
```diff
@@ -125,21 +119,20 @@ MOCKABLE_FUNCTION(, WORKER_THREAD_SCHEDULE_PROCESS_RESULT, worker_thread_schedul
 
 **SRS_WORKER_THREAD_01_042: [** If `sm_exec_begin` does not grant the execution, `worker_thread_schedule_process` shall fail and return `WORKER_THREAD_SCHEDULE_PROCESS_INVALID_STATE`. **]**
 
-**SRS_WORKER_THREAD_01_017: [** `worker_thread_schedule_process` shall set the process event created in `worker_thread_create`. **]**
+**SRS_WORKER_THREAD_01_017: [** `worker_thread_schedule_process` shall set the state to `WORKER_THREAD_STATE_PROCESS_ITEM`. **]**
```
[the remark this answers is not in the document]
jebrando: Calling multiple worker_thread_schedule_process before the WORKER_FUNC is called will result in 1 WORKER_FUNC call

jebrando:
> **SRS_WORKER_THREAD_01_034: [** `worker_thread_close` shall signal the thread shutdown event in order to indicate that the thread shall shutdown. **]**

I'll change this to say the state.

---
In reply to: [1696379748](https://github.com/Azure/c-util/pull/297#issuecomment-1696379748) [](http://example.com/codeflow?ancestors=1696379748)

---
Refers to: devdoc/worker_thread_requirements.md:102 in f1fcb0d. [](commit_id = f1fcb0dc84d22ebe1e6b81802844b1a25d39b1b1, deletion_comment = False)

devdoc/worker_thread_requirements.md
jebrando: This state is only for the thread, not the module.  SM is for the module.  I'll make that clearer.

devdoc/worker_thread_requirements.md
```diff
@@ -125,21 +119,20 @@ MOCKABLE_FUNCTION(, WORKER_THREAD_SCHEDULE_PROCESS_RESULT, worker_thread_schedul
...
-**SRS_WORKER_THREAD_01_020: [** When the shutdown event is signaled, the worker thread shall exit. **]**
+**SRS_WORKER_THREAD_01_020: [** If the module state is `WORKER_THREAD_STATE_CLOSE`, the worker thread shall exit. **]**
 
-**SRS_WORKER_THREAD_01_021: [** When the execute worker function event is signaled, the worker thread shall call the `worker_func` function passed to `worker_thread_create` and it shall pass `worker_func_context` as argument. **]**
+**SRS_WORKER_THREAD_01_021: [** If the module state is `WORKER_THREAD_STATE_PROCESS_ITEM`, the worker thread shall call the `worker_func` function passed to `worker_thread_create` and it shall pass `worker_func_context` as argument... **]**
 
-**SRS_WORKER_THREAD_01_025: [** In case of any error, the worker thread shall exit. **]**
+**SRS_WORKER_THREAD_11_001: [** ... and set the module state to idle if it has not been changed. **]**
 
+**SRS_WORKER_THREAD_11_002: [** If the module state is `WORKER_THREAD_STATE_IDLE`, the worker thread shall wait for the state to transition to something else. **]**
```
anporumb, line 138: >**]** [](http://example.com/codeflow?start=159&length=5)

having read everything until now, I find that the promises of this module are a bit lacking in regards to timing and GREAT care needs to be exercised by the user.

Here's what I mean. Let's say the user wants to execute 3 (times) the `worker_func`. Let's say `worker func` take 100 ms to execute.

Here's 2 timelines
Timeline1:
T+0 => schedule process
T+1 => `worker func` starts to execute.
T+101 = `worker func` is done executing(1st execution).

T+200 => schedule process
T+201 => `worker func` starts to execute.
T+301 = `worker func` is done executing(2nd execution).

T+400 => schedule process
T+401 => `worker func` starts to execute.
T+501 = `worker func` is done executing(3rd execution).

Timeline2:
T+0 => schedule process
T+1 => `worker func` starts to execute.
T+25 => schedule process (sets state to "WORKER_THREAD_STATE_PROCESS_ITEM as per `SRS_WORKER_THREAD_01_017`)
T+75 => schedule process (sets state to "WORKER_THREAD_STATE_PROCESS_ITEM as per `SRS_WORKER_THREAD_01_017`)
T+101 => `worker func` is done executing(1st execution)
T+102 => `worker func` executes again (because state is WORKER_THREAD_STATE_PROCESS_ITEM)
T+202 => `worker func` ends and NOTHING executed anymore.

The difference between Timeline 1 and Timeline 2 is that in Timeline 2 there are 2 executions, while in Timeline 1 there are 3, and the only difference is in the timing when `schedule process` is called.

I believe the user's intention is to always have 3 executions. And we can talk about it.

src/worker_thread.c
```diff
@@ -25,42 +25,48 @@ typedef struct WORKER_THREAD_TAG
     WORKER_FUNC worker_func;
     void* worker_func_context;
     THREAD_HANDLE thread_handle;
-    HANDLE worker_thread_shutdown_event;
-    HANDLE execute_worker_event;
+    volatile_atomic int32_t state;
```
anporumb, line 28: >volatile_atomic int32_t state; [](http://example.com/codeflow?start=4&length=30)

if this state is "the same as" `WORKER_THREAD_STATE` then consider having a union over the two. When printing it with PRI_MU_ENUM, use the enum field of the union; when manipulating it with interlocked, use the volatile_atomic member of the union.

src/worker_thread.c
```diff
@@ -25,42 +25,48 @@ typedef struct WORKER_THREAD_TAG
...
+        // Codes_SRS_WORKER_THREAD_01_019: [ The worker thread started by worker_thread_create shall get the module state. ]
+        WORKER_THREAD_STATE current_state = interlocked_add(&worker_thread->state, 0);
+        switch (current_state)
         {
-            /* Codes_SRS_WORKER_THREAD_01_020: [ When the shutdown event is signaled, the worker thread shall exit. ]*/
-            result = 0;
-            break;
+            default:
+            case WORKER_THREAD_STATE_IDLE:
+                // Codes_SRS_WORKER_THREAD_11_002: [ If the module state is WORKER_THREAD_STATE_IDLE, the worker thread shall wait for the state to transition to something else. ]
```
anporumb, line 54: >the state to transition to something else [](http://example.com/codeflow?start=135&length=41)

What happens if you'd use InterlockedHL_WaitForNotValue?

src/worker_thread.c
```diff
@@ -25,42 +25,48 @@ typedef struct WORKER_THREAD_TAG
...
+                (void)wait_on_address(&worker_thread->state, current_state, UINT32_MAX);
```
anporumb, line 55: >wait_on_address [](http://example.com/codeflow?start=22&length=15)

can have spurious wake ups.

src/worker_thread.c
```diff
@@ -25,42 +25,48 @@ typedef struct WORKER_THREAD_TAG
...
+                break;
+            case WORKER_THREAD_STATE_PROCESS_ITEM:
+                // Codes_SRS_WORKER_THREAD_11_001: [ ... and set the module state to idle if it has not been changed. ]
+                (void)interlocked_compare_exchange(&worker_thread->state, WORKER_THREAD_STATE_IDLE, current_state);
```
anporumb, line 59: >interlocked_compare_exchange [](http://example.com/codeflow?start=22&length=28)

why not interlocked_exchange? it doesn't look like the "current_state" is examined.

anporumb:
>             LogError("Cannot allocate memory for BS worker thread");

wouold be nice if the API would take an optional RC_STRING that would contain some "thread_name" just so we know which one of the many worker threads failed...

---
Refers to: src/worker_thread.c:92 in b044c02. [](commit_id = b044c0297dd29e5da79598b0f89075ffe4cdefed, deletion_comment = False)

src/worker_thread.c
```diff
@@ -100,39 +106,14 @@ WORKER_THREAD_HANDLE worker_thread_create(WORKER_FUNC worker_func, void* worker_
...
-
-                        //(void)CloseHandle(worker_thread->execute_worker_event);
-                    }
-
-                    (void)CloseHandle(worker_thread->worker_thread_shutdown_event);
-                }
-
-                sm_destroy(worker_thread->sm);
+                /* Codes_SRS_WORKER_THREAD_01_006: [ worker_thread_create shall initialize the internal objects. ]*/
+                (void)interlocked_exchange(&worker_thread->state, 0);
```
anporumb, line 110: >0 [](http://example.com/codeflow?start=66&length=1)

what's the meaning of this "0" here? `WORKER_THREAD_STATE` enum doesn't have it listed explicitly.

devdoc/worker_thread_requirements.md
jebrando: I too was concerned about this when I looked at it.  It looks like the contract is that schedule process must not be called until WORKER_FUNC has been called (this is stated on cryptic line 13 in the spec).  If this is indeed the contract (which I looked at the ebs code that is calling this module) then the 2nd timeline breaks the contract.

src/worker_thread.c
```diff
@@ -144,7 +125,9 @@ static void internal_close(WORKER_THREAD_HANDLE worker_thread)
     int dont_care;
 
     /* Codes_SRS_WORKER_THREAD_01_034: [ worker_thread_close shall signal the thread shutdown event in order to indicate that the thread shall shutdown. ]*/
-    (void)SetEvent(worker_thread->worker_thread_shutdown_event);
+    (void)interlocked_exchange(&worker_thread->state, WORKER_THREAD_STATE_CLOSE);
```
anporumb, line 128: >WORKER_THREAD_STATE_CLOSE [](http://example.com/codeflow?start=54&length=25)

I disagree here.

Imagine the following sequence of events:
1) create
2) open
3) close => sets `worker_thread->state, WORKER_THREAD_STATE_CLOSE`
4) open. For the life of me if I see open resetting the state from `WORKER_THREAD_STATE_CLOSE to something else. So I would assume that it does not. So... then the thread will die as soon as it is spawned.

What this scenario tested?

src/worker_thread.c
jebrando: I do not want to change the value if it has been changed somewhere else.  For example between line 49 and 59 the user calls close, I don't want to change it back to IDLE, I want this call to not change the value and then handle the close call in the next iteration of the loop.

anporumb:
>     sm_close_end(worker_thread->sm);

the fact that sm_close_end has its sm_close_begin in 2!!! different places in the SW makes it really difficult to track.

Note: line 150 and line 214.

If at all possible (and I suspect it is) we should keep begin/ends in the same function.

---
Refers to: src/worker_thread.c:135 in b044c02. [](commit_id = b044c0297dd29e5da79598b0f89075ffe4cdefed, deletion_comment = False)

anporumb:
> END_TEST_SUITE(TEST_SUITE_NAME_FROM_CMAKE)

will be waiting because I think the semantics are not clear.

---
Refers to: tests/worker_thread_ut/worker_thread_ut.c:635 in b044c02. [](commit_id = b044c0297dd29e5da79598b0f89075ffe4cdefed, deletion_comment = False)

anporumb (changes requested):
:clock1:

jebrando:
>     sm_close_end(worker_thread->sm);

Yeah, that was inherited, so I'll move it.

---
In reply to: [1696412937](https://github.com/Azure/c-util/pull/297#issuecomment-1696412937) [](http://example.com/codeflow?ancestors=1696412937)

---
Refers to: src/worker_thread.c:135 in f1fcb0d. [](commit_id = f1fcb0dc84d22ebe1e6b81802844b1a25d39b1b1, deletion_comment = False)

jebrando:
>             LogError("Cannot allocate memory for BS worker thread");

/considering.  If I do I'll create a new PR, because of the changes that would entail.

---
In reply to: [1696406671](https://github.com/Azure/c-util/pull/297#issuecomment-1696406671) [](http://example.com/codeflow?ancestors=1696406671)

---
Refers to: src/worker_thread.c:92 in f1fcb0d. [](commit_id = f1fcb0dc84d22ebe1e6b81802844b1a25d39b1b1, deletion_comment = False)

devdoc/worker_thread_requirements.md
jebrando: I may know know what you're meaning here, but these are already enums.

jebrando:
> **SRS_WORKER_THREAD_01_028: [** `worker_thread_open` shall start a `worker_thread` thread that will call `worker_func` by calling `ThreadAPI_Create`. **]**

After looking at this, it only requires 1 thread to be created and it doesn't destroy the thread till the module is complete, so having a threadpool thread would not make any sense in having this.

---
In reply to: [1696372262](https://github.com/Azure/c-util/pull/297#issuecomment-1696372262) [](http://example.com/codeflow?ancestors=1696372262,1696357552)

---
Refers to: devdoc/worker_thread_requirements.md:80 in f1fcb0d. [](commit_id = f1fcb0dc84d22ebe1e6b81802844b1a25d39b1b1, deletion_comment = False)

devdoc/worker_thread_requirements.md
anporumb: This module needs to prohibit the overcluttering of calls (which results in missed executions).

devdoc/worker_thread_requirements.md
```diff
@@ -5,11 +5,19 @@
 `worker_thread` is a helper module that implements execution of a work item on a single separate thread. It can be used in batching operations, sequencing completions, etc.
 The user instantiates a worker thread instance and then signals when the work shall be executed by calling `worker_thread_schedule_process`.
 
+## Thread Concerns
+
+A calls to `worker_thread_schedule_process` must wait for the resulting `WORKER_FUNC` call or the call will not be delivered.  For this reason `worker_thread_schedule_process` function should be called in a single threaded manner.
```
anporumb, line 10: >should be called in a single threaded manner [](http://example.com/codeflow?start=185&length=44)

I believe it should enforce that.

What would happen if it would collect the threadId at the time of the `worker_thread_schedule_process` and then compare it (if `WORKER_FUNC` is still executing) with the thread id of the second call to `worker_thread_schedule_process`?

anporumb (approved):
:shipit:

devdoc/worker_thread_requirements.md
jebrando: It only creates 1 thread (in open), so the thread Id would be the same.

PR #300, opened 2023-09-15
Worker thread fix

dcristoloveanu:
-Add worker thread chaos test
-Fix worker thread failing if the state is PROCESS_ITEM


jebrando (approved):
:shipit:

PR #306, opened 2023-09-26
Fix real_hash for missing external linker error

mattdurak:
Fixes `error LNK2001: unresolved external symbol "unsigned int __cdecl MurmurHash2(void const *,int,unsigned int)" (?MurmurHash2@@YAIPEBXHI@Z)`


dcristoloveanu (approved):
:shipit:

PR #310, opened 2023-10-04
fix rc_string to not need umock_prod

anporumb:


tests/reals/real_rc_string.h
```diff
@@ -26,6 +26,8 @@
 #include <stdint.h>
 #include <stdarg.h>
 
+#include "umock_c/umock_c_prod.h"
```
mattdurak, line 29: >umock_c_prod [](http://example.com/codeflow?start=18&length=12)

I'm confused about how this fixes anything. rc_string.h is including umock_c_prod.h already so this line likely doesn't really do anything. And we don't include umock_c_prod.h explicitly in other reals from what I can tell.

mattdurak: oh, is it because `THANDLE_TYPE_DECLARE` generates mockables? Should we also have like `#undef ENABLE_MOCKS` before this include?

PR #315, opened 2023-10-11
add spec for 2d array

M-iceberg:


devdoc/two_dimensional_array_requirements.md
```diff
@@ -0,0 +1,84 @@
...
+## Overview
+
+`two_dimensional_array` is a module provide to declare a 2D array with variable size of rows and columns. Each rows has the same size to ensure easier access. The first dimentional of array is resizable, it contains the pointer to the second dimention of array. The pointer is intiazlied to NULL and points to the allocated memory address when there's not enough space. When the memory stored in a row is not needed any more, the entire row shall be freed and set back to `NULL`.
+
+## Exposed API
+```c
+typedef struct TWO_D_ARRAY_TAG
+{
+    uint64_t rows;
+    uint64_t cols;
```
darobs, line 13: I'd expect to see some field for the rows array at the end of the struct.

devdoc/two_dimensional_array_requirements.md
```diff
@@ -0,0 +1,84 @@
...
+} TWO_D_ARRAY_HANDLE;
```
darobs, line 14: >TWO_D_ARRAY_HANDLE [](http://example.com/codeflow?start=2&length=18)

Is the definition of the struct required here?
Can be:
typedef struct TWO_D_ARRAY_TAG* TWO_D_ARRAY_HANDLE;

devdoc/two_dimensional_array_requirements.md
```diff
@@ -0,0 +1,84 @@
...
+
+MOCKABLE_FUNCTION(, TWO_D_ARRAY_HANDLE, two_d_array_create, uint64_t, row_size, uint_64_t, col_size);
+MOCKABLE_FUNCTION(, int, two_d_array_free_row, TWO_D_ARRAY_HANDLE, handle, uint64_t, row_index);
+MOCKABLE_FUNCTION(, int, two_d_array_allocate_new_row, TWO_D_ARRAY_HANDLE, handle, uint64_t, row_index);
+MOCKABLE_FUNCTION(, void, two_d_array_destroy, TWO_D_ARRAY_HANDLE, handle);
+```
```
darobs, line 20: I don't see any access function here - how do we get a row, column, or element at row, column?

devdoc/two_dimensional_array_requirements.md
M-iceberg: added

devdoc/two_dimensional_array_requirements.md
M-iceberg: added

devdoc/two_dimensional_array_requirements.md
M-iceberg: added

devdoc/two_dimensional_array_requirements.md
```diff
@@ -0,0 +1,106 @@
+# two_dimensional_array_requirements
+============================
+
+## Overview
+
+`two_dimensional_array` provides a 2D array with variable size of rows and columns. Each row has the same size to ensure easier access. The first dimention of array is resizable, it contains the pointer to the second dimention of array. The pointer is initiazlied to NULL and points to the allocated memory address when there's not enough space. When the memory stored in a row is not needed any more, the entire row shall be freed and set back to `NULL`.
```
dcristoloveanu, line 6: >dimention [](http://example.com/codeflow?start=217&length=9)

dimension

dcristoloveanu, line 6: >dimention [](http://example.com/codeflow?start=146&length=9)

dimension

dcristoloveanu, line 6: Can we have some CRLFs so we break the lines so that we can read the text better?

dcristoloveanu, line 6: >initiazlied [](http://example.com/codeflow?start=252&length=11)

initialized

dcristoloveanu, line 6: >The pointer is initiazlied to NULL and points to the allocated memory address when there's not enough space. When the memory stored in a row is not needed any more, the entire row shall be freed and set back to `NULL`. [](http://example.com/codeflow?start=237&length=218)

I think this part is starting to explain some of the design/implementation. It belongs to a different section, let's have one called design.

That section can detail even more.

devdoc/two_dimensional_array_requirements.md
```diff
@@ -0,0 +1,106 @@
...
+
```
dcristoloveanu, line 7: Frankly a picture would help.

devdoc/two_dimensional_array_requirements.md
```diff
@@ -0,0 +1,106 @@
...
+{
+    uint64_t rows;
+    uint64_t cols;
+    size_t item_size;
+    void * row_arrays[];
+}TWO_D_ARRAY;
+
+typedef struct TWO_D_ARRAY_TAG* TWO_D_ARRAY_HANDLE;
+MOCKABLE_FUNCTION(, TWO_D_ARRAY_HANDLE, two_d_array_create, uint64_t, row_size, uint64_t, col_size, size_t, item_size);
+MOCKABLE_FUNCTION(, int, two_d_array_free_row, TWO_D_ARRAY_HANDLE, two_d_array_handle, uint64_t, row_index);
```
dcristoloveanu, line 20: This is a free row generic API, how does this help with sp_index? I think we started at redoing sp_index....

dcristoloveanu (changes requested):
:clock1:

dcristoloveanu:
Let's talk tomorrow as I think some of the comments need clarification.

PR #316, opened 2023-10-11
fix for a channel int tests

jebrando:


tests/channel_int/channel_int.c
```diff
@@ -232,19 +234,25 @@ TEST_FUNCTION(test_channel_create_and_destroy)
     /// arrange
 
     /// act
-    THANDLE(CHANNEL) channel = channel_create(g.g_threadpool);
+    THANDLE(THREADPOOL) threadpool = create_threadpool();
```
dcristoloveanu, line 237: >  THANDLE(THREADPOOL) thre [](http://example.com/codeflow?start=2&length=26)

This is part of arrange, so it should be placed appropriately in the test.

tests/channel_int/channel_int.c
```diff
@@ -232,19 +234,25 @@ TEST_FUNCTION(test_channel_create_and_destroy)
...
+
+    THANDLE(CHANNEL) channel = channel_create(threadpool);
 
     /// assert
     ASSERT_IS_NOT_NULL(channel);
 
     // cleanup
     THANDLE_ASSIGN(CHANNEL)(&channel, NULL);
+    threadpool_close(threadpool);
+    THANDLE_ASSIGN(THREADPOOL)(&threadpool, NULL);
```
dcristoloveanu, line 247: Isn't the fact that we're creating a threadpool for each test hiding a deeper issue?
I am not getting how this is solving the crash actually ...

dcristoloveanu (changes requested):
:clock1:

tests/channel_int/channel_int.c
jebrando: Fixed, this is also spawned a new task here [https://msazure.visualstudio.com/One/_workitems/edit/25500968](https://msazure.visualstudio.com/One/_workitems/edit/25500968)

build/devops_gated.yml
```diff
@@ -122,7 +122,7 @@ jobs:
 - job: linuxubuntu
   displayName: 'Build Linux Ubuntu'
   pool:
-    name: Azure-MessagingStore-LinuxBuildMachinePool
+    name: Azure-MsgStore-Linux2204BuildMachinePool
```
dcristoloveanu, line 125: >me: Azure-MsgStore-Linux2204BuildMachinePool [](http://example.com/codeflow?start=6&length=44)

Hmm, is this intended?

dcristoloveanu: This seems to be from another change? Or is it needed for this test to pass?

dcristoloveanu (approved):
:shipit:

build/devops_gated.yml
jebrando: The test fails more on the new version of linux, new helgrind I'm guessing.

PR #331, opened 2023-11-16
Fix real_worker_thread to not have mock interlocked calls

mattdurak:
Missing real renames causes problems later with using real clds (which starts a worker thread)


darobs (approved):
:shipit:

jebrando (approved):
:shipit:

PR #336, opened 2023-12-11
Copy a few test helpers from EBS

BertKleewein:


test_helpers/inc/c_util_test_helpers/c_util_test_env_params.h
```diff
@@ -0,0 +1,16 @@
+// Copyright (C) Microsoft Corporation. All rights reserved.
+
+#ifndef C_UTIL_TEST_ENV_PARAMS_H
```
dcristoloveanu, line 3: >C_UTIL_ [](http://example.com/codeflow?start=8&length=7)

Do we need the c_util prefix? A little bit of a mouth full. Is it because we plan to have another one somewhere else? Maybe we should not have another one as it is getting a bit too much to have multiple env params for test purposes. Will get pretty confusing me thinks.

dcristoloveanu (dismissed):
:shipit:

test_helpers/inc/c_util_test_helpers/c_util_test_env_params.h
BertKleewein: this function was pulled out of test_env_params.h in the EBS test_utils code.  Renaming to avoid naming conflict seemed prudent.

test_helpers/src/c_util_test_env_params.c
```diff
@@ -0,0 +1,13 @@
+// Copyright (C) Microsoft Corporation. All rights reserved.
+
+#include <stdlib.h>
+#include "c_util_test_helpers/c_util_test_env_params.h"
+
+#define DEFAULT_LEAK_LOG_FOLDER_NAME "g$\\blockstorage"
```
mattdurak, line 6: >blockstorage [](http://example.com/codeflow?start=42&length=12)

ha, uh block store is "leaking" into c_utils (here and below)

mattdurak: Is there really value in having a common environment variable like this across projects?

mattdurak (approved):
:shipit:

test_helpers/src/c_util_test_env_params.c
BertKleewein: You're right.  I was trying to limit code changes but it's cleaner to pass the root path in. I'm going to update this.

test_helpers/inc/c_util_test_helpers/c_util_test_env_params.h
BertKleewein: changing the code to accept a root path. Then we don't need this file at all and you and Matt are both happy.

dcristoloveanu (approved):
:shipit:

PR #339, opened 2024-01-02
Add cancellation token module

avranju:
This PR adds the cancellation token module which implements an object that can be used to propagate cancellation signals through multi-threaded programs. The design is roughly equivalent to the [`CancellationToken`](https://learn.microsoft.com/en-us/dotnet/api/system.threading.cancellationtoken?view=net-8.0) type in .NET. The module allows applications to create cancellation token handles which can then be passed around to various functions. The following operations can be performed on cancellation tokens:

- Check if the token has been canceled.
- Register a callback to be invoked when the token is canceled.
- Initiate cancellation of the token causing the callback registrations to be invoked and the state of the token to be changed to "canceled".


devdoc/cancellation_token_requirements.md
```diff
@@ -0,0 +1,138 @@
+# Cancellation Token Requirements
```
anporumb, line 1: >Cancellation Token Requirements [](http://example.com/codeflow?start=2&length=31)

what's the relation between this and ASYNC_OP in c_util?

My understanding is that to cancel something, it needs to be ASYNC ("in general").

So would be really sweet if there would be some clarification around that...

devdoc/cancellation_token_requirements.md
```diff
@@ -0,0 +1,138 @@
...
+
+**SRS_CANCELLATION_TOKEN_04_012: [** `cancellation_token_register_notify` shall fail and return `NULL` when any underlying call fails. **]**
+
+**SRS_CANCELLATION_TOKEN_04_013: [** `cancellation_token_register_notify` shall call `TCALL_DISPATCHER_REGISTER_TARGET(CANCELLATION_TOKEN_CANCEL_CALL)` to register `on_cancel` with the dispatcher. **]**
+
+**SRS_CANCELLATION_TOKEN_04_014: [** `cancellation_token_register_notify` shall call `THANDLE_MALLOC(CANCELLATION_TOKEN_REGISTRATION)` to allocate a `CANCELLATION_TOKEN_REGISTRATION` handle. **]**
+
+**SRS_CANCELLATION_TOKEN_04_015: [** `cancellation_token_register_notify` shall initialize and return a valid `THANDLE(CANCELLATION_TOKEN_REGISTRATION)` when successful. **]**
+
+**SRS_CANCELLATION_TOKEN_04_016: [** `cancellation_token_register_notify` shall call `TCALL_DISPATCHER_UNREGISTER_TARGET(CANCELLATION_TOKEN_CANCEL_CALL)` when an error occurs to undo the callback registration. **]**
```
mattdurak, line 120: > call `TCALL_DISPATCHER_UNREGISTER_TARGET(CANCELLATION_TOKEN_CANCEL_CALL)` when an error occurs to undo the callback registration. **]** [](http://example.com/codeflow?start=79&length=136)

if you malloc first then register, this wouldn't be needed, right? That also avoids races where you fail to malloc here and a cancel comes in and calls the callback that was briefly registered.

devdoc/cancellation_token_requirements.md
```diff
@@ -0,0 +1,138 @@
...
+**SRS_CANCELLATION_TOKEN_04_003: [** `cancellation_token_create` shall create a `TCALL_DISPATCHER` handle by calling `TCALL_DISPATCHER_CREATE(CANCELLATION_TOKEN_CANCEL_CALL)`. **]**
+
+**SRS_CANCELLATION_TOKEN_04_004: [** `cancellation_token_create` shall set the initial state to be equal to `canceled` parameter. **]**
+
+**SRS_CANCELLATION_TOKEN_04_005: [** `cancellation_token_create` shall return a valid `THANDLE(CANCELLATION_TOKEN)` when successful. **]**
+
+## cancellation_token_is_canceled
+
+```c
+MOCKABLE_FUNCTION(, int, cancellation_token_is_canceled, THANDLE(CANCELLATION_TOKEN), cancellation_token, bool*, canceled);
```
mattdurak, line 87: >int [](http://example.com/codeflow?start=20&length=3)

I'm tempted to suggest this just returns bool. Simplifies the calling code slightly since this really can't fail unless the token is NULL. If a NULL is given, returning false makes sense (a NULL token is not canceled).

mattdurak: /consider of course

devdoc/cancellation_token_requirements.md
```diff
@@ -0,0 +1,138 @@
...
+**SRS_CANCELLATION_TOKEN_04_007: [** `cancellation_token_is_canceled` shall return a non-zero value if `canceled` is `NULL`. **]**
+
+**SRS_CANCELLATION_TOKEN_04_008: [** `cancellation_token_is_canceled` shall assign `true` to `*canceled` if the token has been canceled and `false` otherwise. **]**
+
+**SRS_CANCELLATION_TOKEN_04_009: [** `cancellation_token_is_canceled` shall return `0` when successful. **]**
+
+## cancellation_token_register_notify
+
+```c
+MOCKABLE_FUNCTION(, THANDLE(CANCELLATION_TOKEN_REGISTRATION), cancellation_token_register_notify, THANDLE(CANCELLATION_TOKEN), cancellation_token, TCALL_DISPATCHER_TARGET_FUNC_TYPE_NAME(CANCELLATION_TOKEN_CANCEL_CALL), on_cancel, void*, context);
```
mattdurak, line 103: >CANCELLATION_TOKEN_REGISTRATION [](http://example.com/codeflow?start=28&length=31)

How do I unregister? Just by the reference count going to 0? If so maybe the spec for the "dispose" function on that thandle is missing here to say it does that.

devdoc/cancellation_token_requirements.md
```diff
@@ -0,0 +1,138 @@
+# Cancellation Token Requirements
+
+## Overview
+
+The cancellation token module implements an object that can be used to propagate cancellation signals through multi-threaded programs. The design is roughly equivalent to the [`CancellationToken`](https://learn.microsoft.com/en-us/dotnet/api/system.threading.cancellationtoken?view=net-8.0) type in .NET. The module allows applications to create cancellation token handles which can then be passed around to various functions. The following operations can be performed on cancellation tokens:
```
mattdurak, line 5: >The design is roughly equivalent to the [`CancellationToken`](https://learn.microsoft.com/en-us/dotnet/api/system.threading.cancellationtoken?view=net-8.0) type in .NET. [](http://example.com/codeflow?start=135&length=169)

I know you say "roughly" but there is one point that may be interesting. The .NET `Register` function says: "If this token is already in the canceled state, the delegate will be run immediately and synchronously." It looks like we differ there.

devdoc/cancellation_token_requirements.md
```diff
@@ -0,0 +1,138 @@
...
+
+**SRS_CANCELLATION_TOKEN_04_010: [** `cancellation_token_register_notify` shall fail and return `NULL` when `cancellation_token` is `NULL`. **]**
+
+**SRS_CANCELLATION_TOKEN_04_011: [** `cancellation_token_register_notify` shall fail and return `NULL` when `on_cancel` is `NULL`. **]**
+
+**SRS_CANCELLATION_TOKEN_04_012: [** `cancellation_token_register_notify` shall fail and return `NULL` when any underlying call fails. **]**
+
+**SRS_CANCELLATION_TOKEN_04_013: [** `cancellation_token_register_notify` shall call `TCALL_DISPATCHER_REGISTER_TARGET(CANCELLATION_TOKEN_CANCEL_CALL)` to register `on_cancel` with the dispatcher. **]**
+
+**SRS_CANCELLATION_TOKEN_04_014: [** `cancellation_token_register_notify` shall call `THANDLE_MALLOC(CANCELLATION_TOKEN_REGISTRATION)` to allocate a `CANCELLATION_TOKEN_REGISTRATION` handle. **]**
```
anporumb, line 116: >CANCELLATION_TOKEN_REGISTRATION [](http://example.com/codeflow?start=101&length=31)

What happens if between 04_013 and 04_014 `cancellation_token_cancel` is called AND 04_014 fails?

I suspect that `on_cancel` will be called, but the function return is NULL which would be "weird".

Can we do without `TCALL_DISPATCHER_UNREGISTER_TARGET`? Maybe switching 04_013 and 04_14?

...

devdoc/cancellation_token_requirements.md
```diff
@@ -0,0 +1,138 @@
...
+
+**SRS_CANCELLATION_TOKEN_04_017: [** `cancellation_token_cancel` shall fail and return a non-zero value if `cancellation_token` is `NULL`. **]**
+
+**SRS_CANCELLATION_TOKEN_04_018: [** `cancellation_token_cancel` shall fail and return a non-zero value if the state of the token is already "canceled". **]**
+
+**SRS_CANCELLATION_TOKEN_04_019: [** `cancellation_token_cancel` shall set the state of the token to be "canceled". **]**
+
+**SRS_CANCELLATION_TOKEN_04_020: [** `cancellation_token_cancel` shall return `0` when successful. **]**
+
+**SRS_CANCELLATION_TOKEN_04_021: [** `cancellation_token_cancel` shall invoke all callbacks that have been registered on the token via `cancellation_token_register_notify`. **]**
```
anporumb, line 138: >invoke all callbacks [](http://example.com/codeflow?start=71&length=20)

It does that through TCALL_DISPATCHER means, right? Would be nice to have that explicitly written here.

src/cancellation_token.c
```diff
@@ -0,0 +1,212 @@
...
+#define CANCELLATION_TOKEN_STATE_VALUES \
+    CANCELLATION_TOKEN_STATE_NOT_CANCELED, \
+    CANCELLATION_TOKEN_STATE_CANCELED
+
+MU_DEFINE_ENUM(CANCELLATION_TOKEN_STATE, CANCELLATION_TOKEN_STATE_VALUES);
+MU_DEFINE_ENUM_STRINGS(CANCELLATION_TOKEN_STATE, CANCELLATION_TOKEN_STATE_VALUES);
+
+typedef struct CANCELLATION_TOKEN_TAG
+{
+    volatile_atomic int32_t state;
```
anporumb, line 25: >volatile_atomic int32_t state; [](http://example.com/codeflow?start=4&length=30)

consider putting this state in a union with the `CANCELLATION_TOKEN_STATE` enum, just because it is easier on the debug eyes.

src/cancellation_token.c
```diff
@@ -0,0 +1,212 @@
...
+            }
+
+            TCALL_DISPATCHER_INITIALIZE_MOVE(CANCELLATION_TOKEN_CANCEL_CALL)(&token_ptr->cancel_notification_call_dispatcher, &dispatcher);
+
+            /*Codes_SRS_CANCELLATION_TOKEN_04_005: [ cancellation_token_create shall return a valid THANDLE(CANCELLATION_TOKEN) when successful. ]*/
+            goto all_ok;
+        }
+
+        /*Codes_SRS_CANCELLATION_TOKEN_04_002: [ If any underlying error occurs cancellation_token_create shall fail and return NULL. ]*/
+        THANDLE_ASSIGN(CANCELLATION_TOKEN)(&token, NULL);
```
anporumb, line 91: > THANDLE_ASSIGN(CANCELLATION_TOKEN)(&token, NULL); [](http://example.com/codeflow?start=7&length=50)

what has been done by THANDLE_MALLOC it shall be undone by THANDLE_FREE...

src/cancellation_token.c
```diff
@@ -0,0 +1,212 @@
...
+    else
+    {
+        CANCELLATION_TOKEN* token_ptr = THANDLE_GET_T(CANCELLATION_TOKEN)(cancellation_token);
+
+        /*Codes_SRS_CANCELLATION_TOKEN_04_013: [ cancellation_token_register_notify shall call TCALL_DISPATCHER_REGISTER_TARGET(CANCELLATION_TOKEN_CANCEL_CALL) to register on_cancel with the dispatcher. ]*/
+        TCALL_DISPATCHER_TARGET_HANDLE(CANCELLATION_TOKEN_CANCEL_CALL) cancel_notification_target_handle = TCALL_DISPATCHER_REGISTER_TARGET(CANCELLATION_TOKEN_CANCEL_CALL)(token_ptr->cancel_notification_call_dispatcher, on_cancel, context);
+        if (cancel_notification_target_handle == NULL)
+        {
+            /*Codes_SRS_CANCELLATION_TOKEN_04_012: [ cancellation_token_register_notify shall fail and return NULL when any underlying call fails. ]*/
+            LogError("TCALL_DISPATCHER_REGISTER_TARGET(CANCELLATION_TOKEN_CANCEL_CALL)(token_ptr->cancel_notification_call_dispatcher, on_cancel, context) failed.");
```
anporumb, line 150: >token_ptr->cancel_notification_call_dispatcher, on_cancel, context [](http://example.com/codeflow?start=87&length=66)

parameters can be listed with their values.

src/cancellation_token.c
```diff
@@ -0,0 +1,212 @@
...
+            THANDLE(CANCELLATION_TOKEN_REGISTRATION) registration = THANDLE_MALLOC(CANCELLATION_TOKEN_REGISTRATION)(cancellation_token_registration_dispose);
+            if (registration == NULL)
+            {
+                /*Codes_SRS_CANCELLATION_TOKEN_04_012: [ cancellation_token_register_notify shall fail and return NULL when any underlying call fails. ]*/
+                LogError("THANDLE_MALLOC(CANCELLATION_TOKEN)(cancellation_token_dispose) failed.");
+            }
+            else
+            {
+                CANCELLATION_TOKEN_REGISTRATION* registration_ptr = THANDLE_GET_T(CANCELLATION_TOKEN_REGISTRATION)(registration);
+                registration_ptr->cancel_notification_target_handle = cancel_notification_target_handle;
```
anporumb, line 164: > registration_ptr->cancel_notification_target_handle = cancel_notification_target_handle; [](http://example.com/codeflow?start=15&length=89)

How did this compile?

TCALLs `should` be kind of THANDLEs, and the MAJOR advantage of THANDLE is that they are not assignable.

src/cancellation_token.c
```diff
@@ -0,0 +1,212 @@
...
+            /*Codes_SRS_CANCELLATION_TOKEN_04_014: [ cancellation_token_register_notify shall call THANDLE_MALLOC(CANCELLATION_TOKEN_REGISTRATION) to allocate a CANCELLATION_TOKEN_REGISTRATION handle. ]*/
+            THANDLE(CANCELLATION_TOKEN_REGISTRATION) registration = THANDLE_MALLOC(CANCELLATION_TOKEN_REGISTRATION)(cancellation_token_registration_dispose);
+            if (registration == NULL)
+            {
+                /*Codes_SRS_CANCELLATION_TOKEN_04_012: [ cancellation_token_register_notify shall fail and return NULL when any underlying call fails. ]*/
+                LogError("THANDLE_MALLOC(CANCELLATION_TOKEN)(cancellation_token_dispose) failed.");
+            }
+            else
+            {
+                CANCELLATION_TOKEN_REGISTRATION* registration_ptr = THANDLE_GET_T(CANCELLATION_TOKEN_REGISTRATION)(registration);
```
anporumb, line 163: >THANDLE_GET_T [](http://example.com/codeflow?start=68&length=13)

I love this correct usage here (and everywhere).

src/cancellation_token.c
```diff
@@ -0,0 +1,212 @@
...
+        result = MU_FAILURE;
+    }
+    else
+    {
+        CANCELLATION_TOKEN* token_ptr = THANDLE_GET_T(CANCELLATION_TOKEN)(cancellation_token);
+
+        /*Codes_SRS_CANCELLATION_TOKEN_04_019: [ cancellation_token_cancel shall set the state of the token to be "canceled". ]*/
+        if (interlocked_compare_exchange(&token_ptr->state, CANCELLATION_TOKEN_STATE_CANCELED, CANCELLATION_TOKEN_STATE_NOT_CANCELED) != CANCELLATION_TOKEN_STATE_NOT_CANCELED)
+        {
+            /*Codes_SRS_CANCELLATION_TOKEN_04_018: [ cancellation_token_cancel shall fail and return a non-zero value if the state of the token is already "canceled". ]*/
```
anporumb, line 198: >/*Codes_SRS_CANCELLATION_TOKEN_04_018: [ cancellation_token_cancel shall fail and return a non-zero value if the state of the token is already "canceled". ]*/ [](http://example.com/codeflow?start=12&length=158)

could log an error here...

anporumb (dismissed):
:shipit:

devdoc/cancellation_token_requirements.md
avranju: Good point. Added the following:

----

### How does this relate to `async_op`?

The `c-util` repo includes a module called [async_op](async_op.md) which provides the following abilities:

- Provide some space on the heap for storing arbitrary data as it pertains to an async operation.
- Allow the calling of a cancel function via the `async_op_cancel` API on the `async_op` handle.

There is some overlap in functionality offered by `async_op` and cancellation tokens. Both can be used to implement cancellation on asynchronous functions. `async_op` can however be used to have only one cancellation callback associated with it. Consider a system for example where an incoming call is serviced by multiple software layers potentially involving multiple I/O calls. In such a system, it might be desirable to have a single cancellation signal trigger a collection of callbacks in various parts of the system that take separate action to implement cancellation. While it is technically feasible to implement such a system with `async_op`, it will involve fairly intricate state management and likely cause tight coupling between the part of the code that does the signalling with the part of the code that implements cancellation.

With cancellation tokens, the cite where the signal is raised does not have any knowledge of what action is taken at the place where the signal is handled. Further, unlike `async_op`, cancellation tokens also support the ability to imperatively check if the signal has been raised and take reactive action as can be necessary in situations where the work being cancelled is CPU-bound instead of being I/O bound.

devdoc/cancellation_token_requirements.md
avranju: Good catch! Perhaps we can do the same thing.

avranju: Added requirement `SRS_CANCELLATION_TOKEN_04_022` and implemented it:

> If the cancellation token is already in canceled state, then `cancellation_token_register_notify` shall immediately call `on_cancel` and take no further action and return `NULL`.

devdoc/cancellation_token_requirements.md
avranju: Done.

devdoc/cancellation_token_requirements.md
avranju: Good catch. Yes, it is by having the ref count fall to `0` that one un-registers the callback. Added the requirements for both `cancellation_token_registration_dispose` and `cancellation_token_dispose` and tests.

devdoc/cancellation_token_requirements.md
avranju: Good catch. Should have thought of that. Matt noted the same thing. Will fix.

avranju: Fixed.

devdoc/cancellation_token_requirements.md
avranju: Good catch! Andrei noticed the same thing. Fixed it.

src/cancellation_token.c
avranju: Took a shot at this. Hope this is what you meant!

```c
typedef union CANCELLATION_TOKEN_STATE_VALUE_TAG
{
    volatile_atomic int32_t state;
    volatile_atomic CANCELLATION_TOKEN_STATE state_enum;
} CANCELLATION_TOKEN_STATE_VALUE;

typedef struct CANCELLATION_TOKEN_TAG
{
    CANCELLATION_TOKEN_STATE_VALUE state;
    TCALL_DISPATCHER(CANCELLATION_TOKEN_CANCEL_CALL) cancel_notification_call_dispatcher;
} CANCELLATION_TOKEN;
```

src/cancellation_token.c
avranju: In this case `token` needs to be set to `NULL` before exiting the function. If I call `THANDLE_FREE` then `token` continues to be non-`NULL`. I see that `THANDLE_ASSIGN` does a `DEC_REF` when the RHS is `NULL`. So the freeing still happens as far as I can tell.

avranju: Ok. Added another `result` local variable and that handles it.

src/cancellation_token.c
avranju: Sure.

src/cancellation_token.c
avranju: TBH I didn't know there was another way!

src/cancellation_token.c
avranju: Not sure. Looks like this is _not_ a `THANDLE`? `cancel_notification_target_handle` is a `TCALL_DISPATCHER_TARGET_HANDLE(CANCELLATION_TOKEN_CANCEL_CALL)` which expands to `TCALL_DISPATCHER_TARGET_HANDLE_CANCELLATION_TOKEN_CANCEL_CALL` which, according to VS intellisense, is this:

image

And that looks like a regular old pointer to me.

avranju: I shall leave this be since it doesn't seem to be a `THANDLE` to me. Further, in `TCALL_DISPATCHER_LL_UNREGISTER_TARGET` in `tcall_dispatcher_ll.h`, I see that it treats `call_target` as a regular pointer and calls `free` on it.

parth21999 (approved):

parth21999 (approved):
:shipit:

PR #350, opened 2024-02-05
Use `real_interlocked` in `real_channel`

parth21999:
If a test is mocking `interlocked.h` and using `real_channel.h`, `THANDLE` functions cause `interlocked` calls to get recorded.


anporumb (approved):
:shipit:

PR #352, opened 2024-02-06
add_2d_array_code_and_ut will add int test next pr

M-iceberg:


tests/two_d_array_ut/two_d_array_ut.c
```diff
@@ -0,0 +1,642 @@
...
+
+TEST_FUNCTION_CLEANUP(cleans)
+{
+    umock_c_negative_tests_deinit();
+}
+
+/*TWO_D_ARRAY_CREATE(T)*/
+
+/* Tests_SRS_TWO_D_ARRAY_07_003: [ TWO_D_ARRAY_CREATE(T) shall call THANDLE_MALLOC_FLEX with TWO_D_ARRAY_FREE(T) as dispose function, nmemb set to row_size and size set to sizeof(T*). ]*/
+/* Tests_SRS_TWO_D_ARRAY_07_004: [ TWO_D_ARRAY_CREATE(T) shall set all rows point to NULL. ]*/
```
dcristoloveanu, line 90: >rows point [](http://example.com/codeflow?start=71&length=10)

row pointers?

tests/two_d_array_ut/two_d_array_ut.c
```diff
@@ -0,0 +1,642 @@
...
+{
+    //arrange
+    STRICT_EXPECTED_CALL(malloc_flex(IGNORED_ARG, 5, sizeof(uint32_t*)));
+
+    //act
+    TWO_D_ARRAY(uint32_t) tdarr = TWO_D_ARRAY_CREATE(uint32_t)(5, 5);
+
+    //assert
+    ASSERT_IS_NOT_NULL(tdarr);
+    for (unsigned int i = 0; i < 5; i++)
```
dcristoloveanu, line 102: >unsigned int [](http://example.com/codeflow?start=9&length=12)

We usually use uint32_t (the fixed size types)

tests/two_d_array_ut/two_d_array_ut.c
```diff
@@ -0,0 +1,642 @@
...
+/* Tests_SRS_TWO_D_ARRAY_07_003: [ TWO_D_ARRAY_CREATE(T) shall call THANDLE_MALLOC_FLEX with TWO_D_ARRAY_FREE(T) as dispose function, nmemb set to row_size and size set to sizeof(T*). ]*/
+/* Tests_SRS_TWO_D_ARRAY_07_004: [ TWO_D_ARRAY_CREATE(T) shall set all rows point to NULL. ]*/
+/* Tests_SRS_TWO_D_ARRAY_07_006: [ TWO_D_ARRAY_CREATE(T) shall succeed and return a non-NULL value. ]*/
+TEST_FUNCTION(TWO_D_ARRAY_CREATE_create_uint32_t_succeeds)
+{
+    //arrange
+    STRICT_EXPECTED_CALL(malloc_flex(IGNORED_ARG, 5, sizeof(uint32_t*)));
+
+    //act
+    TWO_D_ARRAY(uint32_t) tdarr = TWO_D_ARRAY_CREATE(uint32_t)(5, 5);
```
dcristoloveanu, line 98: >5, 5 [](http://example.com/codeflow?start=63&length=4)

We usually test at the boundaries (inflexion points in the range of values).

Imagine that the range of values for rows is 0 .. UINT32_MAX.
These inflexion points are the points where we expect behavior to be different.

For example I would test what happens when called with 0 - we'd expact that to fail, right?
Also we should test that when called with 1 for rows a 2D array is allocated.

from 1 to UINT32_MAX-1 the expected behavior is the same, so it adds very little value to test for 2 and 3 and 4. So a best practice is to pick something in the interval and test to see the behavior is kept (5 is Ok to test with for example).

So, we should have tests for row values:
0, 1, 5, UINT32_MAX - 1 and UINT32_MAX).

tests/two_d_array_ut/two_d_array_ut.c
```diff
@@ -0,0 +1,642 @@
...
+    ASSERT_ARE_EQUAL(char_ptr, umock_c_get_expected_calls(), umock_c_get_actual_calls());
+
+    //clean
+    TWO_D_ARRAY_ASSIGN(uint32_t)(&tdarr, NULL);
+}
+
+/* Tests_SRS_TWO_D_ARRAY_07_003: [ TWO_D_ARRAY_CREATE(T) shall call THANDLE_MALLOC_FLEX with TWO_D_ARRAY_FREE(T) as dispose function, nmemb set to row_size and size set to sizeof(T*). ]*/
+/* Tests_SRS_TWO_D_ARRAY_07_004: [ TWO_D_ARRAY_CREATE(T) shall set all rows point to NULL. ]*/
+/* Tests_SRS_TWO_D_ARRAY_07_006: [ TWO_D_ARRAY_CREATE(T) shall succeed and return a non-NULL value. ]*/
+TEST_FUNCTION(TWO_D_ARRAY_CREATE_create_THANDLE_succeeds)
```
dcristoloveanu, line 134: >_THANDLE_ [](http://example.com/codeflow?start=39&length=9)

Consider being more explicit in the test name. Like 
TWO_D_ARRAY_CREATE_create_with_THANDLE_type_succeeds

dcristoloveanu, line 134: >TWO_D_ARRAY_CREATE_create_THANDLE_succeeds [](http://example.com/codeflow?start=14&length=42)

Nice test

dcristoloveanu, line 134: >TWO_D_ARRAY_CREATE_create_THANDLE_succeeds [](http://example.com/codeflow?start=14&length=42)

Do we have a test with a struct type?

tests/two_d_array_ut/two_d_array_ut.c
```diff
@@ -0,0 +1,642 @@
...
+
+    //act
+    TWO_D_ARRAY(uint32_t) tdarr = TWO_D_ARRAY_CREATE(uint32_t)(5, 5);
+
+    //assert
+    ASSERT_IS_NULL(tdarr);
+    ASSERT_ARE_EQUAL(char_ptr, umock_c_get_expected_calls(), umock_c_get_actual_calls());
+
+    //clean
+}
```
dcristoloveanu, line 188: Consider using umock_negative_tests
We usually use that as a general pattern in all the when_underlying_calls_fails_XXX_api_fails

tests/two_d_array_ut/two_d_array_ut.c
```diff
@@ -0,0 +1,642 @@
...
+    TWO_D_ARRAY(uint32_t) tdarr = TWO_D_ARRAY_CREATE(uint32_t)(5, 5000000);
+    umock_c_reset_all_calls();
+    STRICT_EXPECTED_CALL(malloc_2(5000000, sizeof(uint32_t)));
+
+    //act
+    result = TWO_D_ARRAY_ALLOCATE_NEW_ROW(uint32_t)(tdarr, 0);
+
+    //assert
+    ASSERT_IS_NOT_NULL(tdarr);
+    ASSERT_ARE_EQUAL(int, result, 0);
```
dcristoloveanu, line 236: This assert should be between ln 227 and 228.
We should not proceed with the test if we were unable to allocate the array in the first place

tests/two_d_array_ut/two_d_array_ut.c
```diff
@@ -0,0 +1,642 @@
...
+    ASSERT_IS_NOT_NULL(tdarr);
+    ASSERT_ARE_NOT_EQUAL(int, result, 0);
+    ASSERT_ARE_EQUAL(char_ptr, umock_c_get_expected_calls(), umock_c_get_actual_calls());
+
+    //clean
+    TWO_D_ARRAY_ASSIGN(uint32_t)(&tdarr, NULL);
+}
+
+/* Tests_SRS_TWO_D_ARRAY_07_016: [ If the row specified by row_index has already been allocated, TWO_D_ARRAY_ALLOCATE_NEW_ROW(T) shall fail and return a non-zero value. ]*/
+TEST_FUNCTION(TWO_D_ARRAY_ALLOCATE_NEW_ROW_fails_when_that_row_has_already_been_added)
```
dcristoloveanu, line 338: >added [](http://example.com/codeflow?start=80&length=5)

I think add is misleading, allocated is the correct term me thinks.
add usually implies that anything existing stays the same and something is added to the collection.

dcristoloveanu, line 338: >TWO_D_ARRAY_ALLOCATE_NEW_ROW_fails_when_that_row_has_already_been_added [](http://example.com/codeflow?start=14&length=71)

Is there a test that verifies all rows can be allocated?

tests/two_d_array_ut/two_d_array_ut.c
```diff
@@ -0,0 +1,642 @@
...
+}
+
+/* Tests_SRS_TWO_D_ARRAY_07_011: [ If row_index is equal or greater than row_size, TWO_D_ARRAY_FREE_ROW(T) shall fail and return a non-zero value. ]*/
+TEST_FUNCTION(TWO_D_ARRAY_FREE_ROW_fails_when_row_index_exceeds)
+{
+    //arrange
+    int result;
+    TWO_D_ARRAY(uint32_t) tdarr = TWO_D_ARRAY_CREATE(uint32_t)(5, 5);
+    int add_row_result = TWO_D_ARRAY_ALLOCATE_NEW_ROW(uint32_t)(tdarr, 0);
+    ASSERT_IS_NOT_NULL(tdarr);
```
dcristoloveanu, line 474: This assert has to be before we try to use the tdarr
Here and everywhere else where applicable.

tests/two_d_array_ut/two_d_array_ut.c
```diff
@@ -0,0 +1,642 @@
...
+
+    //assert
+    ASSERT_ARE_NOT_EQUAL(int, result, 0);
+    ASSERT_ARE_EQUAL(char_ptr, umock_c_get_expected_calls(), umock_c_get_actual_calls());
+
+    //clean
+}
+
+/* Tests_SRS_TWO_D_ARRAY_07_011: [ If row_index is equal or greater than row_size, TWO_D_ARRAY_FREE_ROW(T) shall fail and return a non-zero value. ]*/
+TEST_FUNCTION(TWO_D_ARRAY_FREE_ROW_fails_when_row_index_exceeds)
```
dcristoloveanu, line 468: >exceeds [](http://example.com/codeflow?start=56&length=7)

Consider saying too_high

tests/two_d_array_ut/two_d_array_ut.c
```diff
@@ -0,0 +1,642 @@
...
+    uint32_t* row_result;
+    TWO_D_ARRAY(uint32_t) tdarr = TWO_D_ARRAY_CREATE(uint32_t)(5, 5);
+    int add_row_result = TWO_D_ARRAY_ALLOCATE_NEW_ROW(uint32_t)(tdarr, 0);
+    ASSERT_IS_NOT_NULL(tdarr);
+    ASSERT_ARE_EQUAL(int, add_row_result, 0);
+    ASSERT_IS_NOT_NULL(tdarr->row_arrays[0]);
+    umock_c_reset_all_calls();
+
+    //act
+    row_result = TWO_D_ARRAY_GET(uint32_t)(tdarr, 1);
```
dcristoloveanu, line 598: >GET [](http://example.com/codeflow?start=29&length=3)

This is really get_row. We should name it like that

dcristoloveanu (changes requested):
:clock1:

tests/two_d_array_ut/two_d_array_ut.c
dcristoloveanu: I guess this would be UINT64_MAX, so you won't actually be able to allocate that much memory, but you can trick the test to not allocate anything, but rather just return a dummy pointer.

dcristoloveanu: The beauty of hooks

inc/c_util/two_d_array_ll.h
```diff
@@ -15,10 +15,236 @@
...
+        /* Codes_SRS_TWO_D_ARRAY_07_003: [ TWO_D_ARRAY_CREATE(T) shall call THANDLE_MALLOC_FLEX with TWO_D_ARRAY_FREE(T) as dispose function, nmemb set to row_size and size set to sizeof(T*). ]*/\
+        result = THANDLE_MALLOC_FLEX(TWO_D_ARRAY_TYPEDEF_NAME(C))(TWO_D_ARRAY_LL_FREE_NAME(C), row_size, sizeof(T*));                                                                              \
+        /* Codes_SRS_TWO_D_ARRAY_07_005: [ If there are any errors then TWO_D_ARRAY_CREATE(T) shall fail and return NULL. ]*/                                                                      \
+        if (result == NULL)                                                                                                                                                                        \
+        {                                                                                                                                                                                          \
+            LogError("failure in malloc_flex(sizeof(TWO_D_ARRAY_HANDLE)=%zu, row_size=%" PRIu64 ", sizeof(unsigned char*)=%zu);", sizeof(T*), row_size, sizeof(unsigned char*));                   \
```
dcristoloveanu, line 118: > sizeof(unsigned char*) [](http://example.com/codeflow?start=98&length=23)

This should be sizeof(T*)

dcristoloveanu: And we should print the T with MU_TO_STRING) so that the log line has information about what we're working with.

inc/c_util/two_d_array_ll.h
```diff
@@ -15,10 +15,236 @@
...
+        }                                                                                                                                                           \
+        else                                                                                                                                                        \
+        {                                                                                                                                                           \
+            /* Codes_SRS_TWO_D_ARRAY_07_012: [ TWO_D_ARRAY_FREE_ROW(T) shall free the memory associated with the row specified by row_index and set it to NULL. ]*/ \
+            void* nullptr = NULL;                                                                                                                                   \
+            free((void*)two_d_array->row_arrays[row_index]);                                                                                                        \
+            (void)memcpy((void*)&two_d_array->row_arrays[row_index], &nullptr, sizeof(void*));                                                                      \
```
dcristoloveanu, line 160: >(void)memcpy((void*)&two_d_array->row_arrays[row_index], &nullptr, sizeof(void*));                                                                      \ [](http://example.com/codeflow?start=12&length=153)

This is an interesting trick. Even for me :-). Why is this needed?

inc/c_util/two_d_array_ll.h
```diff
@@ -15,10 +15,236 @@
...
+                    LogError("malloc(sizeof(col)=%zu) failed", two_d_array->cols * sizeof(T));                                                                                               \
+                    result = MU_FAILURE;                                                                                                                                                     \
+                }                                                                                                                                                                            \
+                else                                                                                                                                                                         \
+                {                                                                                                                                                                            \
+                    (void)memcpy((void*)&two_d_array->row_arrays[row_index], &new_row, sizeof(T*));                                                                                          \
```
dcristoloveanu, line 201: >(void)memcpy((void*)&two_d_array->row_arrays[row_index], &new_row, sizeof(T*));  [](http://example.com/codeflow?start=20&length=80)

I don't think we need to call memcpy to set the row, do we?

inc/c_util/two_d_array_ll.h
```diff
@@ -15,10 +15,236 @@
...
+            }                                                                                                                                                                                \
+            else                                                                                                                                                                             \
+            {                                                                                                                                                                                \
+                /* Codes_SRS_TWO_D_ARRAY_07_016: [ If the row specified by row_index has already been allocated, TWO_D_ARRAY_ALLOCATE_NEW_ROW(T) shall fail and return a non-zero value. ]*/ \
+                result = MU_FAILURE;                                                                                                                                                         \
+            }                                                                                                                                                                                \
```
dcristoloveanu, line 209: If this is really an error we should LogError here.
Also we usually put the error first, so consider reversing the logic in 189 to check for != NULL, log the error on the true branch and deal with the allocation on the false branch of the if.

inc/c_util/two_d_array_ll.h
```diff
@@ -15,10 +15,236 @@
...
+            if(result == NULL)                                                                                                                             \
+            {                                                                                                                                              \
+            /* Codes_SRS_TWO_D_ARRAY_07_021: [ If the array stored in row_index is NULL, TWO_D_ARRAY_GET(T) shall fail and return NULL. ]*/                \
+                LogError("The row trying to acquire is not allocated: TWO_D_ARRAY_HANDLE two_d_array=%p, row_index = %" PRIu64, two_d_array, row_index);   \
+            }                                                                                                                                              \
+            /* Codes_SRS_TWO_D_ARRAY_07_022: [ Otherwise, TWO_D_ARRAY_GET(T) shall return the entire column stored in the corresponding row_index. ]*/     \
+            else{                                                                                                                                          \
```
dcristoloveanu, line 240: >{ [](http://example.com/codeflow?start=16&length=1)

Nitpick:indentation.

inc/c_util/two_d_array_ll.h
```diff
@@ -15,10 +15,236 @@
...
+        }                                                                                                                                                  \
+        else                                                                                                                                               \
+        {                                                                                                                                                  \
+            result = two_d_array->row_arrays[row_index];                                                                                                   \
+            if(result == NULL)                                                                                                                             \
+            {                                                                                                                                              \
+            /* Codes_SRS_TWO_D_ARRAY_07_021: [ If the array stored in row_index is NULL, TWO_D_ARRAY_GET(T) shall fail and return NULL. ]*/                \
```
dcristoloveanu, line 236: >/*  [](http://example.com/codeflow?start=12&length=3)

Nitpick: add a tab here to indent the Codes like the line of code.

inc/c_util/two_d_array_ll.h
```diff
@@ -15,10 +15,236 @@
...
+T* TWO_D_ARRAY_LL_GET(C)(TWO_D_ARRAY_LL(T) two_d_array, uint64_t row_index)                                                                                \
+{                                                                                                                                                          \
+    T* result;                                                                                                                                             \
+    /* Codes_SRS_TWO_D_ARRAY_07_019: [ If two_d_array is NULL, TWO_D_ARRAY_GET(T) shall fail return NULL. ]*/                                              \
+    if (two_d_array == NULL)                                                                                                                               \
+    {                                                                                                                                                      \
+        LogError("Invalid arguments: TWO_D_ARRAY_HANDLE two_d_array=%p", two_d_array);                                                                     \
```
dcristoloveanu, line 222: >TWO_D_ARRAY_HANDLE  [](http://example.com/codeflow?start=37&length=19)

All the loglines should printf 
"TWO_D_ARRAY(" MU_TOSTRING(T) ")"

so that we know what we're dealing with as structures rather than a generic TWO_D_ARRAY_HANDLE which does not even exist :-)

dcristoloveanu:
> #include <stdbool.h>

bool is not used, so likely not needed

---
Refers to: inc/c_util/two_d_array_ll.h:10 in ea9bd85. [](commit_id = ea9bd85a5dca269e8d2724629a6790cef1aa6757, deletion_comment = False)

tests/two_d_array_ut/two_d_array_ut.c
```diff
@@ -0,0 +1,642 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+
+#include <stdlib.h>
```
dcristoloveanu, line 5: Pretty sure you want to include inttypes also.

inc/c_util/two_d_array.h
```diff
@@ -10,15 +11,54 @@
 #endif
```
dcristoloveanu, line 12: This is not needed in this header, no usage of the PRI macros or the fixed size types.

M-iceberg: removed

M-iceberg:
> #include <stdbool.h>

removed

---
In reply to: [1931201352](https://github.com/Azure/c-util/pull/352#issuecomment-1931201352) [](http://example.com/codeflow?ancestors=1931201352)

---
Refers to: inc/c_util/two_d_array_ll.h:10 in ea9bd85. [](commit_id = ea9bd85a5dca269e8d2724629a6790cef1aa6757, deletion_comment = False)

inc/c_util/two_d_array_ll.h
M-iceberg: fixed

inc/c_util/two_d_array_ll.h
M-iceberg: fixed

inc/c_util/two_d_array_ll.h
M-iceberg: done

inc/c_util/two_d_array_ll.h
M-iceberg: added the logError and changed the order

inc/c_util/two_d_array_ll.h
M-iceberg: changed to use MU_TOSTRING here and everywhere else

inc/c_util/two_d_array_ll.h
M-iceberg: removed memcpy

inc/c_util/two_d_array_ll.h
M-iceberg: removed memcpy

tests/two_d_array_ut/two_d_array_ut.c
M-iceberg: modified

tests/two_d_array_ut/two_d_array_ut.c
M-iceberg: changed to use uint32_t here and everywhere else

tests/two_d_array_ut/two_d_array_ut.c
M-iceberg: added

tests/two_d_array_ut/two_d_array_ut.c
M-iceberg: changed to get_row here and everywhere else

tests/two_d_array_ut/two_d_array_ut.c
M-iceberg: done

tests/two_d_array_ut/two_d_array_ut.c
M-iceberg: changed to use umock_c_negative_test pattern

tests/two_d_array_ut/two_d_array_ut.c
M-iceberg: modified the order here and everywhere else

tests/two_d_array_ut/two_d_array_ut.c
M-iceberg: moved line 235 to between line 227 and 228

tests/two_d_array_ut/two_d_array_ut.c
M-iceberg: modified

M-iceberg: added the test with struct type

tests/two_d_array_ut/two_d_array_ut.c
M-iceberg: TWO_D_ARRAY_ALLOCATE_NEW_ROW_with_uint32_t_type_all_rows_can_be_allocated is testing if all the rows can be allocated in line 310 to line 363

M-iceberg: changed to allocated

tests/two_d_array_ut/two_d_array_ut.c
M-iceberg: Thank you so much for explaining the details. I tried to test UINT64_MAX before and that much memory can't be allocated. I try to return a dummy pointer now, but in the two_d_array_ll.h line 125-128, it try to access the returned address by malloc and set every row to NULL, which will give a access violation when I set return a dummy pointer in malloc_flex. I'm confused about whether UINT64_MAX could be tested here.

inc/c_util/two_d_array_ll.h
```diff
@@ -7,18 +7,248 @@
...
+                T* new_row = malloc_2(two_d_array->cols, sizeof(T));                                                                                                                         \
+                if (new_row == NULL)                                                                                                                                                         \
+                {                                                                                                                                                                            \
+                    /* Codes_SRS_TWO_D_ARRAY_07_018: [ If there are any errors then TWO_D_ARRAY_ALLOCATE_NEW_ROW(T) shall fail and return a non-zero value. ]*/                              \
+                    LogError("failure in malloc_2(number of cols=%" PRIu64 ", sizeof(" MU_TOSTRING(T) ")=%zu)",                                                                              \
+                    two_d_array->cols, sizeof(T));                                                                                                                                           \
```
dcristoloveanu, line 202: >two_d_array->cols, sizeof(T));     [](http://example.com/codeflow?start=20&length=34)

This should be indented, since it belongs to the line above.

inc/c_util/two_d_array_ll.h
```diff
@@ -7,18 +7,248 @@
...
+        {                                                                                                                                                  \
+            result = two_d_array->row_arrays[row_index];                                                                                                   \
+            if(result == NULL)                                                                                                                             \
+            {                                                                                                                                              \
+                /* Codes_SRS_TWO_D_ARRAY_07_021: [ If the array stored in row_index is NULL, TWO_D_ARRAY_GET_ROW(T) shall fail and return NULL. ]*/            \
+                LogError("The row trying to acquire is not allocated: TWO_D_ARRAY (" MU_TOSTRING(T) ") two_d_array=%p, row_index = %" PRIu64,              \
+                two_d_array, row_index);                                                                                                                   \
```
dcristoloveanu, line 240: >two_d_array, row_index);      [](http://example.com/codeflow?start=16&length=29)

Same here, indentation needed

tests/two_d_array_ut/two_d_array_ut.c
dcristoloveanu: Yes, there is no computer with that kind of memory.
For mocking, yes I was thinking that you can return a dummy value, but looks like because the constructor is initializing the memory that won't be doable (without extracting the row access and more mocking it, which would be rather unnecessary).

So we won't write any test with UINT64_MAX I assume.
How about a test with UINT32_MAX + 1, do we test that?

dcristoloveanu: Do we really need uint64_t? I don't think we expect having 4B rows or 4B columns.
Frankly uint32_t should be enough IMO. Then you can also test the extremeties of the interval.

inc/c_util/two_d_array.h
```diff
@@ -1,24 +1,58 @@
...
 
-/*macro to be used in headers*/                                                                                     \
+/*macro to be used in headers*/                                                                                       \
 #define TWO_D_ARRAY_TYPE_DECLARE(T)                                                                                   \
+    /*hint: have TWO_D_ARRAY_TYPE_DEFINE(T) before TWO_D_ARRAY_TYPE_DECLARE                                        */ \
+    /*hint: have THANDLE_TYPE_DECLARE(TWO_D_ARRAY_TYPEDEF_NAME(T)) before TWO_D_ARRAY_TYPE_DECLARE                 */ \
+    TWO_D_ARRAY_CREATE_DECLARE(T)                                                                                     \
+    TWO_D_ARRAY_FREE_ROW_DECLARE(T)                                                                                   \
+    TWO_D_ARRAY_ALLOCATE_NEW_ROW_DECLARE(T)                                                                           \
+    TWO_D_ARRAY_GET_ROW_DECLARE(T)                                                                                        \
```
dcristoloveanu, line 49: Let's tidy up the \ to have them all nicely aligned

tests/two_d_array_ut/two_d_array_ut.c
M-iceberg: done

inc/c_util/two_d_array_ll.h
M-iceberg: done

inc/c_util/two_d_array_ll.h
M-iceberg: done

inc/c_util/two_d_array.h
M-iceberg: done

tests/two_d_array_ut/two_d_array_ut.c
```diff
@@ -0,0 +1,811 @@
...
+
+    //clean
+    TWO_D_ARRAY_ASSIGN(uint32_t)(&tdarr, NULL);
+}
+*/
+
+/* Tests_SRS_TWO_D_ARRAY_07_003: [ TWO_D_ARRAY_CREATE(T) shall call THANDLE_MALLOC_FLEX with TWO_D_ARRAY_FREE(T) as dispose function, nmemb set to row_size and size set to sizeof(T*). ]*/
+/* Tests_SRS_TWO_D_ARRAY_07_004: [ TWO_D_ARRAY_CREATE(T) shall set all rows pointers to NULL. ]*/
+/* Tests_SRS_TWO_D_ARRAY_07_006: [ TWO_D_ARRAY_CREATE(T) shall succeed and return a non-NULL value. ]*/
+/* running too long in the gate, comment out
```
dcristoloveanu, line 169: >comment [](http://example.com/codeflow?start=33&length=7)

Nitpick:commented

tests/two_d_array_ut/two_d_array_ut.c
```diff
@@ -0,0 +1,811 @@
...
+    ASSERT_ARE_EQUAL(char_ptr, umock_c_get_expected_calls(), umock_c_get_actual_calls());
+
+    //clean
+    TWO_D_ARRAY_ASSIGN(uint32_t)(&tdarr, NULL);
+}
+
+/* Tests_SRS_TWO_D_ARRAY_07_003: [ TWO_D_ARRAY_CREATE(T) shall call THANDLE_MALLOC_FLEX with TWO_D_ARRAY_FREE(T) as dispose function, nmemb set to row_size and size set to sizeof(T*). ]*/
+/* Tests_SRS_TWO_D_ARRAY_07_004: [ TWO_D_ARRAY_CREATE(T) shall set all rows pointers to NULL. ]*/
+/* Tests_SRS_TWO_D_ARRAY_07_006: [ TWO_D_ARRAY_CREATE(T) shall succeed and return a non-NULL value. ]*/
+/* running too long in the gate, comment out
```
dcristoloveanu, line 148: >/* [](http://example.com/codeflow?start=0&length=2)

Can we have the tests disabled with #if 0 and #endif
It makes it much easier to enable if someone wants to run the tests - whoever wants to run them would change it to be #if 1.

dcristoloveanu (dismissed):
:shipit:

tests/two_d_array_ut/two_d_array_ut.c
M-iceberg: done

tests/two_d_array_ut/two_d_array_ut.c
M-iceberg: done

dcristoloveanu (approved):
:shipit:

PR #353, opened 2024-02-09
Add watchdog.h/c to common build.

darobs:
Move watchdog.c and watchdog.h out of WIN32 only into common files.


darobs:
Not read yet... need to add tests to build.

darobs:
> Not read yet... need to add tests to build.

Tests are added and run with caveats.

darobs:
> > Not read yet... need to add tests to build.
> 
> Tests are added and run with caveats.

All tests should pass.

devdoc/watchdog.md
```diff
@@ -59,7 +60,10 @@ Callback for the timer.
 
 **SRS_WATCHDOG_42_027: [** If `context` is `NULL` then `watchdog_expired_callback` shall terminate the process. **]**
 
-**SRS_WATCHDOG_42_021: [** If the state of the watchdog is `RUNNING` then `watchdog_expired_callback` shall call `callback` with the `context` and `message` from `watchdog_start`. **]**
+**SRS_WATCHDOG_45_005: [** If the state of the watchdog is `RUNNING` then **]**
+- **SRS_WATCHDOG_45_001: [** `watchdog_expired_callback` shall set the state to `EXPIRING` **]**
```
anporumb, line 64: >`EXPIRING` [](http://example.com/codeflow?start=80&length=10)

"if it is not STOP"?

anporumb:
>     volatile_atomic int32_t state; // WATCHDOG_STATE

consider adding here a "union" between this volatile_atomic and enum WATCHDOG_STATE.

It looks better in the debugger, this is the only reason.

---
Refers to: src/watchdog.c:30 in 7006c6c. [](commit_id = 7006c6c8f18c4a4b6f171fcb47dac81b7491e6da, deletion_comment = False)

src/watchdog.c
```diff
@@ -129,6 +135,8 @@ IMPLEMENT_MOCKABLE_FUNCTION(, void, watchdog_reset, WATCHDOG_HANDLE, watchdog)
     }
     else
     {
+        /* Codes_SRS_WATCHDOG_45_003: [ watchdog_reset shall wait until state is not EXPIRING. ]*/
+        (void)InterlockedHL_WaitForNotValue(&watchdog->state, WATCHDOG_EXPIRING, UINT32_MAX);
         /*Codes_SRS_WATCHDOG_42_032: [ watchdog_reset shall set the state of the watchdog to STOP. ]*/
         (void)interlocked_exchange(&watchdog->state, WATCHDOG_STOP);
```
mattdurak, line 141: seems like something that needs a compare_exchange too (in a loop), because it could go to EXPIRING state between lines 139 and 141

src/watchdog.c
```diff
@@ -43,14 +45,17 @@ static void watchdog_expired_callback(void* context)
     else
     {
         WATCHDOG_HANDLE handle = context;
-
-        WATCHDOG_STATE state = interlocked_add(&handle->state, 0);
+        /* Codes_SRS_WATCHDOG_45_005: [ If the state of the watchdog is RUNNING then ]*/
+          /* Codes_SRS_WATCHDOG_45_001: [ watchdog_expired_callback shall set the state to EXPIRING ]*/
```
anporumb, line 49: >/* [](http://example.com/codeflow?start=10&length=2)

align these comments to the one above it.

anporumb:
>             LogVerbose("Starting watchdog %p", result);

if we have some more context, we should log some more context.

Here, the context is THANDLE(RC_STRING) message. Would be nice to have it logged in every Log line

---
Refers to: src/watchdog.c:104 in 7006c6c. [](commit_id = 7006c6c8f18c4a4b6f171fcb47dac81b7491e6da, deletion_comment = False)

anporumb (approved):
:shipit:

mattdurak:
>     volatile_atomic int32_t state; // WATCHDOG_STATE

sounds like a nice macro we should add to c_pal :) (noting that because this is a pattern used in many places)

---
In reply to: [1942282217](https://github.com/Azure/c-util/pull/353#issuecomment-1942282217) [](http://example.com/codeflow?ancestors=1942282217)

---
Refers to: src/watchdog.c:30 in 7006c6c. [](commit_id = 7006c6c8f18c4a4b6f171fcb47dac81b7491e6da, deletion_comment = False)

src/watchdog.c
mattdurak: same below

mattdurak (dismissed):
:shipit:

devdoc/watchdog.md
```diff
@@ -59,7 +60,10 @@ Callback for the timer.
 
 **SRS_WATCHDOG_42_027: [** If `context` is `NULL` then `watchdog_expired_callback` shall terminate the process. **]**
 
-**SRS_WATCHDOG_42_021: [** If the state of the watchdog is `RUNNING` then `watchdog_expired_callback` shall call `callback` with the `context` and `message` from `watchdog_start`. **]**
+**SRS_WATCHDOG_45_005: [** If the state of the watchdog is `RUNNING` then **]**
+- **SRS_WATCHDOG_45_001: [** `watchdog_expired_callback` shall set the state to `EXPIRING` **]**
+- **SRS_WATCHDOG_42_021: [** `watchdog_expired_callback` shall call `callback` with the `context` and `message` from `watchdog_start`. **]**
+- **SRS_WATCHDOG_45_002: [** `watchdog_expired_callback` shall return the state to `RUNNING`. **]**
```
dcristoloveanu, line 66: >e to `RUNNING`.  [](http://example.com/codeflow?start=78&length=16)

This implies that we're back to the same state - in which we could have the watchdog_expired_callback called and all would be well - we would execute again the user callback for the watchdog.

Shouldn't this user callback only be executed once?

dcristoloveanu: In my mind it sounds like the state machine should have a separate state like USER_CALLBACK_CALLED

dcristoloveanu:
>     volatile_atomic int32_t state; // WATCHDOG_STATE

This could be a union with a pretty enum variable and an volatile int32_t for changing the state.

---
Refers to: src/watchdog.c:30 in 7006c6c. [](commit_id = 7006c6c8f18c4a4b6f171fcb47dac81b7491e6da, deletion_comment = False)

src/watchdog.c
```diff
@@ -43,14 +45,17 @@ static void watchdog_expired_callback(void* context)
     else
     {
         WATCHDOG_HANDLE handle = context;
-
-        WATCHDOG_STATE state = interlocked_add(&handle->state, 0);
+        /* Codes_SRS_WATCHDOG_45_005: [ If the state of the watchdog is RUNNING then ]*/
+          /* Codes_SRS_WATCHDOG_45_001: [ watchdog_expired_callback shall set the state to EXPIRING ]*/
+        WATCHDOG_STATE state = interlocked_compare_exchange(&handle->state, WATCHDOG_EXPIRING, WATCHDOG_RUNNING);
```
dcristoloveanu, line 50: >EXPIRING [](http://example.com/codeflow?start=85&length=8)

Nitpick: I think a better name for the state is CALLING_USER_CALLBACK

src/watchdog.c
dcristoloveanu, line 141: This is racy. 
Imagine a watchdog with state RUNNING.
thread T1 getting stopped between ln 139 and 141 (it passed the check that state is not EXPIRING.

Thread T2 executes the callback which changes the state to EXPIRING. Right after changing the state it gets switched out to T1.

T1 continues executing, setting the state to STOP, freeing stuff.

T2 resumes execution and winamps.

dcristoloveanu: I think this would have to be done with interlocked_compare_exchange

dcristoloveanu: The comment about the fact that we'd free memory is applicable to watchdog_stop.
But the state would be inconsistent also for watchdog_reset, just in different ways.

dcristoloveanu (dismissed):
:shipit:

src/watchdog.c
darobs: Based on these comments, I think I'll change watchdog to use sm. Changes to follow.

devdoc/watchdog.md
darobs: Switched to state management for... state management.

src/watchdog.c
darobs: I'm expecting SM to handle this for me.

src/watchdog.c
```diff
@@ -129,18 +145,27 @@ IMPLEMENT_MOCKABLE_FUNCTION(, void, watchdog_reset, WATCHDOG_HANDLE, watchdog)
...
+        /*Codes_SRS_WATCHDOG_45_011: [ watchdog_reset shall call sm_close_begin. ]*/
+        if (sm_close_begin(watchdog->state) != SM_EXEC_GRANTED)
+        {
+            LogError("Watchdog could not be closed for reset");
+        }
+
         /*Codes_SRS_WATCHDOG_42_033: [ watchdog_reset shall cancel the current timer by calling threadpool_timer_cancel. ]*/
         threadpool_timer_cancel(watchdog->timer);
+        /*Codes_SRS_WATCHDOG_45_012: [ watchdog_reset shall call sm_close_end. ]*/
+        sm_close_end(watchdog->state);
```
dcristoloveanu, line 159: >sm_close_end(watchdog->state); [](http://example.com/codeflow?start=8&length=30)

sm_close_end should not be called unless sm_close_begin was granted.

src/watchdog.c
```diff
@@ -129,18 +145,27 @@ IMPLEMENT_MOCKABLE_FUNCTION(, void, watchdog_reset, WATCHDOG_HANDLE, watchdog)
...
-
-        /*Codes_SRS_WATCHDOG_42_034: [ watchdog_reset shall set the state of the watchdog to RUNNING. ]*/
-        (void)interlocked_exchange(&watchdog->state, WATCHDOG_RUNNING);
+        /*Codes_SRS_WATCHDOG_45_013: [ watchdog_reset shall call sm_open_begin. ]*/
+        if (sm_open_begin(watchdog->state) != SM_EXEC_GRANTED)
+        {
+            LogError("Watchdog could not be reopened for reset");
+        }
+        /*Codes_SRS_WATCHDOG_45_014: [ watchdog_reset shall call sm_open_end. ]*/
+        sm_open_end(watchdog->state, true);
```
dcristoloveanu, line 168: >sm_open_end(watchdog->state, true) [](http://example.com/codeflow?start=8&length=34)

Same, I think we should only call open end if we were able to start it.

devdoc/watchdog.md
```diff
@@ -9,9 +9,10 @@ This may be used to detect code that is stuck in a long-running operation.
 
 Individual watchdog instances can be started with specific timeout times in milliseconds and a callback to call if the watchdog expires. The watchdog instances should be stopped by the caller when the task they are watching completes so that no callback is fired.
 
-Each time a watchdog timer instances is started, it has the following state:
- - `RUNNING` : The timer has started and if it expires it will call the callback.
- - `STOP` : The explicit stop was called and if the timer expires then nothing will happen.
+The watchdog timer uses state management to operate:
+- On start, it will move to the open state on success.
+- callbacks will only be called in the open state. The callback will call `sm_exec_begin` to ensure timer is ready and will call `sm_exec_end` upon completion.
+- When timer is reset or stopped, the state is transistioned to closed first to ensure any executing callback is complete.
```
dcristoloveanu, line 15: Nice that we explain this. I would suggest going even further a bit and explaining what happens to sm in stop and reset separately.

devdoc/watchdog.md
```diff
@@ -71,11 +82,15 @@ MOCKABLE_FUNCTION(, void, watchdog_reset, WATCHDOG_HANDLE, watchdog);
...
 
-**SRS_WATCHDOG_42_032: [** `watchdog_reset` shall set the state of the watchdog to `STOP`. **]**
+**SRS_WATCHDOG_45_011: [** `watchdog_reset` shall call `sm_close_begin`. **]**
 
 **SRS_WATCHDOG_42_033: [** `watchdog_reset` shall cancel the current timer by calling `threadpool_timer_cancel`. **]**
 
-**SRS_WATCHDOG_42_034: [** `watchdog_reset` shall set the state of the watchdog to `RUNNING`. **]**
+**SRS_WATCHDOG_45_012: [** `watchdog_reset` shall call `sm_close_end`. **]**
+
+**SRS_WATCHDOG_45_013: [** `watchdog_reset` shall call `sm_open_begin`. **]**
```
dcristoloveanu, line 91: >l `sm_open_beg [](http://example.com/codeflow?start=53&length=14)

What happens if sm_close_begin fails? Or sm_open_begin? We should spec that and test it.

dcristoloveanu:
I am waiting curiously for the next iteration. I suggest asking Matt to look at it also as I think he wrote initially the module.

devdoc/watchdog.md
darobs: Offline conversations with Matt suggest that if `close_begin_fails`, it's because it's already closed, and that's OK. It does implies that 2 calls to `watchdog_reset` might race - maybe I'll add a test for that.

src/watchdog.c
```diff
@@ -129,19 +145,33 @@ IMPLEMENT_MOCKABLE_FUNCTION(, void, watchdog_reset, WATCHDOG_HANDLE, watchdog)
...
+            LogError("Watchdog could not be reopened for reset");
+        }
+        else
+        {
+            /*Codes_SRS_WATCHDOG_45_014: [ watchdog_reset shall call sm_open_end if sm_open_begin succeeds. ]*/
+            sm_open_end(watchdog->state, true);
+        }
+        
         /*Codes_SRS_WATCHDOG_42_035: [ watchdog_reset shall restart the timer by calling threadpool_timer_restart with the original timeout_ms from the call to start. ]*/
         (void)threadpool_timer_restart(watchdog->timer, watchdog->timeout_ms, 0);
```
mattdurak, line 176: >threadpool_timer_restart [](http://example.com/codeflow?start=14&length=24)

If we were not able to do sm_open_begin, then we must be in a weird state...not sure we should bother restarting the timer in that case...

mattdurak (approved):
:shipit:

src/watchdog.c
darobs: Good point - done.

mattdurak (approved):
:shipit:

PR #355, opened 2024-02-12
add 2d_array_int_test

M-iceberg:


tests/two_d_array_int/two_d_array_int.c
```diff
@@ -0,0 +1,251 @@
...
+TEST_FUNCTION(TWO_D_ARRAY_create_and_allocate_succeeds)
+{
+    //arrange
+    TWO_D_ARRAY(TEST_THANDLE) tdarr = TWO_D_ARRAY_CREATE(TEST_THANDLE)(ROW_NUM, 5);
+    ASSERT_IS_NOT_NULL(tdarr);
+
+    //act
+    for (uint32_t i = 0; i < ROW_NUM; i++)
+    {
+        TWO_D_ARRAY_ALLOCATE_NEW_ROW(TEST_THANDLE)(tdarr, i);
```
mattdurak, line 83: >TWO_D_ARRAY_ALLOCATE_NEW_ROW [](http://example.com/codeflow?start=8&length=28)

should check the return and assert

tests/two_d_array_int/two_d_array_int.c
```diff
@@ -0,0 +1,251 @@
...
+    ASSERT_IS_NOT_NULL(tdarr);
+    for (uint32_t i = 0; i < ROW_NUM; i++)
+    {
+        TWO_D_ARRAY_ALLOCATE_NEW_ROW(TEST_THANDLE)(tdarr, i);
+    }
+
+    //act
+    for (uint32_t i = 0; i < ROW_NUM; i++)
+    {
+        TWO_D_ARRAY_FREE_ROW(TEST_THANDLE)(tdarr, i);
```
mattdurak, line 109: >TWO_D_ARRAY_FREE_ROW [](http://example.com/codeflow?start=8&length=20)

looks like this also returns an error code which should be checked and asserted if not success.

Here and everywhere

tests/two_d_array_int/two_d_array_int.c
mattdurak: Here and everywhere

tests/two_d_array_int/two_d_array_int.c
```diff
@@ -0,0 +1,251 @@
...
+    ASSERT_IS_NOT_NULL(tdarr);
+    ASSERT_ARE_EQUAL(int, add_row_result, 0);
+    ASSERT_IS_NOT_NULL(tdarr->row_arrays[0]);
+    for (unsigned int i = 0; i < 5; i++)
+    {
+        tdarr->row_arrays[0][i].a = i;
+    }
+
+    //act
+    row_result = TWO_D_ARRAY_GET_ROW(TEST_THANDLE)(tdarr, 0);
```
mattdurak, line 157: >row_result = TWO_D_ARRAY_GET_ROW(TEST_THANDLE)(tdarr, 0); [](http://example.com/codeflow?start=4&length=57)

I'm curious about the expected usage here. A few tests up we directly access `tdarr->row_arrays`, but now I see we have a function to do it (which seems to do the bounds checking. Should line 115 and others above be calling this instead?

Do we want to "hide" the `row_arrays` field from "users"?

tests/two_d_array_int/two_d_array_int.c
```diff
@@ -0,0 +1,251 @@
...
+        TWO_D_ARRAY_ALLOCATE_NEW_ROW(TEST_THANDLE)(tdarr, i);
+    }
+
+    //action
+    uint32_t run_time = 0;
+    while (run_time < N_CHAOS_API_CALLS)
+    {
+        run_time++;
+        // perform one of the several actions
+        CHAOS_TEST_ACTION action = (CHAOS_TEST_ACTION)((rand() * (MU_ENUM_VALUE_COUNT(CHAOS_TEST_ACTION_VALUES) - 1) / RAND_MAX) + 1);
```
mattdurak, line 225: >rand [](http://example.com/codeflow?start=56&length=4)

missing an `srand(time)` call in suite init

tests/two_d_array_int/two_d_array_int.c
```diff
@@ -0,0 +1,251 @@
...
+    for (uint32_t i = 0; i < ROW_NUM; i++)
+    {
+        TWO_D_ARRAY_ALLOCATE_NEW_ROW(TEST_THANDLE)(tdarr, i);
+    }
+
+    //action
+    uint32_t run_time = 0;
+    while (run_time < N_CHAOS_API_CALLS)
+    {
+        run_time++;
```
mattdurak, line 223: >run_time [](http://example.com/codeflow?start=8&length=8)

seems like this is `iteration_count` as `run_time` implies we are doing some time based test.

tests/two_d_array_int/two_d_array_int.c
```diff
@@ -0,0 +1,251 @@
...
+
+#define N_CHAOS_API_CALLS 5000
+TEST_FUNCTION(TWO_D_ARRAY_chaos)
+{
+    //arrange
+    TWO_D_ARRAY(TEST_THANDLE) tdarr = TWO_D_ARRAY_CREATE(TEST_THANDLE)(ROW_NUM, 5);
+    ASSERT_IS_NOT_NULL(tdarr);
+    for (uint32_t i = 0; i < ROW_NUM; i++)
+    {
+        TWO_D_ARRAY_ALLOCATE_NEW_ROW(TEST_THANDLE)(tdarr, i);
```
mattdurak, line 216: >TWO_D_ARRAY_ALLOCATE_NEW_ROW [](http://example.com/codeflow?start=8&length=28)

need to check return code

tests/two_d_array_int/two_d_array_int.c
```diff
@@ -0,0 +1,251 @@
...
+        CHAOS_TEST_ACTION action = (CHAOS_TEST_ACTION)((rand() * (MU_ENUM_VALUE_COUNT(CHAOS_TEST_ACTION_VALUES) - 1) / RAND_MAX) + 1);
+        uint32_t rand_row_index = rand() % ROW_NUM;
+        switch (action)
+        {
+        default:
+            LogError("Invalid action: %" PRI_MU_ENUM "", MU_ENUM_VALUE(CHAOS_TEST_ACTION, action));
+            break;
+        case CHAOS_TEST_ACTION_FREE_ROW:
+        {
+            TWO_D_ARRAY_FREE_ROW(TEST_THANDLE)(tdarr, rand_row_index);
```
mattdurak, line 234: >TWO_D_ARRAY_FREE_ROW [](http://example.com/codeflow?start=12&length=20)

for this one and below, if we are not checking the return code (because it could be anything), then we should have `(void)` before the function calls to make it explicit that we are ignoring the result.

mattdurak: I guess technically the test could do some basic book keeping and track whether each row exists or not (just with an array of bools). Then you could check the result of each call (allocate succeeds if it doesn't exist or fails if it does, etc.).

Maybe that would be better for a test like this (especially since there is no extra multi-threading complexity so tracking the state is fairly straight-forward).

tests/two_d_array_int/two_d_array_int.c
```diff
@@ -0,0 +1,251 @@
...
+        }
+        case CHAOS_TEST_ACTION_ALLOCATE_NEW_ROW:
+        {
+            TWO_D_ARRAY_ALLOCATE_NEW_ROW(TEST_THANDLE)(tdarr, rand_row_index);
```
mattdurak, line 238: >TWO_D_ARRAY_ALLOCATE_NEW_ROW [](http://example.com/codeflow?start=12&length=28)

Maybe after allocating we should fill in the array fields? Or GET_ROW could do that?

Just thinking that it would at least maybe crash if we didn't actually allocate the 5 element array.

tests/two_d_array_int/two_d_array_int.c
```diff
@@ -0,0 +1,251 @@
...
+{
+    //arrange
+    TWO_D_ARRAY(TEST_THANDLE) tdarr = TWO_D_ARRAY_CREATE(TEST_THANDLE)(ROW_NUM, 5);
+    ASSERT_IS_NOT_NULL(tdarr);
+    for (uint32_t i = 0; i < ROW_NUM; i++)
+    {
+        TWO_D_ARRAY_ALLOCATE_NEW_ROW(TEST_THANDLE)(tdarr, i);
+    }
+
+    //action
```
mattdurak, line 219: >action [](http://example.com/codeflow?start=6&length=6)

nit: act

tests/two_d_array_int/two_d_array_int.c
```diff
@@ -0,0 +1,251 @@
...
+            TWO_D_ARRAY_ALLOCATE_NEW_ROW(TEST_THANDLE)(tdarr, rand_row_index);
+        }
+        case CHAOS_TEST_ACTION_GET_ROW:
+        {
+            TWO_D_ARRAY_GET_ROW(TEST_THANDLE)(tdarr, rand_row_index);
+        }
+        }
+    }
+
+    //cleanup
```
mattdurak, line 247: missing "assert", but in this case, maybe we should just have a comment like:

```c
// assert
// Test verifies that nothing crashes
```

tests/two_d_array_int/two_d_array_int.c
```diff
@@ -0,0 +1,251 @@
...
+{
+    int64_t a;
+} TEST_THANDLE;
+
+TWO_D_ARRAY_DEFINE_STRUCT_TYPE(TEST_THANDLE);
+THANDLE_TYPE_DECLARE(TWO_D_ARRAY_TYPEDEF_NAME(TEST_THANDLE));
+THANDLE_TYPE_DEFINE(TWO_D_ARRAY_TYPEDEF_NAME(TEST_THANDLE));
+
+TWO_D_ARRAY_TYPE_DECLARE(TEST_THANDLE);
+TWO_D_ARRAY_TYPE_DEFINE(TEST_THANDLE);
```
mattdurak, line 30: >TWO_D_ARRAY_TYPE_DEFINE [](http://example.com/codeflow?start=0&length=23)

would be nice if this implicitly did `TWO_D_ARRAY_DEFINE_STRUCT_TYPE` so the code that uses 2D array just has `TWO_D_ARRAY_TYPE_DECLARE` in the header and `TWO_D_ARRAY_TYPE_DEFINE` in the c file.

mattdurak (dismissed):
:shipit:

tests/two_d_array_int/two_d_array_int.c
M-iceberg: done

tests/two_d_array_int/two_d_array_int.c
M-iceberg: done

tests/two_d_array_int/two_d_array_int.c
M-iceberg: yes, want to hide row_arrays, changed to use TWO_D_ARRAY_GET_ROW in other places

tests/two_d_array_int/two_d_array_int.c
M-iceberg: added the state for each row and the check for each call

tests/two_d_array_int/two_d_array_int.c
M-iceberg: modified

tests/two_d_array_int/two_d_array_int.c
M-iceberg: changed to randomly allocate the rows below, deleted this allocation

tests/two_d_array_int/two_d_array_int.c
M-iceberg: modified

tests/two_d_array_int/two_d_array_int.c
M-iceberg: added

tests/two_d_array_int/two_d_array_int.c
M-iceberg: GTE_ROW only return the row pointer, added fill in all array fields

tests/two_d_array_int/two_d_array_int.c
M-iceberg: I followed the structure pattern in TARRAY and TQUEUE, can I keep them consistent like this?

mattdurak: ah, I see. Sure that's fine. It does seem like extra boilerplate needed to use this, but there may have been a good reason (maybe for tests/mocks), so let's keep the same design.

tests/two_d_array_int/two_d_array_int.c
```diff
@@ -0,0 +1,299 @@
...
+        switch (action)
+        {
+        default:
+            LogError("Invalid action: %" PRI_MU_ENUM "", MU_ENUM_VALUE(CHAOS_TEST_ACTION, action));
+            break;
+        case CHAOS_TEST_ACTION_FREE_ROW:
+        {
+
+            int free_result = TWO_D_ARRAY_FREE_ROW(TEST_THANDLE)(tdarr, rand_row_index);
+            if (row_state[rand_row_index] == true)
```
mattdurak, line 247: > == true [](http://example.com/codeflow?start=41&length=8)

nit: "== true" is redundant, same below on line 261, 280

tests/two_d_array_int/two_d_array_int.c
```diff
@@ -0,0 +1,299 @@
...
+
+#define N_CHAOS_API_CALLS 5000
+TEST_FUNCTION(TWO_D_ARRAY_chaos)
+{
+    //arrange
+    TWO_D_ARRAY(TEST_THANDLE) tdarr = TWO_D_ARRAY_CREATE(TEST_THANDLE)(ROW_NUM, 5);
+    ASSERT_IS_NOT_NULL(tdarr);
+
+    //act
+    uint32_t iteration_time = 0;
```
mattdurak, line 230: >iteration_time [](http://example.com/codeflow?start=13&length=14)

nit: still says "time" so maybe "iteration_count" or "iteration" or "iteration_number" or something like that.

mattdurak (approved):
:shipit:

tests/two_d_array_int/two_d_array_int.c
M-iceberg: removed

tests/two_d_array_int/two_d_array_int.c
M-iceberg: rename to iteration_count

mattdurak (approved):
:shipit:

PR #360, opened 2024-02-26
fix msrc vuln

anporumb:


inc/c_util/logging_stacktrace.h
```diff
@@ -1,25 +0,0 @@
-// Copyright (c) Microsoft. All rights reserved.
```
dcristoloveanu, line 1: Consider not including this change in this PR, just in case we have to reference it in our case workflow and if it needs to be reviewed by other folks in other orgs - when I looked at it I was wondering if the vuln had anything to do with logging also - then realized that we removed the file because it is not used.

We can easily file a second PR to remove the file from the repo, should be cheap enough.

dcristoloveanu (approved):

inc/c_util/logging_stacktrace.h
anporumb: Codeflow doesn't work for me here. So ... I fumbled around but I think I've got it.

avranju (approved):

PR #386, opened 2024-04-08
constbuffer_array_splitter_split_to_2d_array_impl

M-iceberg:


devdoc/constbuffer_array_splitter_requirements.md
```diff
@@ -58,3 +62,38 @@ This takes the `buffers` from a `CONSTBUFFER_ARRAY_HANDLE` and splits them into
...
+
+This takes the `buffers` from a `CONSTBUFFER_ARRAY_HANDLE` and splits them into a new `CONSTBUFFER_ARRAY_HANDLE*` where each `CONSTBUFFER_ARRAY_HANDLE` is of size `max_buffer_size` (or smaller for the final remaining buffer). The caller is responsible for cleaning up the returned `CONSTBUFFER_ARRAY_HANDLE*`.
+
+If `buffers` is `NULL` then `constbuffer_array_splitter_split_to_array_of_array` shall fail and return `NULL`.
+
+If `max_buffer_size` is `0` then `constbuffer_array_splitter_split_to_array_of_array` shall fail and return `NULL`.
+
+`constbuffer_array_splitter_split_to_array_of_array` shall call `constbuffer_array_get_buffer_count`.
+
+If the buffer count is 0 then `constbuffer_array_splitter_split_to_array_of_array` shall call `constbuffer_array_create_empty` and return the result.
```
mattdurak, line 80: This isn't exactly correct because we need to return an array of `CONSTBUFFER_ARRAY_HANDLE`.

src/constbuffer_array_splitter.c
```diff
@@ -208,3 +208,150 @@ CONSTBUFFER_ARRAY_HANDLE constbuffer_array_splitter_split(CONSTBUFFER_ARRAY_HAND
...
+        result = NULL;
+    }
+    else
+    {
+        uint32_t buffer_count;
+        (void)constbuffer_array_get_buffer_count(buffers, &buffer_count);
+
+        if (buffer_count == 0)
+        {
+            result[0] = constbuffer_array_create_empty();
```
mattdurak, line 229: > result[0] [](http://example.com/codeflow?start=11&length=10)

this is accessing uninitialized memory

mattdurak: need to malloc (e.g. `malloc_2` with a count of 1 and size `sizeof(CONSTBUFFER_ARRAY_HANDLE)`)

mattdurak: Related to this, I think Dan suggested making the result `TARRAY(CONSTBUFFER_ARRAY_HANDLE)`.

TARRAY does two things:

1. The array is ref-counted
2. The array can grow as needed by calling "ensure capacity".

The second point may help for the normal case below where we split to multiple constbuffer arrays.

Alternatively you would need to iterate over everything once to figure out how many CONSTBUFFER_ARRAY_HANDLE's you need (how many splits), and then allocate the array, then create the array entries.

src/constbuffer_array_splitter.c
```diff
@@ -208,3 +208,150 @@ CONSTBUFFER_ARRAY_HANDLE constbuffer_array_splitter_split(CONSTBUFFER_ARRAY_HAND
...
+                                if(arr == NULL)
+                                {
+                                    LogError("constbuffer_array_create_from_start_and_end failed buffers=%p, uint32_t start_buffer_index=%" PRIu32 ", uint32_t buffer_count=%" PRIu32 ", uint32_t start_buffer_offset=%" PRIu32 ", uint32_t end_buffer_size=%" PRIu32 ,
+                                        buffers, start_buffer_index, current_buffer_count, start_buffer_offset, buffer->size);
+                                    result = NULL;
+                                    break;
+                                }
+                                else
+                                {
+                                    result[index] = arr;
```
mattdurak, line 290: >esult[index] [](http://example.com/codeflow?start=37&length=12)

will need to ensure capacity of the TARRAY, or make sure we've allocated result to be large enough before accessing this.

src/constbuffer_array_splitter.c
```diff
@@ -208,3 +208,150 @@ CONSTBUFFER_ARRAY_HANDLE constbuffer_array_splitter_split(CONSTBUFFER_ARRAY_HAND
...
+                    uint32_t start_buffer_offset = 0;
+                    uint32_t end_buffer_size = 0;
+                    uint32_t current_buffer_count = 0; //buffer count in sub-array
+
+                    uint32_t current_buffer_size = 0; //sub-array size
+                    uint32_t i = 0; //index in original buffer
+                    uint32_t index = 0; //index in result array handle*
+                    uint32_t total_size = 0; //total buffer size in current result
+
+                    while(i < buffer_count)
```
mattdurak, line 270: >le( [](http://example.com/codeflow?start=23&length=3)

nit: missing space before parenthesis for control statement, here and below

src/constbuffer_array_splitter.c
```diff
@@ -208,3 +208,150 @@ CONSTBUFFER_ARRAY_HANDLE constbuffer_array_splitter_split(CONSTBUFFER_ARRAY_HAND
...
+                            current_buffer_count++;
+                            current_buffer_size += buffer->size;
+                            if(i == buffer_count - 1)
+                            {
+                                CONSTBUFFER_ARRAY_HANDLE arr = constbuffer_array_create_from_start_and_end(buffers, start_buffer_index, current_buffer_count, start_buffer_offset, buffer->size);
+                                if(arr == NULL)
+                                {
+                                    LogError("constbuffer_array_create_from_start_and_end failed buffers=%p, uint32_t start_buffer_index=%" PRIu32 ", uint32_t buffer_count=%" PRIu32 ", uint32_t start_buffer_offset=%" PRIu32 ", uint32_t end_buffer_size=%" PRIu32 ,
+                                        buffers, start_buffer_index, current_buffer_count, start_buffer_offset, buffer->size);
+                                    result = NULL;
```
mattdurak, line 285: > result = NULL; [](http://example.com/codeflow?start=35&length=15)

will need some cleanup for the result array itself and all of the constbuffer arrays that were successfully created.

src/constbuffer_array_splitter.c
```diff
@@ -208,3 +208,150 @@ CONSTBUFFER_ARRAY_HANDLE constbuffer_array_splitter_split(CONSTBUFFER_ARRAY_HAND
...
+                                    break;
+                                }
+                                else
+                                {
+                                    result[index] = arr;
+                                }
+                            }
+                            i++;
```
mattdurak, line 293: > i++; [](http://example.com/codeflow?start=27&length=5)

nit: seems that you do "i++" or break in every branch, so maybe a for loop would be better just for clarity.

src/constbuffer_array_splitter.c
```diff
@@ -208,3 +208,150 @@ CONSTBUFFER_ARRAY_HANDLE constbuffer_array_splitter_split(CONSTBUFFER_ARRAY_HAND
...
+                                end_buffer_size = 0;
+                                current_buffer_count = 1;
+                                current_buffer_size = buffer->size - end_buffer_size;
+
+                                index++;
+                                i++;
+                            }
+                        }
+                        else
+                        {
```
mattdurak, line 325: suspect that this can be combined with the case above (if greater than or equal). Loop will exit anyway in the "equal" case, but should be less code duplication.

devdoc/constbuffer_array_requirements.md
```diff
@@ -123,6 +123,8 @@ MOCKABLE_FUNCTION(, CONSTBUFFER_ARRAY_HANDLE, constbuffer_array_create_from_star
 
 **SRS_CONSTBUFFER_ARRAY_07_007: [**  `constbuffer_array_create_from_start_and_end` shall allocate memory for a new `CONSTBUFFER_ARRAY_HANDLE`.  **]**
 
+**SRS_CONSTBUFFER_ARRAY_07_015: [** `constbuffer_array_create_from_start_and_end` shall skip any empty buffers in the middle of the original buffer.  **]**
```
dcristoloveanu, line 126: Hmmmm, I wonder why we do that

dcristoloveanu:
> IMPLEMENT_MOCKABLE_FUNCTION(, CONSTBUFFER_ARRAY_HANDLE, constbuffer_array_create_from_start_and_end, CONSTBUFFER_ARRAY_HANDLE, original, uint32_t, start_buffer_index, uint32_t, buffer_count, uint32_t, start_buffer_offset, uint32_t, end_buffer_size)

Ah, this change since I wrote the spike :-).
To be fair constbuffer_array_create_from_start_and_end conveys that we would create a const buffer array given a start and end "address" within the original const buffer array.
I think now the name is not matching as well with the arguments.

---
Refers to: src/constbuffer_array.c:200 in 587cefa. [](commit_id = 587cefada0e0429760be5badf4547e45222bace0, deletion_comment = False)

dcristoloveanu:
> IMPLEMENT_MOCKABLE_FUNCTION(, CONSTBUFFER_ARRAY_HANDLE, constbuffer_array_create_from_start_and_end, CONSTBUFFER_ARRAY_HANDLE, original, uint32_t, start_buffer_index, uint32_t, buffer_count, uint32_t, start_buffer_offset, uint32_t, end_buffer_size)

There is something not nice about mixing offset and size.
I think it is actually end_buffer_offset after all.

---
Refers to: src/constbuffer_array.c:200 in 587cefa. [](commit_id = 587cefada0e0429760be5badf4547e45222bace0, deletion_comment = False)

src/constbuffer_array.c
```diff
@@ -204,7 +204,7 @@ IMPLEMENT_MOCKABLE_FUNCTION(, CONSTBUFFER_ARRAY_HANDLE, constbuffer_array_create
     if (original == NULL)
     {
         /* Codes_SRS_CONSTBUFFER_ARRAY_07_001: [ If original is NULL then constbuffer_array_create_from_start_and_end shall fail and return NULL. ]*/
-        LogError("Invalid arguments: CONSTBUFFER_ARRAY_HANDLE original=%p, uint32_t start_buffer_index=%" PRIu32 ", uint32_t buffer_count=%" PRIu32 ", uint32_t start_buffer_offset=%" PRIu32 ", uint32_t end_buffer_size=%" PRIu32 ,
+        LogError("Invalid arguments: CONSTBUFFER_ARRAY_HANDLE original=%p, uint32_t start_buffer_index=%" PRIu32 ", uint32_t buffer_count=%" PRIu32 ", uint32_t start_buffer_offset=%" PRIu32 ", uint32_t end_buffer_size=%" PRIu32,
```
dcristoloveanu, line 207: >2, [](http://example.com/codeflow?start=226&length=2)

Missing "" at the end

dcristoloveanu:
>         {

This could actually have an optimization that if the entire array is asked for then we simply inc_ref.

---
Refers to: src/constbuffer_array.c:227 in 587cefa. [](commit_id = 587cefada0e0429760be5badf4547e45222bace0, deletion_comment = False)

src/constbuffer_array_splitter.c
```diff
@@ -208,3 +208,353 @@ CONSTBUFFER_ARRAY_HANDLE constbuffer_array_splitter_split(CONSTBUFFER_ARRAY_HAND
...
+        else
+        {
+            uint32_t buffer_array_count = 0; //used to store total number of constbuffer_array in result
+            uint32_t buffer_count;
+
+            /* Codes_SRS_CONSTBUFFER_ARRAY_SPLITTER_07_005: [ constbuffer_array_splitter_split_to_array_of_array shall call constbuffer_array_get_buffer_count to get the total number of buffers. ]*/
+            (void)constbuffer_array_get_buffer_count(buffers, &buffer_count);
+
+            /* Codes_SRS_CONSTBUFFER_ARRAY_SPLITTER_07_006: [ constbuffer_array_splitter_split_to_array_of_array shall call TARRAY_ENSURE_CAPACITY to be able to add a CONSTBUFFER_ARRAY_HANDLE entry in the TARRAY(CONSTBUFFER_ARRAY_HANDLE). ]*/
+            if (TARRAY_ENSURE_CAPACITY(CONSTBUFFER_ARRAY_HANDLE)(temp, 1) != 0)
```
dcristoloveanu, line 243: >   if (TARRAY_ENSURE_CAPACITY(CONSTBUFFER_ARRAY_HANDLE)(temp, 1) != 0) [](http://example.com/codeflow?start=9&length=70)

This will do a realloc behind. It is not necessary. If buffer_count == 0, we can directly create the TARRAY with size 1.

Also for the case where there is no splitting, but we use the entire original array as is we know we have to create it with 1 as capacity.

As a matter of fact we can easily compute in one pass over all the buffers in the original array how many entries we need in the TARRAY so we don't do extra mallocs.

src/constbuffer_array.c
```diff
@@ -281,19 +282,29 @@ IMPLEMENT_MOCKABLE_FUNCTION(, CONSTBUFFER_ARRAY_HANDLE, constbuffer_array_create
...
                             {
-                                result->buffers[i] = end_buffer;
+                                result->buffers[i - empty_buffer_count] = end_buffer;
                             }
                             else
                             {
-                                /* Codes_SRS_CONSTBUFFER_ARRAY_07_008: [ constbuffer_array_create_from_start_and_end shall copy all of the CONSTBUFFER_HANDLES except first and last buffer from each const buffer array in buffer_arrays to the newly constructed array by calling CONSTBUFFER_IncRef. ]*/
-                                CONSTBUFFER_IncRef(original->buffers[start_buffer_index + i]);
-                                result->buffers[i] = original->buffers[start_buffer_index + i];
+                                /* Codes_SRS_CONSTBUFFER_ARRAY_07_015: [ constbuffer_array_create_from_start_and_end shall skip any empty buffers in the middle of the original buffer. ]*/
```
dcristoloveanu, line 293: > /* Codes_SRS_CONSTBUFFER_ARRAY_07_015: [ constbuffe [](http://example.com/codeflow?start=31&length=52)

I don't think we should do this.
This is effectively saying that empty buffers are bad (otherwise why would we skip them?).
If they are bad I would be thinking we should reject them when we create a const buffer array.
Buuuut we do not reject them and maybe there are cases where we actually put them in the array.

I guess the question is why do we care to eliminate the zero buffers in this process of splitting?

dcristoloveanu (changes requested):
:clock1:

devdoc/constbuffer_array_splitter_requirements.md
```diff
@@ -13,6 +13,10 @@ For example, it is used to upload large payloads to Blob store, split across lar
 MOCKABLE_FUNCTION(, CONSTBUFFER_ARRAY_HANDLE, constbuffer_array_splitter_split, CONSTBUFFER_ARRAY_HANDLE, buffers, uint32_t, max_buffer_size);
 ```
 
+```c
```
mattdurak, line 16: nit: remove, can keep these all in the same code block

inc/c_util/constbuffer_array_splitter_array.h
```diff
@@ -0,0 +1,30 @@
+// Copyright (c) Microsoft. All rights reserved.
+
+#ifndef CONSTBUFFER_ARRAY_SPLITTER_ARRAY_H
```
mattdurak, line 3: >CONSTBUFFER_ARRAY_SPLITTER_ARRAY_H [](http://example.com/codeflow?start=8&length=34)

maybe this file should be called `constbuffer_array_tarray` instead

devdoc/constbuffer_array_requirements.md
M-iceberg: removed

src/constbuffer_array.c
M-iceberg: deleted this

M-iceberg:
> IMPLEMENT_MOCKABLE_FUNCTION(, CONSTBUFFER_ARRAY_HANDLE, constbuffer_array_create_from_start_and_end, CONSTBUFFER_ARRAY_HANDLE, original, uint32_t, start_buffer_index, uint32_t, buffer_count, uint32_t, start_buffer_offset, uint32_t, end_buffer_size)

changed to name constbuffer_array_create_from_buffer_offset_and_count

---
In reply to: [2062991837](https://github.com/Azure/c-util/pull/386#issuecomment-2062991837) [](http://example.com/codeflow?ancestors=2062991837)

---
Refers to: src/constbuffer_array.c:200 in 587cefa. [](commit_id = 587cefada0e0429760be5badf4547e45222bace0, deletion_comment = False)

src/constbuffer_array.c
M-iceberg: it's the end of the params, guess don't need a '' ?

M-iceberg:
> IMPLEMENT_MOCKABLE_FUNCTION(, CONSTBUFFER_ARRAY_HANDLE, constbuffer_array_create_from_start_and_end, CONSTBUFFER_ARRAY_HANDLE, original, uint32_t, start_buffer_index, uint32_t, buffer_count, uint32_t, start_buffer_offset, uint32_t, end_buffer_size)

modified

---
In reply to: [2062996585](https://github.com/Azure/c-util/pull/386#issuecomment-2062996585) [](http://example.com/codeflow?ancestors=2062996585)

---
Refers to: src/constbuffer_array.c:200 in 587cefa. [](commit_id = 587cefada0e0429760be5badf4547e45222bace0, deletion_comment = False)

M-iceberg:
>         {

done

---
In reply to: [2062997494](https://github.com/Azure/c-util/pull/386#issuecomment-2062997494) [](http://example.com/codeflow?ancestors=2062997494)

---
Refers to: src/constbuffer_array.c:227 in 587cefa. [](commit_id = 587cefada0e0429760be5badf4547e45222bace0, deletion_comment = False)

src/constbuffer_array_splitter.c
```diff
@@ -208,3 +208,353 @@ CONSTBUFFER_ARRAY_HANDLE constbuffer_array_splitter_split(CONSTBUFFER_ARRAY_HAND
...
+                        failed = true;
+                        LogError("constbuffer_array_create_empty failed");
+                    }
+                }
+                else
+                {
+                    uint32_t remaining_buffer_size;
+
+                    /* Codes_SRS_CONSTBUFFER_ARRAY_SPLITTER_07_008: [ constbuffer_array_splitter_split_to_array_of_array shall call constbuffer_array_get_all_buffers_size for buffers and store the result as remaining_buffer_size. ]*/
+                    int temp_result = constbuffer_array_get_all_buffers_size(buffers, &remaining_buffer_size);
```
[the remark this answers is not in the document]
M-iceberg: done

src/constbuffer_array_splitter.c
M-iceberg: changed to one malloc

devdoc/constbuffer_array_splitter_requirements.md
M-iceberg: removed

inc/c_util/constbuffer_array_splitter_array.h
M-iceberg: done

devdoc/constbuffer_array_requirements.md
```diff
@@ -99,35 +99,37 @@ Note: `constbuffer_array_create_from_buffer_index_and_count` does not increment
...
 
-**SRS_CONSTBUFFER_ARRAY_07_003: [** If `start_buffer_index + buffer_count` is greater than the number of buffers in `original` then `constbuffer_array_create_from_start_and_end` shall fail and return `NULL`.  **]**
+**SRS_CONSTBUFFER_ARRAY_07_003: [** If `start_buffer_index + buffer_count` is greater than the number of buffers in `original` then `constbuffer_array_create_from_buffer_offset_and_count` shall fail and return `NULL`.  **]**
 
-**SRS_CONSTBUFFER_ARRAY_07_013: [** If `buffer_count` is 1, `constbuffer_array_create_from_start_and_end` shall get the only buffer by calling `CONSTBUFFER_CreateFromOffsetAndSize` with paramter `start_buffer_offset` and `end_buffer_size`. **]**
+**SRS_CONSTBUFFER_ARRAY_07_015: [** If buffer count is equal to original buffer count, start buffer offset is `0` and end buffer offset is equal to the size of the last buffer in `original`, then `constbuffer_array_create_from_buffer_offset_and_count` shall increment the reference count of `original` and return `original`.  **]**
```
dcristoloveanu, line 116: >buffer count [](http://example.com/codeflow?start=39&length=12)

I assume this is `buffer_count`

devdoc/constbuffer_array_splitter_requirements.md
```diff
@@ -58,3 +60,49 @@ This takes the `buffers` from a `CONSTBUFFER_ARRAY_HANDLE` and splits them into
...
+
+**SRS_CONSTBUFFER_ARRAY_SPLITTER_07_001: [** If `buffers` is `NULL` then `constbuffer_array_splitter_split_to_array_of_array` shall fail and return `NULL`. **]**
+
+**SRS_CONSTBUFFER_ARRAY_SPLITTER_07_002: [** If `max_buffer_size` is `0` then `constbuffer_array_splitter_split_to_array_of_array` shall fail and return `NULL`. **]**
+
+**SRS_CONSTBUFFER_ARRAY_SPLITTER_07_005: [** `constbuffer_array_splitter_split_to_array_of_array` shall call `constbuffer_array_get_buffer_count` to get the total number of buffers. **]**
+
+**SRS_CONSTBUFFER_ARRAY_SPLITTER_07_008: [** `constbuffer_array_splitter_split_to_array_of_array` shall call `constbuffer_array_get_all_buffers_size` for `buffers` and store the result as `remaining_buffer_size`. **]**
+
+**SRS_CONSTBUFFER_ARRAY_SPLITTER_07_021: [** If `buffer_count` or remaining_buffer_size is `0`, `constbuffer_array_splitter_split_to_array_of_array` shall call `TARRAY_CREATE_WITH_CAPACITY` with size 1.  **]**
```
dcristoloveanu, line 80: >remaining_buffer_size [](http://example.com/codeflow?start=66&length=21)

remaining_buffer_size sounds like an implementation detail. This is probably because first the code was written and then specs were done :-).

I think the specs should look like:
If the total size for all buffers in `buffers` is 0, then  `constbuffer_array_splitter_split_to_array_of_array` shall call `TARRAY_CREATE_WITH_CAPACITY` with size 1.

dcristoloveanu: Actually it should probably be like:
If the total size for all buffers in buffers is 0:
- `constbuffer_array_splitter_split_to_array_of_array` shall call `TARRAY_CREATE_WITH_CAPACITY` with size 1.
- `constbuffer_array_splitter_split_to_array_of_array` shall call `constbuffer_array_create_empty` and store the created const buffer array in the first entry of the TARRAY that was created.
- `constbuffer_array_splitter_split_to_array_of_array` shall return the TARRAY.

dcristoloveanu: After reading all the specs here, I think if you rework a little the specs to not have `remaining_buffer_size` (which is internal implementation detail of the code it should be pretty much fine.

dcristoloveanu: Yeah, after reading all of this I think the changes should be minor:
- the one mentioned above about the case of 0 buffers
- Normally we should not have remaining_buffer_size here in the spec, which is a detail of how the variable is named in the code (suggest saying "total size of buffers in `buffers`). If you really want to shorten it in the specs you can do something like:

**SRS_CONSTBUFFER_ARRAY_SPLITTER_07_008: [** `constbuffer_array_splitter_split_to_array_of_array` shall call `constbuffer_array_get_all_buffers_size` for `buffers` and to obtain the total size of all buffers in `buffers`. **]** The total size will be reference in subsequent specs as `total_buffer_size`.

inc/c_util/constbuffer_array_tarray.h
```diff
@@ -0,0 +1,30 @@
...
+#define CONSTBUFFER_ARRAY_TARRAY_H
+
+#ifdef __cplusplus
+#include <cstdint>
+#else
+#include <stdint.h>
+#endif
+
+#include "c_util/tarray.h"
+#include "c_pal/thandle.h"
```
dcristoloveanu, line 13: Can we have c_pal before c_util so that c_util includes are also grouped together?

Usually the order we use is:
- macro_utils
- c_pal
- c_util

etc.

src/constbuffer_array_splitter.c
```diff
@@ -208,3 +208,248 @@ CONSTBUFFER_ARRAY_HANDLE constbuffer_array_splitter_split(CONSTBUFFER_ARRAY_HAND
...
+{
+    TARRAY(CONSTBUFFER_ARRAY_HANDLE) result = NULL;
+
+    if (
+        /* Codes_SRS_CONSTBUFFER_ARRAY_SPLITTER_07_001: [ If buffers is NULL then constbuffer_array_splitter_split_to_array_of_array shall fail and return NULL. ]*/
+        buffers == NULL ||
+        /* Codes_SRS_CONSTBUFFER_ARRAY_SPLITTER_07_002: [ If max_buffer_size is 0 then constbuffer_array_splitter_split_to_array_of_array shall fail and return NULL. ]*/
+        max_buffer_size == 0)
+    {
+        LogError("Invalid args : CONSTBUFFER_ARRAY_HANDLE buffers = %p, size_t max_buffer_size = %" PRIu32,
```
dcristoloveanu, line 222: >2, [](http://example.com/codeflow?start=105&length=2)

Let's have a "" at the end. It clearly shows where the log line ends and where the args begin

src/constbuffer_array_splitter.c
```diff
@@ -208,3 +208,248 @@ CONSTBUFFER_ARRAY_HANDLE constbuffer_array_splitter_split(CONSTBUFFER_ARRAY_HAND
...
+            buffers, max_buffer_size);
+    }
+    else
+    {
+        uint32_t buffer_array_count = 0; //used to store total number of constbuffer_array already stored in result -> release when failed
```
dcristoloveanu, line 227: >//used to store total number of constbuffer_array already stored in result -> release when failed [](http://example.com/codeflow?start=41&length=97)

I like this comment

src/constbuffer_array_splitter.c
```diff
@@ -208,3 +208,248 @@ CONSTBUFFER_ARRAY_HANDLE constbuffer_array_splitter_split(CONSTBUFFER_ARRAY_HAND
...
+                /* Codes_SRS_CONSTBUFFER_ARRAY_SPLITTER_07_022: [ If remaining_buffers_size is smaller or equal to max_buffer_size, constbuffer_array_splitter_split_to_array_of_array shall call TARRAY_CREATE_WITH_CAPACITY with size 1, inc ref the original buffer and return it. ]*/
+                TARRAY(CONSTBUFFER_ARRAY_HANDLE) temp = TARRAY_CREATE_WITH_CAPACITY(CONSTBUFFER_ARRAY_HANDLE)(1);
+                if (temp == NULL)
+                {
+                    /* Codes_SRS_CONSTBUFFER_ARRAY_SPLITTER_07_004: [ If there are any other failures then constbuffer_array_splitter_split_to_array_of_array shall fail and return NULL. ]*/
+                    LogError("TARRAY_CREATE_WITH_CAPACITY(CONSTBUFFER_ARRAY_HANDLE)(split_array_count=1) failed");
+                }
+                else
+                {
+                    /* Codes_SRS_CONSTBUFFER_ARRAY_SPLITTER_07_020: [ constbuffer_array_splitter_split_to_array_of_array shall succeed and return the new TARRAY(CONSTBUFFER_ARRAY_HANDLE). ]*/
```
dcristoloveanu, line 283: Very nice, now I could easily step through these cases.

src/constbuffer_array_splitter.c
```diff
@@ -208,3 +208,248 @@ CONSTBUFFER_ARRAY_HANDLE constbuffer_array_splitter_split(CONSTBUFFER_ARRAY_HAND
...
+                                        CONSTBUFFER_ARRAY_HANDLE arr2 = constbuffer_array_create_from_buffer_offset_and_count(buffers, start_buffer_index, 1, start_buffer_offset, current_buffer_remaining_size);
+                                        if (arr2 == NULL)
+                                        {
+                                            /* Codes_SRS_CONSTBUFFER_ARRAY_SPLITTER_07_004: [ If there are any other failures then constbuffer_array_splitter_split_to_array_of_array shall fail and return NULL. ]*/
+                                            failed = true;
+                                            LogError("constbuffer_array_create_from_buffer_offset_and_count failed buffers=%p, uint32_t start_buffer_index=%" PRIu32 ", uint32_t buffer_count=%" PRIu32 ", uint32_t start_buffer_offset=%" PRIu32 ", uint32_t end_buffer_size=%" PRIu32,
+                                                buffers, start_buffer_index, 1, start_buffer_offset, max_buffer_size);
+                                            CONSTBUFFER_DecRef(curr_buffer);
+                                            break;
+                                        }
```
dcristoloveanu, line 418: I don't quite like the fact that we break and dec ref on each failure path.
It makes the code brittle. Imagine if someone changes the code and forgets to do the DecRef on an error path. That is a bug. Which will likely get caught by unit tests if we have coverage. We generally do have coverage, but point is that we can write the code slightly better to make it less brittle.

We could break if failed is set to true between lines 433 and 434. Then we would not need to do DecRef on each failure path

src/constbuffer_array_tarray.c
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) Microsoft. All rights reserved.
+
+#include "c_pal/gballoc_hl.h"
+#include "c_pal/gballoc_hl_redirect.h"
+
+#include "c_util/tarray.h"
+#include "c_pal/thandle.h"
```
dcristoloveanu, line 7: >include "c_pal/thandle.h" [](http://example.com/codeflow?start=1&length=25)

Can we have the c_pal chilling with the other c_pal includes please?

tests/constbuffer_array_splitter_ut/constbuffer_array_splitter_ut.c
```diff
@@ -992,4 +1001,810 @@ TEST_FUNCTION(constbuffer_array_splitter_split_with_all_empty_buffers_fails_if_u
     real_constbuffer_array_dec_ref(buffers);
 }
 
+/* constbuffer_array_splitter_split_to_array_of_array */
+
+/* Tests_SRS_CONSTBUFFER_ARRAY_SPLITTER_07_001: [ If buffers is NULL then constbuffer_array_splitter_split_to_array_of_array shall fail and return NULL. ]*/
+TEST_FUNCTION(constbuffer_array_splitter_split_to_array_of_array_with_null_buffers_fails)
+{
+    /// act
```
dcristoloveanu, line 1009: > /// act [](http://example.com/codeflow?start=3&length=8)

Missing /// arrage
We usually put it in the test for consistency, even if there are no actual arrange calls

tests/constbuffer_array_splitter_ut/constbuffer_array_splitter_ut.c
```diff
@@ -992,4 +1001,810 @@ TEST_FUNCTION(constbuffer_array_splitter_split_with_all_empty_buffers_fails_if_u
...
+    CONSTBUFFER_ARRAY_HANDLE buffers_temp1 = generate_test_buffer_array(4, 1000);
+    CONSTBUFFER_ARRAY_HANDLE empty = generate_test_buffer_array(3, 0);
+    CONSTBUFFER_ARRAY_HANDLE buffers_temp2 = generate_test_buffer_array(6, 1000);
+
+    CONSTBUFFER_ARRAY_HANDLE buffers_to_merge[4];
+    buffers_to_merge[0] = buffers_temp1;
+    buffers_to_merge[1] = empty;
+    buffers_to_merge[2] = buffers_temp2;
+    buffers_to_merge[3] = empty;
+    CONSTBUFFER_ARRAY_HANDLE buffers = real_constbuffer_array_create_from_array_array(buffers_to_merge, sizeof(buffers_to_merge) / sizeof(buffers_to_merge[0]));
```
dcristoloveanu, line 1542: >buffers [](http://example.com/codeflow?start=29&length=7)

Shouldn't this be asserted? 
Here and everywhere

tests/reals/real_constbuffer_array_tarray.c
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#include "real_gballoc_hl.h"
+#include "real_gballoc_hl_renames.h"
+#include "c_pal/gballoc_hl_redirect.h"
```
dcristoloveanu, line 6: >clude "c_pal/gballoc_hl_redirect.h" [](http://example.com/codeflow?start=3&length=35)

Is this needed?

dcristoloveanu (approved):
:shipit:

devdoc/constbuffer_array_requirements.md
M-iceberg: yes

devdoc/constbuffer_array_splitter_requirements.md
M-iceberg: done

inc/c_util/constbuffer_array_tarray.h
M-iceberg: done

src/constbuffer_array_splitter.c
M-iceberg: added, here and everywhere else

src/constbuffer_array_splitter.c
```diff
@@ -208,3 +208,248 @@ CONSTBUFFER_ARRAY_HANDLE constbuffer_array_splitter_split(CONSTBUFFER_ARRAY_HAND
...
+
+                            if(i == buffer_count - 1)
+                            {
+                                /* Codes_SRS_CONSTBUFFER_ARRAY_SPLITTER_07_014: [ If current buffer is the last buffer in the original constbuffer_array, constbuffer_array_splitter_split_to_array_of_array shall store the sub - tarray with size smaller than max_buffer_size to result. ]*/
+                                CONSTBUFFER_ARRAY_HANDLE arr = constbuffer_array_create_from_buffer_offset_and_count(buffers, start_buffer_index, current_buffer_count, start_buffer_offset, end_buffer_size);
+
+                                if(arr == NULL)
+                                {
+                                    /* Codes_SRS_CONSTBUFFER_ARRAY_SPLITTER_07_004: [ If there are any other failures then constbuffer_array_splitter_split_to_array_of_array shall fail and return NULL. ]*/
+                                    failed = true;
```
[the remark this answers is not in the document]
M-iceberg: modified, here and everywhere else

src/constbuffer_array_splitter.c
M-iceberg: done

src/constbuffer_array_tarray.c
M-iceberg: done

tests/reals/real_constbuffer_array_tarray.c
M-iceberg: removed

M-iceberg: added back, it's needed in THANDLE_LL_TYPE_DEFINE

tests/constbuffer_array_splitter_ut/constbuffer_array_splitter_ut.c
M-iceberg: added

tests/constbuffer_array_splitter_ut/constbuffer_array_splitter_ut.c
M-iceberg: added

dcristoloveanu (approved):
:shipit:

PR #392, opened 2024-04-24
Add "real_interlocked_renames.h" to rc_ptr.c

darobs:
This makes it so that we don't need to call out expects for interlocked functions if we also have to mock interlocked functions and rc_ptr functions in the same UT.


mattdurak (approved):
:shipit:

PR #433, opened 2024-10-15
add get_count for tqueue

jasmineymlo:


devdoc/tqueue_requirements.md
```diff
@@ -273,3 +274,21 @@ TQUEUE_POP_RESULT TQUEUE_POP(T)(TQUEUE(T) tqueue, T* item, void* pop_function_co
     - **SRS_TQUEUE_01_045: [** `TQUEUE_POP(T)` shall call `dispose_item_function` with `dispose_item_function_context` as `context` and the array entry value whose state was changed to `POPPING` as `item`. **]**
 
   - **SRS_TQUEUE_01_034: [** `TQUEUE_POP(T)` shall set the state to `NOT_USED` by using `interlocked_exchange`, succeed and return `TQUEUE_POP_OK`. **]**
+
+
+### TQUEUE_POP(T)
```
nishudeshu, line 279: Should this be ### TQUEUE_GET_COUNT(T) ?

devdoc/tqueue_requirements.md
```diff
@@ -273,3 +274,21 @@ TQUEUE_POP_RESULT TQUEUE_POP(T)(TQUEUE(T) tqueue, T* item, void* pop_function_co
...
+
+`TQUEUE_GET_COUNT(T)` returns the item count of the queue if available.
+
+**SRS_TQUEUE_22_001: [** If `tqueue` is `NULL` then `TQUEUE_GET_COUNT(T)` shall return zero. **]**
+
+**SRS_TQUEUE_22_002: [** `TQUEUE_GET_COUNT(T)` shall obtain the current head queue by calling `interlocked_add_64`. **]**
+
+**SRS_TQUEUE_22_003: [** `TQUEUE_GET_COUNT(T)` shall obtain the current tail queue by calling `interlocked_add_64`. **]**
+
+**SRS_TQUEUE_22_004: [** If the queue is empty (current tail >= current head), `TQUEUE_GET_COUNT(T)` shall return zero. **]**
```
nishudeshu, line 292: The condition for empty might be `current_head == (current_tail - 1)` if this is a circular queue.

inc/c_util/tqueue_ll.h
```diff
@@ -378,13 +385,45 @@ TQUEUE_POP_RESULT TQUEUE_LL_POP(C)(TQUEUE_LL(T) tqueue, T* item, void* copy_item
...
+        LogError("Invalid arguments: TQUEUE_LL(" MU_TOSTRING(T) ") tqueue=%p.", tqueue);                                                                            \
+        result = 0;                                                                                                                                                 \
+    }                                                                                                                                                               \
+    else                                                                                                                                                            \
+    {                                                                                                                                                               \
+        /* Codes_SRS_TQUEUE_22_002: [ TQUEUE_GET_COUNT(T) shall obtain the current head queue by calling interlocked_add_64. ]*/                                    \
+        int64_t current_head = interlocked_add_64((volatile_atomic int64_t*)&tqueue->head, 0);                                                                      \
```
nishudeshu, line 402: Should this be `interlocked_add_64(&tqueue->head, 0) % tqueue->queue_size`

jasmineymlo: interlocked_add_64(&tqueue->head, 0) % tqueue->queue_size would get the index to the queue.  GET_COUNT is trying to get the difference between the tail and head.

devdoc/tqueue_requirements.md
jasmineymlo: See SRS_TQUEUE_01_035.  head and tail are intialized to 0 when the queue was created (empty).

devdoc/tqueue_requirements.md
jasmineymlo: fixed

anporumb:
> Note: plans exist to extend this queue into a growable queue.

should list the devops item that tracks this effort.

---
Refers to: devdoc/tqueue_requirements.md:59 in ed43dc9. [](commit_id = ed43dc9774268ae5413f522d0c2438414b227586, deletion_comment = False)

devdoc/tqueue_requirements.md
```diff
@@ -273,3 +274,21 @@ TQUEUE_POP_RESULT TQUEUE_POP(T)(TQUEUE(T) tqueue, T* item, void* pop_function_co
     - **SRS_TQUEUE_01_045: [** `TQUEUE_POP(T)` shall call `dispose_item_function` with `dispose_item_function_context` as `context` and the array entry value whose state was changed to `POPPING` as `item`. **]**
 
   - **SRS_TQUEUE_01_034: [** `TQUEUE_POP(T)` shall set the state to `NOT_USED` by using `interlocked_exchange`, succeed and return `TQUEUE_POP_OK`. **]**
+
+
+### TQUEUE_GET_COUNT(T)
```
anporumb, line 279: >TQUEUE_GET_COUNT(T) [](http://example.com/codeflow?start=4&length=19)

Where can this even be used?

No amount of items in the queue is "real" as in:
step 1: GET_COUNT
step 2: do_something with GET_COUNT... 

between step 1 and step 2 the item count can change. Then how can this information be useful?

I suspect a higher level design errr... "decision" which leads to asking for the number of items in the queue.

devdoc/tqueue_requirements.md
```diff
@@ -273,3 +274,21 @@ TQUEUE_POP_RESULT TQUEUE_POP(T)(TQUEUE(T) tqueue, T* item, void* pop_function_co
...
+```c
+int64_t TQUEUE_GET_COUNT(T)(TQUEUE(T) tqueue);
```
anporumb, line 281: >int64_t [](http://example.com/codeflow?start=0&length=7)

If the queue is created with INT32_MAX+2 elements (note, create takes a uint32_t) and there are in the queue INT32_MAX+1 elements, what does this function return?

anporumb, line 281: >int64_t [](http://example.com/codeflow?start=0&length=7)

so why is this a "int64_t when the queue is created with a uint32_t type?

inc/c_util/tqueue_ll.h
anporumb, line 402: >(volatile_atomic int64_t*) [](http://example.com/codeflow?start=50&length=26)

why is this cast necessary?

Here and everywhere

inc/c_util/tqueue_ll.h
```diff
@@ -378,13 +385,45 @@ TQUEUE_POP_RESULT TQUEUE_LL_POP(C)(TQUEUE_LL(T) tqueue, T* item, void* copy_item
...
+        /* Codes_SRS_TQUEUE_22_003: [ TQUEUE_GET_COUNT(T) shall obtain the current tail queue by calling interlocked_add_64. ]*/                                    \
+        int64_t current_tail = interlocked_add_64((volatile_atomic int64_t*)&tqueue->tail, 0);                                                                      \
```
anporumb, line 404: >current_tail [](http://example.com/codeflow?start=16&length=12)

between lines 404 and 402 the queue configuration can change. Moreover, there are no guarantees which of the head/tail change in which order.

anporumb (dismissed):
:shipit:

devdoc/tqueue_requirements.md
jasmineymlo: This API is mostly used for metric and we only need the count as a data point.  
- change name to GET_VOLATILE_COUNT

inc/c_util/tqueue_ll.h
jasmineymlo: add while loop to read  tail->head->tail to reverify the count

inc/c_util/tqueue_ll.h
jasmineymlo: compile error without

devdoc/tqueue_requirements.md
```diff
@@ -273,3 +274,23 @@ TQUEUE_POP_RESULT TQUEUE_POP(T)(TQUEUE(T) tqueue, T* item, void* pop_function_co
...
 
   - **SRS_TQUEUE_01_034: [** `TQUEUE_POP(T)` shall set the state to `NOT_USED` by using `interlocked_exchange`, succeed and return `TQUEUE_POP_OK`. **]**
+
+
+### TQUEUE_GET_VOLATILE_COUNT(T)
+```c
+int64_t TQUEUE_GET_VOLATILE_COUNT(T)(TQUEUE(T) tqueue);
+```
+
+`TQUEUE_GET_VOLATILE_COUNT(T)` returns the item count of the queue if available.
```
dcristoloveanu, line 284: >if available [](http://example.com/codeflow?start=67&length=12)

This "if available" sounds weird.
The count is always available. Might be zero :-).

dcristoloveanu, line 284: I would put a note here stating the (obvious) fact that the returned value is a point in time value and if the caller needs any synchronization related to the count obtained it is on the caller to lock/use other means of synchronizing.

inc/c_util/tqueue_ll.h
```diff
@@ -378,13 +385,52 @@ TQUEUE_POP_RESULT TQUEUE_LL_POP(C)(TQUEUE_LL(T) tqueue, T* item, void* copy_item
...
+            current_tail = interlocked_add_64((volatile_atomic int64_t*)&tqueue->tail, 0);                                                                          \
+            /* Codes_SRS_TQUEUE_22_002: [ TQUEUE_GET_VOLATILE_COUNT(T) shall obtain the current head queue by calling interlocked_add_64. ]*/                       \
+            current_head = interlocked_add_64((volatile_atomic int64_t*)&tqueue->head, 0);                                                                          \
+            /* Codes_SRS_TQUEUE_22_006: [ TQUEUE_GET_VOLATILE_COUNT(T) shall obtain the current tail queue again by calling interlocked_add_64 and compare with the previosuly obtained tail value.  The tail value is valid only if it has not changed. ]*/   \
+        } while (current_tail != interlocked_add_64((volatile_atomic int64_t*) & tqueue->tail, 0));                                                                 \
+                                                                                                                                                                    \
```
dcristoloveanu, line 411: I have to question why do we really want a loop?
Frankly the first values we compute are as good as the 2nd one or the next one.

The loop achieves "let's obtain the head while we know the tail has not changed".
But it won't achieve anyway "let's obtain the tail while the head does not change".

And any of them can change the microsecond after we exit the loop anyway...

So, simplicity might be better here, just saying.

dcristoloveanu (dismissed):
:shipit:

inc/c_util/tqueue_ll.h
jasmineymlo: The while loop is to ensure we returns the correct count at a point of time.  Consider the following scenario without the loop (only get tail and head once):
assume tail at 1 and head at 2 => 1 item in queue
1) get tail => 1
2) some other threads execute and do push+pop+push+pop+push+pop.... which advanced tail to 100 and head to 101
3) get head => 101
Above would have get_count return 100.  But actually, the max items in queue at any time during the above period was 2.

dcristoloveanu: Makes sense, it is better, basically tail is some sort of invariant.

dcristoloveanu (approved):
:shipit:

PR #462, opened 2024-12-10
Move batch_queue to c_util

dcristoloveanu:
Move batch_queue to c_util


dcristoloveanu:
Meh, cannot be moved to c-util

PR #465, opened 2024-12-16
Async op samples usage

mattdurak:
This adds a "samples" directory with sample usage of `async_op`. It shows how `async_op` may be used by various compositions of layered modules, including when a module calls another module with asynchronous retries or when a module must cancel all pending `async_op`s that are in-flight when it is closed.

The samples include some basic test coverage (though they do not have any "chaos" testing) to validate.

For now, this serves as a sample set of "design patterns", but it may be possible to extract some of this into macros or other modules for re-use.


samples/async_op_samples_int/ml_async_op_module_with_retries.c
```diff
@@ -0,0 +1,458 @@
...
+                        else
+                        {
+                            THANDLE_ASSIGN(ASYNC_OP)(&ll_async_op, NULL);
+                        }
+                        srw_lock_ll_release_exclusive(&async_op_context->ll_async_op_lock);
+                    }
+
+                    // 11. Provide the async_op to the caller
+                    //    After this (but before returning) it is possible that the caller may call cancel because they may have access to the async_op_out pointer in another thread
+                    THANDLE_MOVE(ASYNC_OP)(async_op_out, &async_op);
```
dcristoloveanu, line 431: Hmmm, but that is error in the caller module.
The caller module should never attempt to use an THANDLE(ASYNC_OP) before we return 0.

That is because there is no way for the caller module to know that 0 will be returned. It seems like the pattern in the caller module is wrong, passing directly some THANDLE(ASYNC_OP)* to be filled in, but not actually checking whether the call succeeded.

I understand how the MOVE here might help, but in all fairness it hides a problem in the caller.

samples/async_op_samples_int/hl_async_op_module.c
```diff
@@ -0,0 +1,316 @@
...
+                }
+                else
+                {
+                    // 7. Store the ll_async_op in the context on success so that it can be canceled
+                    //    This must happen before the next line so that the ll_async_op is set before cancel is called
+                    THANDLE_MOVE(ASYNC_OP)(&async_op_context->ll_async_op, &ll_async_op);
+
+                    // 8. Provide the async_op to the caller
+                    //    After this (but before returning) it is possible that the caller may call cancel because they may have access to the async_op_out pointer in another thread
+                    THANDLE_MOVE(ASYNC_OP)(async_op_out, &async_op);
```
dcristoloveanu, line 290: Same comment like in ml, seems like the upper layer would not play nice if they do that.

samples/async_op_samples_int/ml_async_op_module_with_retries.c
```diff
@@ -0,0 +1,458 @@
...
+            {
+                // 2. Any context needed is stored in async_op->context
+                ML_ASYNC_OP_MODULE_EXECUTE_CONTEXT* async_op_context = async_op->context;
+
+                async_op_context->callback = callback;
+                async_op_context->context = context;
+                async_op_context->handle = handle;
+                async_op_context->complete_in_ms = complete_in_ms;
+
+                // 3. Synchronization of ll_async_op between chained async calls requires a lock, a cancelation flag, and a counter (ll_async_op_step)
```
dcristoloveanu, line 387: >chained [](http://example.com/codeflow?start=61&length=7)

This is the with_retries example we're in here

dcristoloveanu, line 387: >ll_async_op_step [](http://example.com/codeflow?start=133&length=16)

There's no step since it is not chained, likely has to be ll_async_op_epoch

dcristoloveanu (dismissed):
:shipit:

samples/async_op_samples_int/hl_async_op_module.c
mattdurak: Added some notes in the MD and simplified the comment with best practice to maybe just return ASYNC_OP instead

samples/async_op_samples_int/async_op_samples_int.c
```diff
@@ -0,0 +1,1441 @@
...
+}
+
+TEST_FUNCTION(all_modules_can_execute_async_successfully_with_fake_cancel)
+{
+    for (uint32_t i = 0; i < g_test_module_count_with_fake_cancel; i++)
+    {
+        // arrange
+        LogInfo("Begin Testing module %s", g_test_modules_with_fake_cancel[i].module_name);
+        THANDLE(ASYNC_OP) async_op = NULL;
+        OPERATION_RESULT_CONTEXT* result_context = malloc(sizeof(OPERATION_RESULT_CONTEXT));
```
mattdurak, line 410: >malloc(sizeof(OPERATION_RESULT_CONTEXT)); [](http://example.com/codeflow?start=51&length=41)

heap allocation seems to make helgrind happy

dcristoloveanu (approved):

PR #469, opened 2024-12-27
CHANNEL revisit considerations

parth21999:


CMakeLists.txt
```diff
@@ -200,6 +201,7 @@ set(c_util_h_files
     ./inc/c_util/cancellation_token.h
     ./inc/c_util/channel.h
     ./inc/c_util/channel_common.h
+    ./inc/c_util/channel_2.h
```
dcristoloveanu, line 204: >channel_2.h [](http://example.com/codeflow?start=17&length=11)

Oh wow, a complete rewrite I see.

devdoc/channel_2.md
```diff
@@ -0,0 +1,72 @@
+# channel_2 requirements:
```
dcristoloveanu, line 1: >channel_2 [](http://example.com/codeflow?start=2&length=9)

Likely wants some backticks

dcristoloveanu, line 1: >: [](http://example.com/codeflow?start=24&length=1)

Nitpick: can lose the colon (it is a title)

devdoc/channel_2.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Overview
+
+The current implementation of `channel` has certain pitfalls. `channel_2` aims to address these pitfalls by making the following changes:
```
dcristoloveanu, line 5: >certain pitfalls. `channel_2` aims to address these pitfalls by making the following changes: [](http://example.com/codeflow?start=44&length=93)

Would be good to mention what the pitfalls are. Otherwise it sounds like this:

"There were some problems with channel. We're not going to say what the problems are but here's a solution to the unknown problems" :-)

devdoc/channel_2.md
```diff
@@ -0,0 +1,72 @@
...
+
+1. Introduce a `channel_2_reset` API: This API will allow the user to reset the `channel_2` to a clean state. It will cause all pending operations to be abandoned.
```
dcristoloveanu, line 7: >channel_2_reset [](http://example.com/codeflow?start=16&length=15)

This reset smells like close and then open again.
I do have to wonder why it is not modelled as channel having an open/closed state and then the user can do whatever they want to "reset"

devdoc/channel_2.md
```diff
@@ -0,0 +1,72 @@
...
+    The receiver calls `channel_pull` 3 times and the sender cancels `op2` concurrently.
+    
+    The callback for `op1` is called with `data1` and `CHANNEL_CALLBACK_RESULT_OK`. 
+
+    The callback for `op2` is called with `NULL`  and `CHANNEL_CALLBACK_RESULT_CANCELLED`.
+
+    The callback for `op3` is called with `data3` and `CHANNEL_CALLBACK_RESULT_OK`.
+
+    The receiver will now contain [`data1`, `data3`].
+
```
dcristoloveanu, line 22: This paragraph is saying "channel has a problem that when one operation is cancelled, the subsequent ones are not.

This  implies that the user of channel wants to preserve the order of the operations. That implies completing POPs in the order that they were PUSHed.

This document is not addressing the topic of ordering for channel.

"channel.md" has "order" in it one time:
"`channel_pull` and `channel_push` can be called from different threads in any order. The callbacks are executed on threadpool threads."

So, we're not defining anywhere even for channel (v1) that the order is of concern.

Now, if we're saying that we do need order to be preserved, I think we should be much crisper which order is to be preserved (the order in which channel sees the PUSH operations for example).

I think this has nothing to do with ASYNC_OP. ASYNC_OP gives a framework for an asynchronous operation which is also cancellable.

devdoc/channel_2.md
```diff
@@ -0,0 +1,72 @@
...
+    This order of the data on the receiver side of the `channel` is not the same as the order of the `PUSH` operations on the sender side. This is not desirable.
+
+    - Insufficient value: There is also not much value in providing the user with an `ASYNC_OP` for a single operation. We have not found a use case where the user would need to cancel a single operation while maintaining the other operations active. The user can always abandon all operations by calling `channel_2_reset`.
```
dcristoloveanu, line 25: >channel_2_reset [](http://example.com/codeflow?start=306&length=15)

chanell_2_close, right? :-)

dcristoloveanu, line 25: >We have not found a use case where the user would need to cancel a single operation while maintaining the other operations active. [](http://example.com/codeflow?start=120&length=130)

I don't think that cancelling operations is bad generally.
This being an attempt to a rather generic module, it sounds to me that the intent to have these operations nice and cancellable is commendable.

I can imagine a scenario where the waiters we use for read and wait would use channel for example.
So this flip flop between let's have cancellation or let's not have cancellation for individual operations seems a bit premature.

I think the right way would be to do the following:
- when the channel is closed, cancel all outstanding operations.
- switch state to closed.

Notice I am not using reset, as I think of reset being sugar for (close+open).

devdoc/channel_2.md
```diff
@@ -0,0 +1,72 @@
...
+
+3. Remove `channel_internal`: The current implementation of `channel` has a `channel_internal` layer that was made necessary because of the `ASYNC_OP`. This layer is not needed in `channel_2` since the `ASYNC_OP` is removed.
+
+4. Adding `open`/`close` APIs: These will allow the `channel_2` to guarantee that all pending operations are abandoned when the `channel_2` is closed.
```
dcristoloveanu, line 29: >`open`/`close` [](http://example.com/codeflow?start=10&length=14)

If we add open/close, why do we care to do reset?

devdoc/channel_2.md
```diff
@@ -0,0 +1,72 @@
...
+    The callback for `op3` is called with `data3` and `CHANNEL_CALLBACK_RESULT_OK`.
+
+    The receiver will now contain [`data1`, `data3`].
+
+    This order of the data on the receiver side of the `channel` is not the same as the order of the `PUSH` operations on the sender side. This is not desirable.
+
+    - Insufficient value: There is also not much value in providing the user with an `ASYNC_OP` for a single operation. We have not found a use case where the user would need to cancel a single operation while maintaining the other operations active. The user can always abandon all operations by calling `channel_2_reset`.
+
+3. Remove `channel_internal`: The current implementation of `channel` has a `channel_internal` layer that was made necessary because of the `ASYNC_OP`. This layer is not needed in `channel_2` since the `ASYNC_OP` is removed.
+
```
dcristoloveanu, line 28: I think we have discussed that channel_internal was done because we did not have proper open/close semantics on the channel.

I was expecting when I opened the review to see some design proposing to add open/close to channel in order to remove the channel internal. This looks like it wants that, but it also goes to do some extra item which we did not discuss - remove ASYNC_OP usage.

dcristoloveanu (changes requested):
:clock1:

devdoc/channel_2.md
BertKleewein, line 5: >channel_2` aims to address these pitfalls by making the following changes: [](http://example.com/codeflow?start=63&length=74)

Why did you decide to make a new API instead of fixing the existing one?  This doc makes it sound like we will have the original `channel.h` API and a new `channel_2.h` API and they will live side-by-side.  Will the old one ever be removed?  When?

This also makes me wonder how the existing code, for `channel_pull` and `channel_push` is related to the new code for `channel_2_pull` and `channel_2_push`? Are you going to completely scrap the existing implementation? Are you going to have some sort of shim later so `channel_2_push` internally calls `channel_push` or vise versa?

If `channel_2.h` is a complete rewrite, does that also mean a new set of requirements and tests for all of the existing functionality?

From what I see here, it looks like you're writing a completely new channel implementation instead of fixing the old one. Is this right? I hope not.

I would suggest making this into a set of incremental improvements to the existing channel, where each one is an isolated piece of work that can completed and verified before the next one is started. For example, you could remove `async_op` as step 1, finish it, propagate it up to geo and make the changes there, then go back and remove channel_internal, propagate it up, and so on.

devdoc/channel_2.md
```diff
@@ -0,0 +1,72 @@
...
+    The callback for `op3` is called with `data3` and `CHANNEL_CALLBACK_RESULT_OK`.
+
+    The receiver will now contain [`data1`, `data3`].
+
+    This order of the data on the receiver side of the `channel` is not the same as the order of the `PUSH` operations on the sender side. This is not desirable.
+
+    - Insufficient value: There is also not much value in providing the user with an `ASYNC_OP` for a single operation. We have not found a use case where the user would need to cancel a single operation while maintaining the other operations active. The user can always abandon all operations by calling `channel_2_reset`.
+
+3. Remove `channel_internal`: The current implementation of `channel` has a `channel_internal` layer that was made necessary because of the `ASYNC_OP`. This layer is not needed in `channel_2` since the `ASYNC_OP` is removed.
```
BertKleewein, line 27: >hat was made necessary because of the `ASYNC_OP` [](http://example.com/codeflow?start=102&length=48)

Huh? Why did async_op make this necessary? I thought you added channel_internal because there was no open/close state, and you needed the internal object so you could treat exterrnal references (from calling code) differently than you treat internal references (from async_op).

I don't think it actually matters why you channel_internal. I just don't think it's _directly_ related to async_op.  It might be indirectly related, but I don't think that "remove async_op usage" and "remove channel_internal" are intrinsically bound.  You could do one, or you could do the other, or you could do both, but I don't think you have to do both of them at the same time.

devdoc/channel_2.md
BertKleewein, line 7: as written, `channel_2_reset` is a synchronous API, and I'm assuming that it can be called in SM_OPENED.  How does this API interact with other API calls?  Does it use `sm_exec_begin` or `sm_barrier_begin`? My worry here is that I don't know what happens to other operations while this reset operation is taking place?  If I call `channel_2_push` around the same time a different thread calls `c hannel_2_reset`, what happens?  I assume three are some cases where my push will be accepted, then the reset will start, and my push will be cancelled.  But what if the reset starts before the push is accepted?  What happens then?  Does my thread block until the reset is complete?  This could be a while. Or, does reset make a copy of the current "queue" so it can cancel the existing operations while the channel is accepting new ones?

This is all assuming that `c hannel_2_reset` doesn't return until after all of the `CHANNEL__2_PUSH_CALLBACK` and `CHANNEL_2_PULL_CALLBACK` calls have been called.  Is this guaranteed?  I assume so. OPtherwise this function only starts the reset process without letting the caller know when the reset process is complete.

devdoc/channel_2.md
```diff
@@ -0,0 +1,72 @@
+# channel_2 requirements:
+
+## Overview
+
+The current implementation of `channel` has certain pitfalls. `channel_2` aims to address these pitfalls by making the following changes:
+
+1. Introduce a `channel_2_reset` API: This API will allow the user to reset the `channel_2` to a clean state. It will cause all pending operations to be abandoned.
+
+2. Remove `ASYNC_OP`: The current implementation of `channel` provides the user with an `ASYNC_OP` when an operation (`channel_push` or `channel_pull`) is initiated. The user can call `async_op_cancel` to do a best-effort cancellation of the operation. There are 2 issues with this:
+    - Complicates object lifetimes: The lifetime of the operation must be linked to the `ASYNC_OP`and the `channel`, since each can be independently destroyed by the user.
```
BertKleewein, line 10: >each can be independently destroyed by the use [](http://example.com/codeflow?start=123&length=46)

I _really_ don't like the word "destroyed" here because channel and async_op are both THANDLEs and THANDLEs can't be "destroyed" by the user. They can be decref'ed , but the can't be destroyed.

This is why we are adding open & close. The user can close the channel, and they can decref the channel, but they cannot destroy the channel.

It may feel like I'm arguing semantics, but "close", "destroy", and "dispose" are 3 very different concepts, and we're in this mess because the existing design of channel combines them all together.

devdoc/channel_2.md
```diff
@@ -0,0 +1,72 @@
...
+#ifdef __cplusplus
+extern "C" {
+#endif /* __cplusplus */
+
+
+typedef struct CHANNEL_2_TAG CHANNEL_2;
+
+THANDLE_TYPE_DECLARE(CHANNEL_2);
+
+    MOCKABLE_FUNCTION(, THANDLE(CHANNEL_2), channel_2_create_and_open, THANDLE(PTR(LOG_CONTEXT_HANDLE)), log_context, THANDLE(THREADPOOL), threadpool);
```
BertKleewein, line 61: >channel_2_create_and_open [](http://example.com/codeflow?start=44&length=25)

please please please don't do this.  If we are adding an OPENED state, we  should add an _open function. Combining "create" and "open" into a single function may feel like you're making things "easier", but you're combining unrelated concepts into a single function, and that just leads to trouble down the road.

devdoc/channel_2.md
```diff
@@ -0,0 +1,72 @@
...
+    MOCKABLE_FUNCTION(, void, channel_2_close, THANDLE(CHANNEL_2), channel_2);
+    MOCKABLE_FUNCTION(, int, channel_2_reset, THANDLE(CHANNEL_2), channel_2);
+    MOCKABLE_FUNCTION(, CHANNEL_2_RESULT, channel_2_pull, THANDLE(CHANNEL_2), channel_2, THANDLE(RC_STRING), correlation_id, CHANNEL_2_PULL_CALLBACK, pull_callback, void*, pull_context);
+    MOCKABLE_FUNCTION(, CHANNEL_2_RESULT, channel_2_push, THANDLE(CHANNEL_2), channel_2, THANDLE(RC_STRING), correlation_id, THANDLE(RC_PTR), data, CHANNEL_2_PUSH_CALLBACK, push_callback, void*, push_context);
```
BertKleewein, line 65: >CHANNEL_2_PUSH_CALLBACK, push_callback [](http://example.com/codeflow?start=148&length=38)

I know this is old history, but I never understood why `channel_2_push` needs to use a callback. Why can't push be a synchronous operation? I can understand why `channel_2_pull` needs a callback, because it might need to wait for data to be ready. But what does `channel_2_push` need to wait for?

devdoc/channel_2.md
BertKleewein, line 5: >certain pitfalls [](http://example.com/codeflow?start=44&length=16)

there is another pitfall that isn't addressed here. With the current code, there is now way for the caller to know when the channel is actually closed.  The callbacks all happen in threadpool workers, but after you cancel an operation, there is no way to know when those workers are done. I talk about this in earlier comments, but I think it's really important to be very crisp about when functions like "cancel" and "close" return. If something queues a callback and immediately returns, then that is not crisp.

PR #473, opened 2025-01-08
C-UTIL:Removed threadpool_open and threadpool_close

nishudeshu:
Removing threadpool open and close from C-util

### DIFF
--- a/src/watchdog_threadpool.c
+++ b/src/watchdog_threadpool.c
@@ -15,7 +15,6 @@
 
 typedef struct WATCHDOG_THREADPOOL_TAG
 {
-    EXECUTION_ENGINE_HANDLE execution_engine;
     THANDLE(THREADPOOL) threadpool;
 } WATCHDOG_THREADPOOL;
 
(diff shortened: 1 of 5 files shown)


tests/watchdog_threadpool_ut/watchdog_threadpool_ut.c
```diff
@@ -33,18 +33,21 @@ static void on_umock_c_error(UMOCK_C_ERROR_CODE error_code)
     ASSERT_FAIL("umock_c reported error :%" PRI_MU_ENUM "", MU_ENUM_VALUE(UMOCK_C_ERROR_CODE, error_code));
 }
 
-static void expect_init_verbose(int threapool_open_result)
+static void expect_init_verbose(int threadpool_result)
```
dcristoloveanu, line 36: >int threadpool_result [](http://example.com/codeflow?start=32&length=21)

This would be better making it a "bool fail_threadpool_create"
It would be more accurately representing what is happening.

dcristoloveanu (dismissed):
:shipit:

src/watchdog_threadpool.c
```diff
@@ -67,27 +67,16 @@ IMPLEMENT_MOCKABLE_FUNCTION(, int, watchdog_threadpool_init)
...
-
-                    /*Codes_SRS_WATCHDOG_THREADPOOL_42_008: [ watchdog_threadpool_init shall return 0. ]*/
-                    result = 0;
-                    goto all_ok;
-                }
-                THANDLE_ASSIGN(THREADPOOL)(&threadpool, NULL);
+                /*Codes_SRS_WATCHDOG_THREADPOOL_42_006: [ watchdog_threadpool_init shall store the threadpool. ]*/
+                THANDLE_INITIALIZE_MOVE(THREADPOOL)(&g_watchdog.threadpool, &threadpool);
+
+                g_watchdog.execution_engine = execution_engine;
```
mattdurak, line 73: >g_watchdog.execution_engine = execution_engine; [](http://example.com/codeflow?start=16&length=47)

a long time ago threadpool did not do an inc_ref on the execution_engine, so this module had to do it itself so the execution engine didn't get freed while the threadpool was using it.

But now threadpool does correctly take a reference (inc_ref).

Since this module has no need for the execution_engine after handing it off to threadpool I think we could remove some more dead code by no longer storing this (instead moving the `execution_engine_dec_ref` up here and removing `execution_engine` from the `WATCHDOG_THREADPOOL` struct entirely.

/consider as another small cleanup while touching this.

mattdurak (dismissed):
:shipit:

tests/watchdog_threadpool_ut/watchdog_threadpool_ut.c
anporumb: renames, rewrote some logic here.

anporumb: also rearranged `{`, very important.

src/watchdog_threadpool.c
anporumb: now execution engine is dec_Ref'd only here.

dcristoloveanu (approved):
:shipit:

PR #476, opened 2025-01-09
Channel open close

parth21999:


tests/channel_int/channel_int.c
```diff
@@ -21,6 +21,9 @@
 #include "c_util/rc_ptr.h"
 #include "c_util/channel.h"
 
+
+#define DISABLE_TEST_FUNCTION(x) static void x(void)
```
dcristoloveanu, line 25: We have this in c test runner switcher now, so no need for it anymore.

src/channel.c
```diff
@@ -80,6 +78,48 @@ IMPLEMENT_MOCKABLE_FUNCTION(, THANDLE(CHANNEL), channel_create, THANDLE(PTR(LOG_
     return result;
 }
 
+IMPLEMENT_MOCKABLE_FUNCTION(, int, channel_open, THANDLE(CHANNEL), channel)
```
dcristoloveanu, line 81: >ION(, int, channel_open, THANDLE(CHANNEL), [](http://example.com/codeflow?start=24&length=42)

Why do we need channel_internal again? I thought that it was born BECAUSE we did not do open/close on the channel and <insert long explanation here>

Now that we do have open/close, why do we need this wrapper layer of channel over channel_internal?

tests/channel_int/channel_int.c
parth21999: Removed.

src/channel.c
parth21999: It can be removed, however it will lead to a big diff.

I wanted to send the smaller change for open/close first so that it is easier to review.

Removal of channel_internal will come in a separate PR later.

#30251622

devdoc/channel_internal.md
```diff
@@ -119,6 +142,8 @@ MOCKABLE_FUNCTION(, CHANNEL_RESULT, channel_internal_push, THANDLE(CHANNEL_INTER
 
  - **SRS_CHANNEL_INTERNAL_43_113: [** `channel_internal_pull` shall call `threadpool_schedule_work` with `execute_callbacks` as `work_function` and the obtained `operation` as `work_function_context`. **]**
 
+ - **SRS_CHANNEL_INTERNAL_43_163: [** If `threadpool_schedule_work` fails, `channel_internal_pull` shall terminate the process. **]**
+
```
BertKleewein, line 146: This seems a bit extreme.  Shouldn't we handle this like any other failure?

BertKleewein: same question for all the "shall terminate the process" that you added.

devdoc/channel_internal.md
```diff
@@ -191,8 +226,10 @@ MOCKABLE_FUNCTION(, CHANNEL_RESULT, channel_internal_push, THANDLE(CHANNEL_INTER
 
 `execute_callbacks` is the work function that is passed to `threadpool_schedule_work` when scheduling the execution of the callbacks for an operation.
 
-**SRS_CHANNEL_INTERNAL_43_148: [** If `channel_internal_op_context` is `NULL`, `execute_callbacks` shall fail. **]**
+**SRS_CHANNEL_INTERNAL_43_148: [** If `channel_internal_op_context` is `NULL`, `execute_callbacks` shall terminate the process. **]**
```
BertKleewein, line 229: This "shall terminate the process" is a different story though...

devdoc/channel_internal.md
```diff
@@ -191,8 +226,10 @@ MOCKABLE_FUNCTION(, CHANNEL_RESULT, channel_internal_push, THANDLE(CHANNEL_INTER
...
 
 **SRS_CHANNEL_INTERNAL_43_145: [** `execute_callbacks` shall call the stored callback(s) with the `result` of the `operation`.  **]**
 
+**SRS_CHANNEL_INTERNAL_43_157: [** `execute_callbacks` shall call `sm_exec_end` for each call callback. **]**
```
BertKleewein, line 233: >for each call callback [](http://example.com/codeflow?start=80&length=22)

poetry.  Do you mean "each called callback"? Or maybe better phrased as "each callback that is called."

devdoc/channel_internal.md
parth21999: `channel` guarantees that the data order is maintained. That means that the pull_callback for the `n`th pull operations will receive the data provided by the `n`th push operation.

If `threadpool_schedule_work` fails, then this guarantee is violated. This could cause the system to go in unpredictable states that would be hard to debug. We also don't have a good way to recover from a `threadpool_schedule_work` failure.

I think it's better if we terminate the process at that point. Open to other suggestions.

devdoc/channel_internal.md
```diff
@@ -191,8 +226,10 @@ MOCKABLE_FUNCTION(, CHANNEL_RESULT, channel_internal_push, THANDLE(CHANNEL_INTER
 
 `execute_callbacks` is the work function that is passed to `threadpool_schedule_work` when scheduling the execution of the callbacks for an operation.
 
-**SRS_CHANNEL_INTERNAL_43_148: [** If `channel_internal_op_context` is `NULL`, `execute_callbacks` shall fail. **]**
+**SRS_CHANNEL_INTERNAL_43_148: [** If `channel_internal_op_context` is `NULL`, `execute_callbacks` shall terminate the process. **]**
 
 **SRS_CHANNEL_INTERNAL_43_145: [** `execute_callbacks` shall call the stored callback(s) with the `result` of the `operation`.  **]**
```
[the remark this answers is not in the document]
parth21999: >one operation can have multiple callbacks?  

Yes, there is the `PUSH_CALLBACK` and the `PULL_CALLBACK`. A pair of `channel_pull` and `channel_push` calls form one "operation".

The `n`th pull call is matched to the `n`th push call and both the callbacks are stored in one operation context.

If there are more pull calls than push calls and the channel is closed, the pending pull callbacks are called with ABANDON, but no push callbacks are called since they were never provided.

>Ah, the "push_callback" isn't called until the pull is complete?  That seems like a really deceptive name then. 

`PUSH_CALLBACK` is intended to be read as "the callback provided by `channel_push`". Open to other names if you have a better suggestion.

devdoc/channel_internal.md
parth21999: Changed to "for each callback that is called"

devdoc/channel_internal.md
BertKleewein: Since we have one process per node, this brings down all replicas on the node. It should be able to fault (or fail or crash or burn down or destroy) a single replica without faulting all replicas.

BertKleewein: I'm not suggesting that we ignore the `threadpool_schedule_work` failure.  We can still return the failure to the caller and the caller can decide what to do. That might be failing the current replicate call, or it might be faulting this replica, or it might be something else. 

The guarantee is only violated if the caller ignores the failure. Even then, it's only a single channel that goes into an unpredictable state.  I'm not sure what you mean by "we don't have a good way to recover from a threadpool_schedule_work failure".  We might not be able to "repair" the channel, but we can limit the blast radius of the failure.  Terminating the process is the opposite of limiting the blast radius.

another way to look at it: how is a `threadpool_schedule_work` failure different than a `malloc` failure?  Would we terminate the process here if `malloc` failed?

BertKleewein:
> `CHANNEL_CALLBACK_RESULT_ABANDONED`: The callbacks are called with this result when the when the module is disposing.

when the channel is closed.

---
Refers to: devdoc/channel.md:52 in d9f6d4b. [](commit_id = d9f6d4b4a580a74238a0db5d3880b7ea634896a1, deletion_comment = False)

BertKleewein:
> The value of the `CHANNEL_CALLBACK_RESULT` stored in the `operation` can only change in the direction shown above.  If the `operation` is abandoned, it cannot be cancelled.

Schrodinger's callback? How can the result be in two states?  I think this says that a operation that is OK can still be cancelled, and then after it's cancelled, it can be abandoned.  

---
Refers to: devdoc/channel.md:67 in d9f6d4b. [](commit_id = d9f6d4b4a580a74238a0db5d3880b7ea634896a1, deletion_comment = False)

BertKleewein:
> `CHANNEL_CALLBACK_RESULT_ABANDONED` trumps `CHANNEL_CALLBACK_RESULT_CANCELLED` trumps `CHANNEL_CALLBACK_RESULT_OK`.

I don't get the verb "trumps" here. I assume you're using the card-game definition, and we might get the same callback with multiple results?  

---
Refers to: devdoc/channel.md:58 in d9f6d4b. [](commit_id = d9f6d4b4a580a74238a0db5d3880b7ea634896a1, deletion_comment = False)

devdoc/channel_internal.md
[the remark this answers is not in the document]
BertKleewein: maybe `on_data_consumed` or `on_operation_consumed` for the push_callback, and `on_operation_available` for the pull_callback?  Just a suggestion.

devdoc/channel.md
```diff
@@ -6,21 +6,24 @@
...
-    E: OP_QUEUE=[]
+    B: CREATED
+    O: OPEN, OP_QUEUE=[]
+    N: OPEN,OP_QUEUE=[push_op,...]
+    R: OPEN,OP_QUEUE=[pull_op,...]
+    E: OPEN,OP_QUEUE=[]
+    C: CLOSED, OP_QUEUE=[]
 
     [*] --> B: channel_create
+    B --> O: channel_open
```
BertKleewein, line 23: >   B --> O: channel_open [](http://example.com/codeflow?start=1&length=24)

Your geo_core change list takes advantage of the closed-> opened transition, but that's missing here.

BertKleewein:
> // Copyright (c) Microsoft. All rights reserved.

I feel like there are some missing scenarios that involve using, closing, re-opening, and then continuing to use a channel.

---
Refers to: tests/channel_int/channel_int.c:1 in d9f6d4b. [](commit_id = d9f6d4b4a580a74238a0db5d3880b7ea634896a1, deletion_comment = False)

devdoc/channel_internal.md
parth21999: Agreed, changing to return error

devdoc/channel_internal.md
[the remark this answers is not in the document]
parth21999: Renamed to `on_data_consumed_cb` and `on_data_available_cb`

parth21999:
> `CHANNEL_CALLBACK_RESULT_ABANDONED` trumps `CHANNEL_CALLBACK_RESULT_CANCELLED` trumps `CHANNEL_CALLBACK_RESULT_OK`.

This statement was meant to address the cases where closing/cancelling/completing are happening concurrently. 

It states the priority of results if an operation can complete with multiple results.

Now that channel uses `sm`, closing the channel will not race with cancelling/completing (since `sm_close_begin_with_cb` will wait for pending ops). 

The only race condition left is between cancelling and completing. I have added the following line to address that case:

"""
In the case when an `operation`'s callbacks are scheduled to be called with `CHANNEL_CALLBACK_RESULT_OK` but the operation is cancelled before they can execute, the callbacks are called with `CHANNEL_CALLBACK_RESULT_CANCELLED`.
"""


---
In reply to: [2622503322](https://github.com/Azure/c-util/pull/476#issuecomment-2622503322) [](http://example.com/codeflow?ancestors=2622503322)

---
Refers to: devdoc/channel.md:58 in 577eb10. [](commit_id = 577eb1067fc60bbd7e7278aa96e6d5870d4124c0, deletion_comment = False)

parth21999:
> The value of the `CHANNEL_CALLBACK_RESULT` stored in the `operation` can only change in the direction shown above.  If the `operation` is abandoned, it cannot be cancelled.

Explained in other comment.

---
In reply to: [2622493420](https://github.com/Azure/c-util/pull/476#issuecomment-2622493420) [](http://example.com/codeflow?ancestors=2622493420)

---
Refers to: devdoc/channel.md:67 in 577eb10. [](commit_id = 577eb1067fc60bbd7e7278aa96e6d5870d4124c0, deletion_comment = False)

parth21999:
> The value of the `CHANNEL_CALLBACK_RESULT` stored in the `operation` can only change in the direction shown above.  If the `operation` is abandoned, it cannot be cancelled.

Added more detail.

This diagram is not accurate now that channel has SM

---
In reply to: [2622493420](https://github.com/Azure/c-util/pull/476#issuecomment-2622493420) [](http://example.com/codeflow?ancestors=2622493420)

---
Refers to: devdoc/channel.md:67 in 577eb10. [](commit_id = 577eb1067fc60bbd7e7278aa96e6d5870d4124c0, deletion_comment = False)

devdoc/channel.md
parth21999: Added it

parth21999:
> // Copyright (c) Microsoft. All rights reserved.

Added int tests to cover those scenarious

---
In reply to: [2622681106](https://github.com/Azure/c-util/pull/476#issuecomment-2622681106) [](http://example.com/codeflow?ancestors=2622681106)

---
Refers to: tests/channel_int/channel_int.c:1 in 577eb10. [](commit_id = 577eb1067fc60bbd7e7278aa96e6d5870d4124c0, deletion_comment = False)

src/channel.c
```diff
@@ -80,52 +78,94 @@ IMPLEMENT_MOCKABLE_FUNCTION(, THANDLE(CHANNEL), channel_create, THANDLE(PTR(LOG_
...
+IMPLEMENT_MOCKABLE_FUNCTION(, void, channel_close, THANDLE(CHANNEL), channel)
+{
+    /*Codes_SRS_CHANNEL_43_097: [If channel is NULL, channel_close shall return immediately.]*/
+    if (channel == NULL)
+    {
+        LogError("Invalid arguments: THANDLE(CHANNEL) channel=%p", channel);
+    }
+    else
+    {
+        CHANNEL* channel_ptr = THANDLE_GET_T(CHANNEL)(channel);
```
BertKleewein, line 117: >CHANNEL* channel_ptr = THANDLE_GET_T(CHANNEL)(channel); [](http://example.com/codeflow?start=8&length=55)

you shouldn't need the GET_T here, right?

tests/channel_ut/channel_ut.c
```diff
@@ -273,6 +294,85 @@ TEST_FUNCTION(channel_dispose_succeeds)
 
 }
 
+/* channel_open */
+
+/*Tests_SRS_CHANNEL_43_095: [If channel is NULL, channel_open shall fail and return a non - zero value.]*/
```
BertKleewein, line 299: >Tests_SRS_CHANNEL_43_095 [](http://example.com/codeflow?start=2&length=24)

I think you're missing a test here

tests/channel_internal_ut/channel_internal_ut.c
```diff
@@ -26,6 +26,8 @@
 #include "c_pal/srw_lock.h"
 #include "c_pal/threadpool.h"
 #include "c_pal/thandle_log_context_handle.h"
+#include "c_pal/sm.h"
+#include "c_pal/log_critical_and_terminate.h"
```
BertKleewein, line 30: >log_critical_and_terminate [](http://example.com/codeflow?start=16&length=26)

not needed

devdoc/channel_internal.md
```diff
@@ -90,32 +109,36 @@ MOCKABLE_FUNCTION(, CHANNEL_RESULT, channel_internal_push, THANDLE(CHANNEL_INTER
...
 ### channel_internal_pull
 ```c
-    MOCKABLE_FUNCTION(, CHANNEL_RESULT, channel_internal_pull, THANDLE(CHANNEL_INTERNAL), channel_internal, THANDLE(RC_STRING), correlation_id, PULL_CALLBACK, pull_callback, void*, pull_context, THANDLE(ASYNC_OP)*, out_op_pull);
+    MOCKABLE_FUNCTION(, CHANNEL_RESULT, channel_internal_pull, THANDLE(CHANNEL_INTERNAL), channel_internal, THANDLE(RC_STRING), correlation_id, ON_DATA_AVAILABLE_CB, on_data_available_cb, void*, pull_context, THANDLE(ASYNC_OP)*, out_op_pull);
 ```
 
-`channel_internal_pull` registers the given `pull_callback` to be called when there is data to be consumed.
+`channel_internal_pull` registers the given `on_data_available_cb` to be called when there is data to be consumed.
+
+**SRS_CHANNEL_INTERNAL_43_152: [** `channel_internal_pull` shall call `sm_exec_begin`. **]**
```
BertKleewein, line 122: I'm having serious misgivings about this style of sm use (calling `sm_exec_begin` in the sync function and `sm_exec_end` in the async callback.)

src/channel_internal.c
```diff
@@ -279,148 +353,166 @@ static int dequeue_operation(THANDLE(CHANNEL_INTERNAL) channel_internal, THANDLE
...
+                {
+                    /* Codes_SRS_CHANNEL_INTERNAL_43_011: [ channel_internal_pull shall succeeds and return CHANNEL_RESULT_OK. ]*/
+                    result = CHANNEL_RESULT_OK;
+                }
             }
+            /*Codes_SRS_CHANNEL_INTERNAL_43_115: [ channel_internal_pull shall call srw_lock_release_exclusive. ]*/
+            srw_lock_release_exclusive(channel_internal_ptr->lock);
         }
-
+        result == CHANNEL_RESULT_OK ? (void)0 : sm_exec_end(channel_internal_ptr->sm);
```
BertKleewein, line 454: Every time I read a line of code that looks like this, I die a little bit inside.

src/channel_internal.c
```diff
@@ -65,17 +68,15 @@ static void channel_internal_dispose(CHANNEL_INTERNAL* channel_internal)
...
 {
-    CHANNEL_INTERNAL* channel_internal_ptr = THANDLE_GET_T(CHANNEL_INTERNAL)(channel_internal);
-
-    /*Codes_SRS_CHANNEL_INTERNAL_43_094: [ channel_internal_close shall call srw_lock_acquire_exclusive. ]*/
-    srw_lock_acquire_exclusive(channel_internal_ptr->lock);
-
-    /*Codes_SRS_CHANNEL_INTERNAL_43_095: [ channel_internal_close shall iterate over the list of pending operations and do the following: ]*/
+    CHANNEL_INTERNAL* channel_internal_ptr = context;
+    /*Codes_SRS_CHANNEL_INTERNAL_43_095: [ abandon_pending_operations shall iterate over the list of pending operations and do the following: ]*/
     for (DLIST_ENTRY* entry = DList_RemoveHeadList(&channel_internal_ptr->op_list); entry != &channel_internal_ptr->op_list; entry = DList_RemoveHeadList(&channel_internal_ptr->op_list))
```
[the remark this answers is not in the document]
BertKleewein: The SM is not enough. At this point, it should be in `SM_OPENED_DRAINING_TO_CLOSE` and other threads could stilll be executing code that touches the `op_list`

src/channel_internal.c
```diff
@@ -16,6 +16,8 @@
 #include "c_pal/srw_lock.h"
 #include "c_pal/threadpool.h"
 #include "c_pal/thandle_log_context_handle.h"
+#include "c_pal/sm.h"
+#include "c_pal/log_critical_and_terminate.h"
```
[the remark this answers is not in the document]
BertKleewein: sorry, I was wrong.

src/channel_internal.c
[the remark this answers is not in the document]
BertKleewein: In addition to possibly corrupting the DLIST, this can also lead to callbacks being called multiple times.

src/channel_internal.c
```diff
@@ -279,148 +353,166 @@ static int dequeue_operation(THANDLE(CHANNEL_INTERNAL) channel_internal, THANDLE
...
     /*Codes_SRS_CHANNEL_INTERNAL_43_113: [ channel_internal_pull shall call threadpool_schedule_work with execute_callbacks as work_function and the obtained operation as work_function_context. ]*/
     /*Codes_SRS_CHANNEL_INTERNAL_43_129: [ channel_internal_push shall call threadpool_schedule_work with execute_callbacks as work_function and the obtained operation as work_function_context. ]*/
     if (threadpool_schedule_work(channel_internal_ptr->threadpool, execute_callbacks, channel_op) != 0)
     {
         LogError("Failure in threadpool_schedule_work(channel_internal_ptr->threadpool=%p, execute_callbacks=%p, channel_op=%p)", channel_internal_ptr->threadpool, execute_callbacks, channel_op);
-        // undo the dequeue
-        THANDLE_ASSIGN(ASYNC_OP)(out_op, NULL);
-        if (pull_callback != NULL)
+        result = MU_FAILURE;
+        // For testing purposes, we must unset the callback
```
BertKleewein, line 379: There's something you're not telling us here. I'm not against doing this, but the vague comment makes me wonder what you're hiding.  If this has to be here for a specific test or a specific check, then put the explanation here please.

src/channel_internal.c
```diff
@@ -279,148 +353,166 @@ static int dequeue_operation(THANDLE(CHANNEL_INTERNAL) channel_internal, THANDLE
...
     {
-        /*Codes_SRS_CHANNEL_INTERNAL_43_112: [ channel_internal_pull shall store the correlation_id, pull_callback and pull_context in the obtained operation. ]*/
+        /*Codes_SRS_CHANNEL_INTERNAL_43_112: [ channel_internal_pull shall store the correlation_id, on_data_available_cb and pull_context in the obtained operation. ]*/
         THANDLE_INITIALIZE(RC_STRING)(&channel_op->pull_correlation_id, pull_correlation_id);
-        channel_op->pull_callback = pull_callback;
+        channel_op->on_data_available_cb = on_data_available_cb;
         channel_op->pull_context = pull_context;
     }
-    else if (push_callback != NULL)
+    else if (on_data_consumed_cb != NULL)
```
BertKleewein, line 363: >else [](http://example.com/codeflow?start=4&length=4)

Uh oh, should this `else` be here.  ie - it looks like we want to ignore `on_data_consumed_cb` if `on_data_available_cb` is set?  Is this correct?  I _think_ this is wrong, but I can't really review the rest of this function until I get my head wrapped around this

src/channel_internal.c
```diff
@@ -279,148 +353,166 @@ static int dequeue_operation(THANDLE(CHANNEL_INTERNAL) channel_internal, THANDLE
...
     CHANNEL_OP* channel_op = CONTAINING_RECORD(op_entry, CHANNEL_OP, anchor);
-    if (pull_callback != NULL)
+    if (on_data_available_cb != NULL)
     {
-        /*Codes_SRS_CHANNEL_INTERNAL_43_112: [ channel_internal_pull shall store the correlation_id, pull_callback and pull_context in the obtained operation. ]*/
+        /*Codes_SRS_CHANNEL_INTERNAL_43_112: [ channel_internal_pull shall store the correlation_id, on_data_available_cb and pull_context in the obtained operation. ]*/
         THANDLE_INITIALIZE(RC_STRING)(&channel_op->pull_correlation_id, pull_correlation_id);
-        channel_op->pull_callback = pull_callback;
+        channel_op->on_data_available_cb = on_data_available_cb;
         channel_op->pull_context = pull_context;
```
BertKleewein, line 361: It looks like you're replacing the old callback with a new callback.  What happens with the old context?  Does that leak?

src/channel_internal.c
BertKleewein: Also, if on_data_available_cb is set, does push_context leak?

src/channel_internal.c
```diff
@@ -279,148 +353,166 @@ static int dequeue_operation(THANDLE(CHANNEL_INTERNAL) channel_internal, THANDLE
...
+        result = MU_FAILURE;
+        // For testing purposes, we must unset the callback
+        if (on_data_available_cb != NULL)
         {
-            channel_op->pull_callback = NULL;
-            channel_op->pull_context = NULL;
+            channel_op->on_data_available_cb = NULL;
         }
-        else if (push_callback != NULL)
+        else if (on_data_consumed_cb != NULL)
```
BertKleewein, line 384: >else [](http://example.com/codeflow?start=8&length=4)

same question about the `else` here.  If `on_data_availabe_cb` is set, do we mean to ignore hte value of `on_data_consumed_cb`?

src/channel_internal.c
```diff
@@ -279,148 +353,166 @@ static int dequeue_operation(THANDLE(CHANNEL_INTERNAL) channel_internal, THANDLE
...
-        THANDLE_ASSIGN(ASYNC_OP)(out_op, NULL);
-        if (pull_callback != NULL)
+        result = MU_FAILURE;
+        // For testing purposes, we must unset the callback
+        if (on_data_available_cb != NULL)
         {
-            channel_op->pull_callback = NULL;
-            channel_op->pull_context = NULL;
+            channel_op->on_data_available_cb = NULL;
         }
```
BertKleewein, line 383: seriously, what is going on here?  in the block above, you replace the old `on_data_availablre_cb` if a new one was passed in.  Here, if there's a failure, and you passed in a new callback, you set the stored one to null, which sounds like you're trying to un-do the replacement above.  I feel like I'm watching someone move shells around and I'm trying to figure out where the callback is being stored.

BertKleewein: [https://youtu.be/pXw7LYWNi5E?t=24](https://youtu.be/pXw7LYWNi5E?t=24)

src/channel_internal.c
```diff
@@ -36,8 +39,8 @@ THANDLE_TYPE_DEFINE(CHANNEL_INTERNAL);
 
 typedef struct CHANNEL_OP_TAG
 {
-    PUSH_CALLBACK push_callback;
-    PULL_CALLBACK pull_callback;
+    ON_DATA_CONSUMED_CB on_data_consumed_cb;
+    ON_DATA_AVAILABLE_CB on_data_available_cb;
     void* push_context;
     void* pull_context;
```
BertKleewein, line 45: the variable names for the contexts no longer match the variable names for the callbacks.  Should they?

src/channel.c
parth21999: Fixed.

tests/channel_internal_ut/channel_internal_ut.c
parth21999: We still have the LogCriticalAndTerminate from SRS_CHANNEL_INTERNAL_43_148

src/channel_internal.c
parth21999: Fixed

tests/channel_ut/channel_ut.c
parth21999: Fixed.

src/channel_internal.c
parth21999: This became unnecessary after adding open/close, removing.

src/channel_internal.c
parth21999: Removed.

src/channel_internal.c
parth21999: Removed

src/channel_internal.c
parth21999: The callback is being set for the first time. It was NULL before this.

src/channel_internal.c
parth21999: The helper function `dequeue_operation` is used in push and pull. This check is mutually exclusive, so the `else` can be there even if it is not strictly necessary.

src/channel_internal.c
parth21999: Open to suggestions

src/channel_internal.c
parth21999: In the block above, the callback is being set ( not replaced) based on whether `dequeue_operation` was called from `channel_internal_push` or `channel_internal_pull`. 

`execute_callbacks` uses the state of whether the callbacks are `NULL` or non-`NULL` to decide the clean up that needs to be performed. 

In the test case where `threadpool_schedule_work` fails, the cleanup is performed immediately. Therefore the callback must be unset so that the cleanup doesn't happen two times. Does that make sense?

src/channel_internal.c
parth21999: The `dequeue_operation` helper function is used by `channel_internal_pull` and `channel_internal_push`.

`channel_internal_pull` passes a valid `on_data_available_cb` and a `NULL` `on_data_consumed_cb`.

`channel_internal_push` passes a valid `on_data_consumed_cb` and a `NULL` `on_data_available_cb`.

The callbacks can be used to determine whether the caller is `channel_internal_pull` or `channel_internal_push`.

Since only one of the callbacks can be non-NULL and the other must be NULL, we will only execute one of the cases. The other one can be ignored.

src/channel_internal.c
parth21999: Added explanation

BertKleewein (approved):
:shipit:

PR #478, opened 2025-01-23
Made changes to move queue processor to c-util

abhinandan-codes:
### Summary
This pull request refactors queue_processor from zrpc lib to c-util along with needed dependencies

### Changes
Moved code from zrpc library to c-util.
Added needed dependencies
Made changes to cmake files
### Testing
Verified the functionality with unit tests to ensure the correct elapsed time is returned.


abhinandan-codes:
@microsoft-github-policy-service agree company="Microsoft"

CMakeLists.txt
```diff
@@ -77,6 +77,11 @@ if ((NOT TARGET c_pal) AND (EXISTS ${CMAKE_CURRENT_LIST_DIR}/deps/c-pal/CMakeLis
     add_subdirectory(deps/c-pal)
 endif()
 
+if ((NOT TARGET clds) AND (EXISTS ${CMAKE_CURRENT_LIST_DIR}/deps/clds/CMakeLists.txt))
+    add_subdirectory(deps/clds)
+    include_directories(${CMAKE_CURRENT_LIST_DIR}/deps/clds/inc)
+endif()
+
```
dcristoloveanu, line 84: Ugh, maybe the idea of moving to c_util was not as good :-(
Not sure what the reasoing for moving it might have been at the time the task was created. I see Riddhi created the task, so we should ask her if she has more details.

But, at the moment it is clear that moving this creates a circular dependency:
clds depends on c_util
c_util depends on clds.

This will not go well..

So the 2 solutions are:
- Do nothing, do not move queue processor. You should vet with Riddhi why we wanted to move it in the first place. If there is no clear reason why we are moving it, then we should not move it. I think even though we would not merge the change, it was a good exercise for you to go through this!
- Move also mpsc_queue to c_util. This feels a bit weird as we're moving things from the lockless structures repo to c_util, so it creates a bit of a hodge podge.

So please talk to Riddhi and let's skip this move if not needed

dcristoloveanu (changes requested):
:clock1:

CMakeLists.txt
ipg2013: Hey @dcristoloveanu, this task was created as part of this comment on read and wait PR - https://msazure.visualstudio.com/DefaultCollection/One/_git/Azure-MessagingStore/pullRequest/7570193#1675665426

We can skip this task.

dcristoloveanu:
This work is not needed

PR #479, opened 2025-01-31
Fixed tqueue_int tests

nishudeshu:
Fixed tqueue_int tests timeout and other issues where pusher_count and popper_count wasn't used.

### DIFF
--- a/tests/tqueue_int/tqueue_int.c
+++ b/tests/tqueue_int/tqueue_int.c
@@ -13,6 +13,8 @@
 #include "c_pal/gballoc_hl_redirect.h"
 
 #include "c_pal/interlocked.h"
+#include "c_pal/interlocked_hl.h"
+#include "c_pal/sync.h"
 #include "c_pal/thandle.h"
 #include "c_pal/threadapi.h"
 #include "c_pal/timer.h"
@@ -342,6 +344,7 @@ static int pusher_thread_func(void* arg)
         if (push_result == TQUEUE_PUSH_OK)
         {
             (void)interlocked_increment_64(&test_context->successful_push_count);
+            wake_by_address_single_64(&test_context->successful_push_count);
         }
         THANDLE_ASSIGN(TEST_THANDLE)(&item, NULL);
 
@@ -370,6 +373,7 @@ static int popper_thread_func(void* arg)
         if (pop_result == TQUEUE_POP_OK)
         {
             (void)interlocked_increment_64(&test_context->successful_pop_count);
+            wake_by_address_single_64(&test_context->successful_pop_count);
             THANDLE_ASSIGN(TEST_THANDLE)(&item, NULL);
         }
 
@@ -393,6 +397,11 @@ static void test_and_terminate_chaos_test(TQUEUE_CHAOS_TEST_THANDLE_CONTEXT *tes
         // get how many pushes and pops at the beginning of the time slice
         int64_t last_successful_push_count = interlocked_add_64(&test_context->successful_push_count, 0);
         int64_t last_successful_pop_count = interlocked_add_64(&test_context->successful_pop_count, 0);
+
+        // Ensures that the tests below for current_successful_push_count > last_successful_push_count and/or current_successful_pop_count > last_successful_pop_count
+        // do not fail due to multi-threading synchronization issues
+        InterlockedHL_WaitForNotValue64(&test_context->successful_push_count, last_successful_push_count, INT_MAX);
+        InterlockedHL_WaitForNotValue64(&test_context->successful_pop_count, last_successful_pop_count, INT_MAX);
 
         ThreadAPI_Sleep(TEST_CHECK_PERIOD);
 
@@ -444,9 +453,6 @@ static void TQUEUE_test_with_N_pushers_and_N_poppers_with_queue_size(uint32_t in
     }
 
     test_and_terminate_chaos_test(&test_context);
-
-    // terminate test
-    (void)interlocked_exchange(&test_context.terminate_test, 1);
 
     for (int i = 0; i < pusher_count; i++)
     {
@@ -526,6 +532,7 @@ static int tqueue_chaos_thread_THANDLE_func(void* arg)
             if (push_result == TQUEUE_PUSH_OK)
             {
                 (void)interlocked_increment_64(&test_context->successful_push_count);
+                wake_by_address_single_64(&test_context->successful_push_count);
             }
             THANDLE_ASSIGN(TEST_THANDLE)(&item, NULL);
             break;
@@ -539,6 +546,7 @@ static int tqueue_chaos_thread_THANDLE_func(void* arg)
             {
                 ASSERT_ARE_NOT_EQUAL(int64_t, -1, item->a_value);
                 (void)interlocked_increment_64(&test_context->successful_pop_count);
+                wake_by_address_single_64(&test_context->successful_pop_count);
                 THANDLE_ASSIGN(TEST_THANDLE)(&item, NULL);
             }
             break;
@@ -552,6 +560,7 @@ static int tqueue_chaos_thread_THANDLE_func(void* arg)
             {
                 ASSERT_ARE_NOT_EQUAL(int64_t, -1, item->a_value);
                 (void)interlocked_increment_64(&test_context->successful_pop_count);
+                wake_by_address_single_64(&test_context->successful_pop_count);
                 THANDLE_ASSIGN(TEST_THANDLE)(&item, NULL);
             }
             break;
@@ -561,6 +570,7 @@ static int tqueue_chaos_thread_THANDLE_func(void* arg)
             int64_t current_count = TQUEUE_GET_VOLATILE_COUNT(THANDLE(TEST_THANDLE))(test_context->queue);
             ASSERT_IS_TRUE(((current_count >= 0) && (current_count <= test_context->max_queue_size)));
             (void)interlocked_increment_64(&test_context->successful_get_volatile_count);
+            wake_by_address_single_64(&test_context->successful_get_volatile_count);
         }
         }
         current_time = timer_global_get_elapsed_ms();
--- a/tests/tqueue_int/tqueue_int.c
+++ b/tests/tqueue_int/tqueue_int.c
@@ -4,6 +4,7 @@
 #include <stdlib.h>
 #include <stdbool.h>
 #include <inttypes.h>
+#include <limits.h>
 
 #include "macro_utils/macro_utils.h"
 #include "testrunnerswitcher.h"
@@ -12,6 +13,8 @@
 #include "c_pal/gballoc_hl_redirect.h"
 
 #include "c_pal/interlocked.h"
+#include "c_pal/interlocked_hl.h"
+#include "c_pal/sync.h"
 #include "c_pal/thandle.h"
 #include "c_pal/threadapi.h"
 #include "c_pal/timer.h"
@@ -60,7 +63,7 @@ TEST_FUNCTION_CLEANUP(method_cleanup)
 {
 }
 
-DISABLED_TEST_FUNCTION(TQUEUE_CREATE_with_NULL_callbacks_succeds)
+TEST_FUNCTION(TQUEUE_CREATE_with_NULL_callbacks_succeds)
 {
     // arrange
 
@@ -74,7 +77,7 @@ DISABLED_TEST_FUNCTION(TQUEUE_CREATE_with_NULL_callbacks_succeds)
     TQUEUE_ASSIGN(FOO)(&queue, NULL);
 }
 
-DISABLED_TEST_FUNCTION(TQUEUE_PUSH_succeeds)
+TEST_FUNCTION(TQUEUE_PUSH_succeeds)
 {
     // arrange
     TQUEUE(FOO) queue = TQUEUE_CREATE(FOO)(16, 1024, NULL, NULL, NULL);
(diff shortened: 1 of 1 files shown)


nishudeshu:
https://msazure.visualstudio.com/One/_workitems/edit/30147387

tests/tqueue_int/tqueue_int.c
```diff
@@ -359,86 +388,106 @@ static int popper_thread_func(void* arg)
...
+    TQUEUE_CHAOS_TEST_THANDLE_CONTEXT test_context = { .queue = TQUEUE_CREATE(THANDLE(TEST_THANDLE))(initial_queue_size, max_queue_size, TEST_THANDLE_copy_item, TEST_THANDLE_dispose, NULL) };
+    ASSERT_IS_NOT_NULL(test_context.queue);
 
+    // count how many successful pushes and pops we have
+    (void)interlocked_exchange_64(&test_context.successful_push_count, 0);
+    (void)interlocked_exchange_64(&test_context.successful_pop_count, 0);
+    (void)interlocked_exchange_64(&test_context.successful_get_volatile_count, 0);
+    (void)interlocked_exchange(&test_context.terminate_test, 0);
+
+    THREAD_HANDLE *pusher_thread = malloc(sizeof(THREAD_HANDLE) * pusher_count);
```
dcristoloveanu, line 439: >pusher_thread [](http://example.com/codeflow?start=19&length=13)

Nit: pusher_threads (plural)

dcristoloveanu, line 439: >malloc [](http://example.com/codeflow?start=35&length=6)

Better to use malloc_2 when generally allocating n items of a given size.
malloc_2 checks for overflows, so a good habbit to be in.

tests/tqueue_int/tqueue_int.c
```diff
@@ -359,86 +388,106 @@ static int popper_thread_func(void* arg)
...
     {
-        int64_t last_expected_pop_number = interlocked_add_64(&test_context.next_expected_pop_number, 0);
+        // get how many pushes and pops at the beginning of the time slice
+        int64_t last_successful_push_count = interlocked_add_64(&test_context->successful_push_count, 0);
+        int64_t last_successful_pop_count = interlocked_add_64(&test_context->successful_pop_count, 0);
 
-        ThreadAPI_Sleep(TEST_CHECK_PERIOD);
+        // Ensures that the tests below for current_successful_push_count > last_successful_push_count and/or current_successful_pop_count > last_successful_pop_count
+        // do not fail due to multi-threading synchronization issues
+        InterlockedHL_WaitForNotValue64(&test_context->successful_push_count, last_successful_push_count, INT_MAX);
```
dcristoloveanu, line 403: >InterlockedHL_WaitForNotValue64 [](http://example.com/codeflow?start=8&length=31)

Likely this should assert the result to be OK.
Here and below.

tests/tqueue_int/tqueue_int.c
```diff
@@ -359,86 +388,106 @@ static int popper_thread_func(void* arg)
...
+        LogInfo("%.02f seconds elapsed, successful push count=%" PRId64 ", successful pop count=%" PRId64 ", successful get_count count=%" PRId64 "",
+            (current_time - start_time) / 1000, current_successful_push_count, current_successful_pop_count, current_successful_get_volatile_count);
+    } while (current_time - start_time < CHAOS_TEST_RUNTIME);
 
     // terminate test
-    (void)interlocked_exchange(&test_context.terminate_test, 1);
+    (void)interlocked_exchange(&test_context->terminate_test, 1);
+}
+
+static void TQUEUE_test_with_N_pushers_and_N_poppers_with_queue_size(uint32_t initial_queue_size, uint32_t max_queue_size, int pusher_count, int popper_count)
```
dcristoloveanu, line 427: >int [](http://example.com/codeflow?start=123&length=3)

This should be uint32_t (no need to use signed). We rarely use signed types when they are not needed.
Same for popper_count.

tests/tqueue_int/tqueue_int.c
```diff
@@ -359,86 +388,106 @@ static int popper_thread_func(void* arg)
...
+    (void)interlocked_exchange_64(&test_context.successful_get_volatile_count, 0);
+    (void)interlocked_exchange(&test_context.terminate_test, 0);
+
+    THREAD_HANDLE *pusher_thread = malloc(sizeof(THREAD_HANDLE) * pusher_count);
+    ASSERT_IS_NOT_NULL(pusher_thread);
+    THREAD_HANDLE *popper_thread = malloc(sizeof(THREAD_HANDLE) * popper_count);
+    ASSERT_IS_NOT_NULL(popper_thread);
+
+    // act
+    for (int i = 0; i < pusher_count; i++)
```
dcristoloveanu, line 445: >int [](http://example.com/codeflow?start=9&length=3)

uint32_t

tests/tqueue_int/tqueue_int.c
```diff
@@ -359,86 +388,106 @@ static int popper_thread_func(void* arg)
...
+    {
+        ASSERT_ARE_EQUAL(THREADAPI_RESULT, THREADAPI_OK, ThreadAPI_Create(&pusher_thread[i], pusher_thread_func, &test_context));
+    }
+
+    for (int i = 0; i < popper_count; i++)
```
dcristoloveanu, line 450: >int [](http://example.com/codeflow?start=9&length=3)

uint32_t

dcristoloveanu: Here and everywhere

dcristoloveanu: Or size_t

tests/tqueue_int/tqueue_int.c
```diff
@@ -461,8 +510,9 @@ MU_DEFINE_ENUM_STRINGS(QUEUE_ENTRY_STATE, QUEUE_ENTRY_STATE_VALUES);
 static int tqueue_chaos_thread_THANDLE_func(void* arg)
 {
     TQUEUE_CHAOS_TEST_THANDLE_CONTEXT* test_context = arg;
-
-    while (interlocked_add(&test_context->terminate_test, 0) == 0)
+    double start_time = timer_global_get_elapsed_ms();
+    double current_time = start_time;
+    while ((interlocked_add(&test_context->terminate_test, 0) == 0) && (current_time - start_time < CHAOS_TEST_RUNTIME))
     {
```
dcristoloveanu, line 516: A comment might be good around here to explain why we want each thread to check the time rather than wait for a signal.
This would prevent someone from switching back to signaling from another thread and reintroducing the same problem again.

tests/tqueue_int/tqueue_int.c
```diff
@@ -516,9 +569,11 @@ static int tqueue_chaos_thread_THANDLE_func(void* arg)
         {
             int64_t current_count = TQUEUE_GET_VOLATILE_COUNT(THANDLE(TEST_THANDLE))(test_context->queue);
             ASSERT_IS_TRUE(((current_count >= 0) && (current_count <= test_context->max_queue_size)));
-            (void)interlocked_increment_64(&test_context->succesful_get_volatile_count);
+            (void)interlocked_increment_64(&test_context->successful_get_volatile_count);
+            wake_by_address_single_64(&test_context->successful_get_volatile_count);
         }
```
dcristoloveanu, line 574: This wake by address is not really needed. wake by address is only needed when someone is actually waiting on the address

dcristoloveanu (approved):
:shipit:

tests/tqueue_int/tqueue_int.c
nishudeshu: Done

tests/tqueue_int/tqueue_int.c
nishudeshu: Fixed

tests/tqueue_int/tqueue_int.c
nishudeshu: Fixed

tests/tqueue_int/tqueue_int.c
nishudeshu: Fixed

tests/tqueue_int/tqueue_int.c
nishudeshu: Fixed

tests/tqueue_int/tqueue_int.c
nishudeshu: Fixed

tests/tqueue_int/tqueue_int.c
nishudeshu: Fixed

dcristoloveanu (dismissed):
:shipit:

tests/tqueue_int/tqueue_int.c
```diff
@@ -359,86 +393,106 @@ static int popper_thread_func(void* arg)
...
     {
-        int64_t last_expected_pop_number = interlocked_add_64(&test_context.next_expected_pop_number, 0);
+        // get how many pushes and pops at the beginning of the time slice
+        int64_t last_successful_push_count = interlocked_add_64(&test_context->successful_push_count, 0);
+        int64_t last_successful_pop_count = interlocked_add_64(&test_context->successful_pop_count, 0);
 
-        ThreadAPI_Sleep(TEST_CHECK_PERIOD);
+        // Ensures that the tests below for current_successful_push_count > last_successful_push_count and/or current_successful_pop_count > last_successful_pop_count
+        // do not fail due to multi-threading synchronization issues
+        ASSERT_IS_TRUE(INTERLOCKED_HL_OK == InterlockedHL_WaitForNotValue64(&test_context->successful_push_count, last_successful_push_count, INT_MAX));
```
mattdurak, line 408: >ASSERT_IS_TRUE(INTERLOCKED_HL_OK  [](http://example.com/codeflow?start=8&length=33)

nit: these can be `ASSERT_ARE_EQUAL(INTERLOCKED_HL_RESULT, INTERLOCKED_HL_OK, InterlockedHL_WaitFor...)`

Advantage is the logging will print the result if it doesn't match.

Will need a `TEST_DEFINE_ENUM_TYPE` if we don't have it already

tests/tqueue_int/tqueue_int.c
```diff
@@ -335,21 +361,29 @@ static int pusher_thread_func(void* arg)
...
-        TQUEUE_POP_RESULT pop_result = TQUEUE_POP(FOO)(test_context->queue, &item, NULL, NULL, NULL);
-        ASSERT_IS_TRUE((pop_result == TQUEUE_POP_OK) || (pop_result == TQUEUE_POP_QUEUE_EMPTY), "TQUEUE_POP(FOO) failed with %" PRI_MU_ENUM "", MU_ENUM_VALUE(TQUEUE_POP_RESULT, pop_result));
+        THANDLE(TEST_THANDLE) item = NULL;
+        TQUEUE_POP_RESULT pop_result = TQUEUE_POP(THANDLE(TEST_THANDLE))(test_context->queue, &item, NULL, NULL, NULL);
+        ASSERT_IS_TRUE((pop_result == TQUEUE_POP_OK) || (pop_result == TQUEUE_POP_QUEUE_EMPTY), "TQUEUE_POP(THANDLE(TEST_THANDLE)) failed with %" PRI_MU_ENUM "", MU_ENUM_VALUE(TQUEUE_POP_RESULT, pop_result));
         if (pop_result == TQUEUE_POP_OK)
         {
-            int64_t expected_pop_number = interlocked_increment_64(&test_context->next_expected_pop_number) - 1;
-            ASSERT_ARE_EQUAL(int, expected_pop_number, item.x);
+            (void)interlocked_increment_64(&test_context->successful_pop_count);
```
mattdurak, line 380: >(void)interlocked_increment_64(&test_context->successful_pop_count); [](http://example.com/codeflow?start=12&length=68)

So I know we can't validate the "next expected" number when there are multiple threads, but we should at least make sure we received a non-NULL `item` and maybe keep the check that the "x" is not "-1".

tests/tqueue_int/tqueue_int.c
```diff
@@ -359,86 +393,106 @@ static int popper_thread_func(void* arg)
...
     // assert
 
     // clean
-    TQUEUE_ASSIGN(FOO)(&test_context.queue, NULL);
+    free(pusher_threads);
+    free(popper_threads);
+    TQUEUE_ASSIGN(THANDLE(TEST_THANDLE))(&test_context.queue, NULL);
 }
 
 // This test has one pusher and one popper and validates the fact that order is preserved in this case
```
mattdurak, line 482: > and validates the fact that order is preserved in this case [](http://example.com/codeflow?start=42&length=60)

In case of 1 pusher and 1 popper we CAN validate the order is preserved (because there are no races between multiple pushers assigning a number and pushing and no races between multiple poppers pulling and checking the number.

But we aren't.

Maybe we should? We could have a flag set and `popper_threads_func` could check that the sequence number is correct, as it was before:

```
            int64_t expected_pop_number = interlocked_increment_64(&test_context->next_expected_pop_number) - 1;
            ASSERT_ARE_EQUAL(int, expected_pop_number, item.x);
```

mattdurak (approved):
:shipit:

tests/tqueue_int/tqueue_int.c
nishudeshu: Fixed

tests/tqueue_int/tqueue_int.c
nishudeshu: I had used that earlier and had to change everything substantially because valgrind and helgrind were throwing data race condition errors and no lock present error

PR #481, opened 2025-02-01
tqueue_int additional changes

nishudeshu:
Addressing few comments that couldn't be merged earlier.

### DIFF
--- a/tests/tqueue_int/tqueue_int.c
+++ b/tests/tqueue_int/tqueue_int.c
@@ -44,6 +44,7 @@ TQUEUE_TYPE_DEFINE(THANDLE(TEST_THANDLE));
 TEST_DEFINE_ENUM_TYPE(THREADAPI_RESULT, THREADAPI_RESULT_VALUES);
 TEST_DEFINE_ENUM_TYPE(TQUEUE_PUSH_RESULT, TQUEUE_PUSH_RESULT_VALUES);
 TEST_DEFINE_ENUM_TYPE(TQUEUE_POP_RESULT, TQUEUE_POP_RESULT_VALUES);
+TEST_DEFINE_ENUM_TYPE(INTERLOCKED_HL_RESULT, INTERLOCKED_HL_RESULT_VALUES);
 
 BEGIN_TEST_SUITE(TEST_SUITE_NAME_FROM_CMAKE)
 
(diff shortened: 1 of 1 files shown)


tests/tqueue_int/tqueue_int.c
```diff
@@ -643,9 +645,9 @@ TEST_FUNCTION(MU_C3(TQUEUE_chaos_knight_test_with_THANDLE_queue_size_16_and_1_pu
     TQUEUE_test_with_N_pushers_and_N_poppers_with_queue_size(16, 16, 1, N_THREADS);
 }
 
-TEST_FUNCTION(MU_C5(TQUEUE_chaos_knight_test_with_THANDLE_queue_size_16_and_, N_THREADS, _pushers_and_, N_THREADS, _poppers))
+TEST_FUNCTION(MU_C5(TQUEUE_chaos_knight_test_with_THANDLE_queue_size_16_and_, 8, _pushers_and_, 8, _poppers))
 {
-    TQUEUE_test_with_N_pushers_and_N_poppers_with_queue_size(16, 16, N_THREADS, N_THREADS);
+    TQUEUE_test_with_N_pushers_and_N_poppers_with_queue_size(16, 16, 8, 8);
```
dcristoloveanu, line 650: > 8, 8); [](http://example.com/codeflow?start=68&length=7)

Why is this change needed? Something is not being told here...

tests/tqueue_int/tqueue_int.c
```diff
@@ -377,6 +378,7 @@ static int popper_threads_func(void* arg)
         ASSERT_IS_TRUE((pop_result == TQUEUE_POP_OK) || (pop_result == TQUEUE_POP_QUEUE_EMPTY), "TQUEUE_POP(THANDLE(TEST_THANDLE)) failed with %" PRI_MU_ENUM "", MU_ENUM_VALUE(TQUEUE_POP_RESULT, pop_result));
         if (pop_result == TQUEUE_POP_OK)
         {
+            ASSERT_ARE_NOT_EQUAL(int64_t, item->a_value, -1);
```
dcristoloveanu, line 381: >  ASSERT_ARE_NOT_EQUAL(int64_t, item->a_value, -1); [](http://example.com/codeflow?start=10&length=51)

Hmmm, a wild assert... what is the assert really trying to validate here?
Sounds like it attempts to check that a_value was never initialized with -1 ... why would that even happen?

dcristoloveanu (changes requested):
:clock1:

tests/tqueue_int/tqueue_int.c
nishudeshu: It just makes the test faster since the Helgrind can still get stuck by 32 threads if we do 16x16 mostly due to yielding issues when it runs with several other tests on that server

tests/tqueue_int/tqueue_int.c
nishudeshu: Ahh, it was recommended by Matt, but looks like this test is remnant of the non THANDLE days of this push and pop test aka using `FOO` instead of `TEST_THANDLE`. So changed it to test for NULL

tests/tqueue_int/tqueue_int.c
dcristoloveanu: Ok ... , then the nice clean way would be to change the N_THREADS to be less with running on Linux with helgrind/valdgrind.

The point of the define was to have the number of threads in a define and not repeated all over.

/consider

dcristoloveanu (approved):
:shipit:

PR #492, opened 2025-03-03
strings_ut.c can overflow buffer if really big integers are passed (unlikely).

darobs:
We can potentially overflow the local string buffers if we pass a big enough integer to sprintf. ("STRING_new failure in test %lu/%lu" can be up to 69 bytes, and we were giving it 64.)

This is not likely (all the values should be < 3 digits), but gcc version 13.3.0 is reporting this as an error.

### DIFF
--- a/tests/strings_ut/strings_ut.c
+++ b/tests/strings_ut/strings_ut.c
@@ -35,6 +35,7 @@ static const char* MODIFIED_STRING_VALUE2 = "*nitial_";
 
 #define NUMBER_OF_CHAR_TOCOPY           8
 #define TEST_INTEGER_VALUE              1234
+#define DEFAULT_STRING_BUFFER_SIZE      128
 
 static const struct JSONEncoding {
     const char* source;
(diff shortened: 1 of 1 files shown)


tests/strings_ut/strings_ut.c
```diff
@@ -134,7 +134,7 @@ BEGIN_TEST_SUITE(TEST_SUITE_NAME_FROM_CMAKE)
         count = umock_c_negative_tests_call_count();
         for (index = 0; index < count; index++)
         {
-            char tmp_msg[64];
+            char tmp_msg[128];
```
jebrando, line 137: >128 [](http://example.com/codeflow?start=25&length=3)

/consider

You might want to add a define for this so you don't have to go through all them again if for some chance.

jebrando (approved):
:shipit:

tests/strings_ut/strings_ut.c
darobs: Good point, will do.

jebrando (approved):
:shipit:

PR #497, opened 2025-03-24
Move channel abandon callbacks out of lock

parth21999:


src/channel_internal.c
```diff
@@ -84,20 +84,20 @@ static void abandon_pending_operations(void* context)
...
+        /*Codes_SRS_CHANNEL_INTERNAL_43_169: [ abandon_pending_operations shall call srw_lock_release_exclusive. ]*/
+        srw_lock_release_exclusive(channel_internal_ptr->lock);
+    }
 
-        /*Codes_SRS_CHANNEL_INTERNAL_43_095: [ abandon_pending_operations shall iterate over the list of pending operations and do the following: ]*/
-        for (DLIST_ENTRY* entry = DList_RemoveHeadList(&channel_internal_ptr->op_list); entry != &channel_internal_ptr->op_list; entry = DList_RemoveHeadList(&channel_internal_ptr->op_list))
-        {
-            CHANNEL_OP* channel_op = CONTAINING_RECORD(entry, CHANNEL_OP, anchor);
+    /*Codes_SRS_CHANNEL_INTERNAL_43_095: [ abandon_pending_operations shall iterate over the list of pending operations and do the following: ]*/
+    for (DLIST_ENTRY* entry = DList_RemoveHeadList(&channel_internal_ptr->op_list); entry != &channel_internal_ptr->op_list; entry = DList_RemoveHeadList(&channel_internal_ptr->op_list))
```
BertKleewein, line 92: >for (DLIST_ENTRY* entry = DList_RemoveHeadList(&channel_internal_ptr->op_list); entry != &channel_internal_ptr->op_list; entry = DList_RemoveHeadList(&channel_internal_ptr->op_list)) [](http://example.com/codeflow?start=4&length=182)

Now you're using the DLIST outside of the lock.  This _might_ be safe, but it is very fragile. :confused:

BertKleewein: I say very fragile because using a lock to protect op_list in all cases except one is the same as not using a lock.  There is now a possible code path where op_list gets corrupted by multithreaded access.  You can try to imagine every possible interleaving of code to prove that using `is_open` is good enough in this case, but:
1. Relying on is_open is never going to be as good as using a lock.
2. The results of that exercise are invalidated as soon as the code changes.

BertKleewein (changes requested):
:clock1:

src/channel_internal.c
parth21999: In the next iteration, I made the change to make a local copy under lock and nulling out the original list.

src/channel_internal.c
```diff
@@ -78,27 +78,36 @@ static void abandon_pending_operations(void* context)
...
+    DList_InitializeListHead(&op_list);
     /*Codes_SRS_CHANNEL_INTERNAL_43_167: [abandon_pending_operations shall call srw_lock_acquire_exclusive.]*/
     srw_lock_acquire_exclusive(channel_internal_ptr->lock);
     {
         /*Codes_SRS_CHANNEL_INTERNAL_43_168: [abandon_pending_operations shall set is_open to false.]*/
         channel_internal_ptr->is_open = false;
 
+        /*Codes_SRS_CHANNEL_INTERNAL_43_174: [abandon_pending_operations shall make a local copy of the list of pending operations.]*/
+        DList_AppendTailList(&op_list, &channel_internal_ptr->op_list);
+        (void)DList_RemoveEntryList(&channel_internal_ptr->op_list);
```
dcristoloveanu, line 92: For my understanding, what does the RemoveEntryList do here?

parth21999: The aim here is to make a local copy of the list in `channel_internal_ptr->op_list` and set `channel_internal_ptr->op_list` to be empty.

There is no direct API that achieves this. My solution is to append the list to the local `op_list`, (this includes the head of `channel_internal_ptr->op_list`) and then remove the head node so the local `op_list` becomes the head.

parth21999: state: (op_list = [], channel_internal_ptr->op_list = [...ops..])
`DList_AppendTailList(&op_list, &channel_internal_ptr->op_list);`
state: (op_list = [channel_internal_ptr-op_list, ...ops...], channel_internal_ptr->op_list = [...ops..])
```
(void)DList_RemoveEntryList(&channel_internal_ptr->op_list);
DList_InitializeListHead(&channel_internal_ptr->op_list);
```

state: (op_list = [...ops...], channel_internal_ptr->op_list = [])

parth21999: Removing this in next iteration and copying the list entry-by-entry instead

src/channel_internal.c
```diff
@@ -74,31 +74,48 @@ static void channel_internal_dispose(CHANNEL_INTERNAL* channel_internal)
...
+    DLIST_ENTRY op_list;
+    DList_InitializeListHead(&op_list);
     /*Codes_SRS_CHANNEL_INTERNAL_43_167: [abandon_pending_operations shall call srw_lock_acquire_exclusive.]*/
     srw_lock_acquire_exclusive(channel_internal_ptr->lock);
     {
         /*Codes_SRS_CHANNEL_INTERNAL_43_168: [abandon_pending_operations shall set is_open to false.]*/
         channel_internal_ptr->is_open = false;
 
+        /*Codes_SRS_CHANNEL_INTERNAL_43_174: [abandon_pending_operations shall make a local copy of the list of pending operations.]*/
+        (void)DList_ForEach(&channel_internal_ptr->op_list, copy_entry, &op_list);
```
BertKleewein, line 99: >DList_ForEach [](http://example.com/codeflow?start=14&length=13)

This looks like it iteratively adding all of the items to the new list without properly removing them from the source list. 

It looks like this corrupting the source list as it copies them into the new list. 

It's assuming that the iteration over the source list will continue to work as you're destroying/modifying/corrupting the source list.

I am shocked and impressed that this happens to work correctly.

parth21999: I can add a `DList_RemoveEntryList` before `DList_InsertTailList(op_list, listEntry)` (line 81) if that makes things more palatable.

Modifying/destroying entries during `DList_ForEach` is explicitly support as per:

`**SRS_DLIST_24_001: [** `DList_ForEach` shall allow deletion of elements during traversal. **]****SRS_DLIST_24_001: [** `DList_ForEach` shall allow deletion of elements during traversal. **]**`

BertKleewein (approved):
:shipit:

PR #521, opened 2025-08-20
draft idle POC

BertKleewein:


src/channel.c
```diff
@@ -522,46 +526,57 @@ IMPLEMENT_MOCKABLE_FUNCTION(, CHANNEL_RESULT, channel_push, THANDLE(CHANNEL), ch
...
             }
             else
             {
-                /*Codes_SRS_CHANNEL_43_117: [ If the list of pending operations is empty or the first operation in the list of pending operations contains a non-NULL on_data_consumed_cb: ]*/
-                if (
-                    DList_IsListEmpty(&channel_ptr->op_list) ||
-                    CONTAINING_RECORD(channel_ptr->op_list.Flink, CHANNEL_OP, anchor)->on_data_consumed_cb != NULL
-                    )
+                if (interlocked_add(&channel_ptr->count_of_operations_pulled, 0) == 0 &&
+                    interlocked_add(&channel_ptr->overflow_level, 0) != 0 &&
```
parth21999, line 538: >interlocked_add(&channel_ptr->overflow_level, 0) [](http://example.com/codeflow?start=20&length=48)

I think it's better to use `UINT32_MAX` to mean no limit. 

Then this if condition can be simplified.

PR #522, opened 2025-08-26
[DO Not Merge POC changes] POC-saving-mem-copy-from-safearray-into-constbuffer

shrjadhav-msft:


src/constbuffer.c
```diff
@@ -183,6 +183,42 @@ IMPLEMENT_MOCKABLE_FUNCTION(, CONSTBUFFER_HANDLE, CONSTBUFFER_CreateWithMoveMemo
     return (CONSTBUFFER_HANDLE)result;
 }
 
+IMPLEMENT_MOCKABLE_FUNCTION(, CONSTBUFFER_HANDLE, CONSTBUFFER_CreateWithMoveMemoryWithCustomFree, unsigned char*, source, uint32_t, size, CONSTBUFFER_CUSTOM_FREE_FUNC, customFreeFunc, void*, customFreeFuncContext)
```
BertKleewein, line 186: >CONSTBUFFER_CreateWithMoveMemoryWithCustomFree [](http://example.com/codeflow?start=50&length=46)

This function looks identical to CONSTBUFFER_CreateWithCustomFree, except for a missing `const`  and a removed NULL check.  I'm not sure how this is different or why we need this.

inc/c_util/constbuffer.h
```diff
@@ -65,6 +65,8 @@ MOCKABLE_INTERFACE(constbuffer,
 
     FUNCTION(, CONSTBUFFER_HANDLE, CONSTBUFFER_CreateWithMoveMemory, unsigned char*, source, uint32_t, size),
 
+    FUNCTION(, CONSTBUFFER_HANDLE, CONSTBUFFER_CreateWithMoveMemoryWithCustomFree, unsigned char*, source, uint32_t, size, CONSTBUFFER_CUSTOM_FREE_FUNC, customFreeFunc, void*, customFreeFuncContext),
```
anporumb, line 68: >unsigned [](http://example.com/codeflow?start=83&length=8)

why is this parameter not const?

anporumb, line 68: >CONSTBUFFER_CUSTOM_FREE_FUNC [](http://example.com/codeflow?start=123&length=28)

I don't think this does what the name of the API implies it does.

The API should get another pointer function that does "move". (CreateWith*Move*).

Otherwise, how does it know how to "move"? It needs to tell to the _current_ owner something along the lines of "I am stealing this memory from under you, and I promise I will call your own free at a later time".

The "steal" part is missing.

anporumb: Normally I would be waiting, but this is not a real code review, just something to get a preview of what the collective is thinking.

PR #531, opened 2025-09-20
Use the ENABLE_TEST_FILES_PRECOMPILED_HEADERS option for PCH

dcristoloveanu:


tests/async_retry_wrapper_ut/CMakeLists.txt
```diff
@@ -10,29 +10,21 @@ set(${theseTestsName}_c_files
     ../../src/async_retry_wrapper.c
     test_async.c
     test_async_retry_wrappers.c
+    real_test_ref_counted.c
```
jebrando, line 13: >real_test_ref_counted.c [](http://example.com/codeflow?start=4&length=23)

I'm guessing this should be in the reals folder somewhere and not in here.  There's a couple of places that it's used so it's better to have a single source of truth.

PR #578, opened 2026-02-21
update dependencies and fix srs validation

parth21999:
Propagating dependency updates

### DIFF
--- a/tests/async_retry_wrapper_ut/async_retry_wrapper_ut.c
+++ b/tests/async_retry_wrapper_ut/async_retry_wrapper_ut.c
@@ -1,6 +1,14 @@
 // Copyright (c) Microsoft. All rights reserved.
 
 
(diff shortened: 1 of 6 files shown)


tests/async_retry_wrapper_ut/async_retry_wrapper_ut.c
```diff
@@ -1,6 +1,14 @@
 // Copyright (c) Microsoft. All rights reserved.
 
 
+// Tags below document macro expansions in test_async_retry_wrappers.c
```
mattdurak: I kind of liked the old version, but I get that this doesn't pass the new repo validation.

I think I'd like to see the validation script updated to allow the other one but it's kind of a "one-off" and I don't feel strongly.

Not a blocker, signing off

mattdurak (dismissed):
:shipit:

PR #621, opened 2026-04-15
Convert all tests to use TIMED_TEST_SUITE_INITIALIZE/CLEANUP

parth21999:
Converts all test files from TEST_SUITE_INITIALIZE/CLEANUP to TIMED_TEST_SUITE_INITIALIZE/CLEANUP.

Changes:
- Update c-pal submodule to latest master (with timed test suite support)
- 65 test .c files: macro replacement
- 49 PCH files: added c_pal/timed_test_suite.h before ENABLE_MOCKS
- 17 int test .c files: added c_pal/timed_test_suite.h
- UT CMakeLists: added real_process_watchdog + c_pal
- Int CMakeLists: added c_pal where missing

[Task 37635814](https://msazure.visualstudio.com/One/_workitems/edit/37635814): convert c-util to use TIMED_TEST_SUITE


build_functions/c_util_suppressions.sup
```diff
@@ -6,6 +6,19 @@
...
+   match-leak-kinds: possible
+   fun:calloc
+   ...
+   fun:allocate_dtv
+   fun:_dl_allocate_tls
+   fun:allocate_stack
+   fun:pthread_create@@GLIBC_*
+   fun:timer_helper_thread
+   fun:start_thread
+}
```
anporumb, line 21: why is this needed? #Resolved

tests/async_op_ut/CMakeLists.txt
```diff
@@ -15,6 +15,6 @@ set(${theseTestsName}_h_files
 )
 
 build_test_artifacts(${theseTestsName} "tests/c_util" 
-    ADDITIONAL_LIBS c_pal c_pal_reals
+    ADDITIONAL_LIBS real_process_watchdog c_pal c_pal c_pal_reals
```
anporumb: >c_pal [](http://example.com/codeflow?start=48&length=5)

why do we need c_pal in a unittest? #Pending

tests/async_retry_wrapper_ut/CMakeLists.txt
```diff
@@ -25,6 +25,6 @@ set(${theseTestsName}_h_files
 include_directories(../../inc)
 
 build_test_artifacts(${theseTestsName} "tests/c_util" 
-    ADDITIONAL_LIBS c_util c_pal_reals
+    ADDITIONAL_LIBS real_process_watchdog c_pal c_util c_pal_reals
```
anporumb: >c_pal c_util [](http://example.com/codeflow?start=42&length=12)

so here's a comment for all the libs that get linked in the UnitTests: why do we need the prod "c_pal" and "c_util"? What functionality is needed? What if you need a mock, how would the UT discriminate between the mock and the lib code? #Pending

tests/thread_notifications_dispatcher_int/thread_notifications_dispatcher_int.c
```diff
@@ -16,19 +16,20 @@
 #include "c_pal/interlocked.h"
 #include "c_pal/interlocked_hl.h"
 #include "c_pal/sync.h"
+#include "c_pal/timed_test_suite.h"
```
anporumb, line 19: >timed_test_suite.h [](http://example.com/codeflow?start=16&length=18)

actually let me ask this question: why is something "test related" coming from c_pal?

The expectation would be that the header is coming from ctest/ctestrunner... whatever, just not from c_pal. What does c_pal have to do with anything test-y #Pending

anporumb (changes requested):
:clock1:

tests/thread_notifications_dispatcher_int/thread_notifications_dispatcher_int.c
parth21999: c-pal provides the process_watchdog that runs a timer to collect a dump if the test hangs.

build_functions/c_util_suppressions.sup
parth21999: The timer code on linux creates a false positive on helgrind. This is documented in c-pal.

TIMED_TEST_SUITE brings in the timer in c_util, so c_util also needs the suppression now.

tests/thread_notifications_dispatcher_int/thread_notifications_dispatcher_int.c
parth21999: There was a fair bit of discussion regarding where process_watchdog should be placed. It was finally decided that c-pal is the correct repo for it.

Since the TIMED_TEST_SUITE macros need the process_watchdog, they must also sit in c-pal.

tests/async_retry_wrapper_ut/CMakeLists.txt
parth21999: Removed c_pal from here. c_util is pre-existing.

tests/async_op_ut/CMakeLists.txt
parth21999: Proabably because the UT did not rename interlocked_ symbols to real_ versions.

It is a pre-existing issue. I tried removing it and compiler complains. 

Added: [https://msazure.visualstudio.com/DefaultCollection/One/_workitems/edit/38360022](https://msazure.visualstudio.com/DefaultCollection/One/_workitems/edit/38360022)

I also created a separate real_process_watchdog lib for including the process watchdog, so the entire c-pal doesn't need to be linked because TIMED_TEST_SUITE needs it.

PR #629, opened 2026-05-07
Fix channel_int_valgrind 'possibly lost' leaks: cleanup before user callback

parth21999:
## Summary

Fix the `channel_int_valgrind` test failure (build [161912975](https://msazure.visualstudio.com/One/_build/results?buildId=161912975)) by reordering `execute_callbacks` so cleanup happens BEFORE the user callbacks fire.

## The bug

Today's `execute_callbacks`:
1. `sm_exec_end` (early — supports re-entrant `channel_close` from callback)
2. **User callback fires** — caller wakes from `InterlockedHL_WaitFor*`
3. **Cleanup tail** — releases `channel_op->channel`, `channel_op->async_op` self-ref, etc.

When the caller wakes at step 2, it proceeds with its own cleanup (releases its `async_op` and `channel` references). Meanwhile the worker is still at step 3, **holding `channel_op->channel`** which transitively holds a `THREADPOOL` ref. Under valgrind's 10-30× slowdown this race window is wide enough that the worker is reliably still in the cleanup tail when `main()` returns:

- threadpool refcount > 0 at exit
- `threadpool_dispose` never runs
- worker threads are never joined
- glibc never reaps pthread TLS
- valgrind reports 12 "possibly lost" allocations (~33 KB): the threadpool struct, TQUEUE buffer, pthread TLS, channel struct, srw_lock, sm, async_op, work item, RC_STRINGs

## The fix

Release `channel_op->channel` and the `channel_op->async_op` self-reference **before** invoking the user callbacks. To keep `channel_op` memory valid for the user callbacks (which still reference `channel_op->pull_correlation_id`, `channel_op->push_correlation_id`, and `channel_op->data` directly), take a single local strong reference on `async_op` first. Move the cleanup of correlation_ids and data to `dispose_channel_op` so it runs whenever the final `async_op` reference is released.

After the fix, when the caller wakes at the user-callback signal point, the worker has already released its chain of refs to channel/threadpool. The caller's cleanup brings `channel` refcount cleanly to 0 → `channel_dispose` releases threadpool ref → at suite cleanup the threadpool ref drops to 0 → `threadpool_dispose` joins workers → TLS reaped. No leaks.

## Why not the alternative (channel-owns-threadpool with stronger close contract)

Considered restructuring `channel` to own its threadpool internally and have `channel_close` drain via `threadpool_dispose`. That's a cleaner architecture but requires:
- API break (`channel_create` takes `EXECUTION_ENGINE_HANDLE`)
- Forbidding `channel_close` from inside a callback
- Cascading update to `bounded_channel` in `Azure-Messaging-GeoReplication` and to a known production callsite (`on_geo_core_secondary_receiver_get_next_complete` EndOfStream branch) that currently calls close from inside a callback

This in-place reorder fix avoids all of that — same valgrind outcome, no API break, no consumer changes needed. The contract is preserved: callbacks are still allowed to call `channel_close` re-entrantly (`sm_exec_end` stays early).

## What changed

- `src/channel.c`:
  - `execute_callbacks`: cleanup of `channel_op->channel` and `channel_op->async_op` self-ref now happens BEFORE the user callbacks. A local `THANDLE(ASYNC_OP)` ref keeps `channel_op` memory valid for the duration of the user callbacks.
  - `dispose_channel_op`: now releases `channel_op->pull_correlation_id`, `channel_op->push_correlation_id`, and `channel_op->data` (it used to be a no-op since cleanup happened in `execute_callbacks`).
- `devdoc/channel_requirements.md`: SRS updated to reflect new ordering. New requirements `SRS_CHANNEL_88_001/002/003` added (spec author 88).
- `tests/channel_ut/channel_ut.c`: `setup_op_cleanup_expectations` updated to reflect the new mock-call order. The RC_STRING/RC_PTR ASSIGNs that used to happen inside `execute_callbacks` now happen during `dispose_channel_op` (after the test's assertion window) and so are intentionally not in the expectations anymore — verified all 20 UTs pass.

## Validation

| Test | Result |
|------|--------|
| `channel_ut` (Windows) | 20/20 passing |
| `channel_int` (Windows) | 16/16 passing |
| `channel_int` (Linux) | 15/15 passing |
| **`channel_int_valgrind` (WSL)** | **passing in 1908s, 0 leaks** ✅ |
| `channel_int_helgrind` (WSL) | passing in 873s, 0 races |

The valgrind run is longer than the original buggy run (1908s vs 874s) because the fix lets `threadpool_dispose` run and join workers at exit; the original run skipped that cleanup. CI may need a timeout adjustment if the current `channel_int_valgrind` timeout is < ~32 minutes.

### DIFF
--- a/devdoc/channel_requirements.md
+++ b/devdoc/channel_requirements.md
@@ -307,6 +307,25 @@ channel_open` opens the given `channel`.
 
 **SRS_CHANNEL_43_157: [** `execute_callbacks` shall call `sm_exec_end` for each callback that is called. **]**
 
+`sm_exec_end` is called BEFORE the user callbacks so that a user callback may call `channel_close` re-entrantly without deadlocking on its own outstanding `sm_exec_begin` reference. This is covered by the existing `channel_close_does_not_deadlock_if_called_from_pull_callback` and `channel_close_does_not_deadlock_if_called_from_push_callback` integration tests.
+
+**SRS_CHANNEL_88_001: [** `execute_callbacks` shall take a local `THANDLE` reference on `channel_op->async_op` so `channel_op` memory remains valid for the user callbacks after the self-reference is released. **]**
+
+**SRS_CHANNEL_43_147: [** `execute_callbacks` shall release `channel_op->channel` and the `channel_op->async_op` self-reference before invoking the user callbacks. **]**
+
+The `channel_op->channel` reference and the `channel_op->async_op` self-reference are released BEFORE the user callbacks are invoked. The user callbacks signal completion to the caller; once signaled, the caller may proceed with releasing its references to `async_op` and `channel`. Performing this cleanup AFTER the user callbacks would leave the worker thread holding `channel_op->channel` (and transitively a `THREADPOOL` reference) past the point where the caller releases its references. At process exit the threadpool refcount could still be > 0, preventing `threadpool_dispose` from joining workers and freeing pthread TLS - reported by valgrind as "possibly lost" allocations totaling ~33 KB (build 161912975 was the original trigger).
+
+The local `THANDLE` reference taken on `async_op` (per `SRS_CHANNEL_88_001`) keeps `channel_op` memory alive across the user callbacks so they can still reference `channel_op->pull_correlation_id`, `channel_op->push_correlation_id`, and `channel_op->data` directly. These fields are released by `dispose_channel_op` when the final `THANDLE(ASYNC_OP)` reference is released and the `async_op` is disposed.
+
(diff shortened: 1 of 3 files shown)


parth21999:
Closing in favor of a much simpler fix on branch `paag/fix-channel-int`. The leak is broken just by releasing `channel_op->channel` before the user callback (one moved line) - no need to move the rest of the cleanup, no new SRS requirements, no test/traceability changes needed.

Verified the simpler fix passes `channel_int_valgrind` in 1842s with 0 leaks.

New PR coming.

PR #631, opened 2026-05-07
Fix channel_int_valgrind 'possibly lost' leaks: release channel ref before user callbacks

parth21999:
## Summary

Fix the `channel_int_valgrind` test failure (build [161912975](https://msazure.visualstudio.com/One/_build/results?buildId=161912975)) by releasing `channel_op->channel` BEFORE invoking the user callbacks.

## The bug

Today's `execute_callbacks`:
1. `sm_exec_end` (early — supports re-entrant `channel_close` from callback)
2. **User callback fires** — caller wakes from `InterlockedHL_WaitFor*`
3. **Cleanup tail** — releases `channel_op->channel`, `channel_op->async_op` self-ref, etc.

When the caller wakes at step 2, it proceeds with its own cleanup (releases its `async_op` and `channel` references). Meanwhile the worker is still at step 3, **holding `channel_op->channel`** which transitively holds a `THREADPOOL` ref. Under valgrind's 10-30× slowdown the race window is wide enough that the worker is reliably still in the cleanup tail when `main()` returns:

- threadpool refcount > 0 at exit
- `threadpool_dispose` never runs
- worker threads are never joined
- glibc never reaps pthread TLS
- valgrind reports 12 "possibly lost" allocations (~33 KB)

## The fix

Move the single `THANDLE_ASSIGN(CHANNEL)(&channel_op->channel, NULL)` call to BEFORE the user callbacks fire. By the time a user callback signals the test, the worker no longer holds any chain reaching the threadpool. Test cleanup proceeds, channel refcount drops cleanly to 0, threadpool ref released, suite_cleanup releases its threadpool ref → `threadpool_dispose` joins workers → TLS reaped.

The other `channel_op` fields (`async_op` self-ref, correlation_ids, data) don't transitively hold threadpool references so they can stay where they are. Re-entrant `channel_close` from a user callback continues to work because `sm_exec_end` is still called early.

## What changed

`src/channel.c`: the existing `THANDLE_ASSIGN(CHANNEL)(&channel_op->channel, NULL)` was moved from after the user callbacks to right after the `sm_exec_end` calls (before the user callbacks). Net diff: +3, -1 lines in one file.

No SRS / devdoc changes (the existing `SRS_CHANNEL_43_147` already covers cleanup, and the order is implementation detail).

No test changes (`THANDLE_ASSIGN(CHANNEL)` isn't mocked since CHANNEL is the type under test, so reordering its release doesn't affect any `STRICT_EXPECTED_CALL` sequence).

## Validation

| Test | Result |
|------|--------|
| `channel_ut` (Windows) | 20/20 passing |
| `channel_int` (Windows) | 16/16 passing |
| `channel_int` (Linux) | 15/15 passing |
| **`channel_int_valgrind` (WSL)** | **passing in 1842s, 0 leaks** ✅ |

The valgrind run is longer than the original buggy run (1842s vs 874s) because the fix lets `threadpool_dispose` run and join workers at exit, whereas the original run exited prematurely with refs still held. **CI may need a timeout adjustment if the current `channel_int_valgrind` timeout is < ~32 minutes.**

### DIFF
--- a/src/channel.c
+++ b/src/channel.c
@@ -265,14 +265,6 @@ static void execute_callbacks(void* context)
             sm_exec_end(channel_op->channel->sm);
         }
 
-        // Release the reference to channel BEFORE invoking the user callbacks. The user
-        // callbacks signal completion to the caller; once the caller wakes it may proceed
-        // to release its own reference to channel. If channel_op->channel were released
-        // AFTER the user callbacks, this thread would still be holding a channel reference
-        // (and transitively a THREADPOOL reference) past the point where the caller drops
-        // its own references. At process exit the threadpool refcount could still be > 0,
-        // preventing threadpool_dispose from joining workers and freeing pthread TLS;
-        // valgrind reports this as "possibly lost" allocations totaling ~33 KB.
         THANDLE_ASSIGN(CHANNEL)(&channel_op->channel, NULL);
 
         if (channel_op->on_data_available_cb != NULL)
--- a/src/channel.c
+++ b/src/channel.c
@@ -254,6 +254,7 @@ static void execute_callbacks(void* context)
         // We have to call sm_exec_end for both of them before we call our first callback.
         // Otherwise, if one of the callbacks calls channel_close, it might deadlock waiting for an sm_exec_end.
 
+
         if (channel_op->on_data_available_cb != NULL)
         {
             /*Codes_SRS_CHANNEL_43_157: [ execute_callbacks shall call sm_exec_end for each callback that is called. ]*/
@@ -264,6 +265,8 @@ static void execute_callbacks(void* context)
             /*Codes_SRS_CHANNEL_43_157: [ execute_callbacks shall call sm_exec_end for each callback that is called. ]*/
             sm_exec_end(channel_op->channel->sm);
         }
+
+        THANDLE_ASSIGN(CHANNEL)(&channel_op->channel, NULL);
 
         if (channel_op->on_data_available_cb != NULL)
         {
@@ -280,7 +283,6 @@ static void execute_callbacks(void* context)
         THANDLE_ASSIGN(RC_STRING)(&channel_op->pull_correlation_id, NULL);
         THANDLE_ASSIGN(RC_STRING)(&channel_op->push_correlation_id, NULL);
         THANDLE_ASSIGN(RC_PTR)(&channel_op->data, NULL);
-        THANDLE_ASSIGN(CHANNEL)(&channel_op->channel, NULL);
 
         // copy to local reference to avoid cutting the branch you are sitting on
         THANDLE(ASYNC_OP) temp = NULL;
--- a/src/channel.c
+++ b/src/channel.c
@@ -254,7 +254,6 @@ static void execute_callbacks(void* context)
         // We have to call sm_exec_end for both of them before we call our first callback.
         // Otherwise, if one of the callbacks calls channel_close, it might deadlock waiting for an sm_exec_end.
 
-
         if (channel_op->on_data_available_cb != NULL)
         {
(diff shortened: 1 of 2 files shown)


src/channel.c
```diff
@@ -254,6 +254,7 @@ static void execute_callbacks(void* context)
         // We have to call sm_exec_end for both of them before we call our first callback.
         // Otherwise, if one of the callbacks calls channel_close, it might deadlock waiting for an sm_exec_end.
 
+
         if (channel_op->on_data_available_cb != NULL)
```
parth21999, line 257: remove extra newline #Resolved

parth21999 (commented):
Done in e741feb. Also added a why-comment explaining the rationale for the early channel-ref release, and a new regression test `test_pull_and_cancel_with_sleep_in_callback` that uses a callback variant which sleeps 100ms after wake_by_address_single.

Validated locally:
- pre-fix code + new test under valgrind: 12 'possibly lost' blocks, 33,730 B (same signature as build 161912975)
- post-fix code + new test under valgrind: 0 leaks

tests/channel_int/channel_int.c
```diff
@@ -69,6 +69,32 @@ static int32_t push_abandoned = 0x0007;
...
+    // own cleanup (release async_op, channel_close, release channel) before
+    // this callback returns and execute_callbacks proceeds to its cleanup tail.
+    // This widens the window where the worker would otherwise hold
+    // channel_op->channel (and transitively a THREADPOOL reference) past the
+    // point where the caller has dropped its references. Without the fix
+    // (release channel BEFORE the user callback), this guarantees that at
+    // process exit threadpool refcount > 0, threadpool_dispose never runs,
+    // and valgrind reports ~33 KB of "possibly lost" allocations. With the
+    // fix in place this sleep is harmless; the channel ref was already
+    // dropped before this callback ran.
```
parth21999: make this a 2 line comment. make it SHORT #Resolved

src/channel.c
```diff
@@ -265,6 +265,16 @@ static void execute_callbacks(void* context)
...
         }
 
+        // Release the reference to channel BEFORE invoking the user callbacks. The user
+        // callbacks signal completion to the caller; once the caller wakes it may proceed
+        // to release its own reference to channel. If channel_op->channel were released
+        // AFTER the user callbacks, this thread would still be holding a channel reference
+        // (and transitively a THREADPOOL reference) past the point where the caller drops
+        // its own references. At process exit the threadpool refcount could still be > 0,
+        // preventing threadpool_dispose from joining workers and freeing pthread TLS;
+        // valgrind reports this as "possibly lost" allocations totaling ~33 KB.
```
parth21999: REMOVE #Resolved

parth21999: Done in 92ecc23.

tests/channel_int/channel_int.c
parth21999: Shortened to 2 lines in 92ecc23.

src/channel.c
```diff
@@ -265,6 +265,8 @@ static void execute_callbacks(void* context)
             sm_exec_end(channel_op->channel->sm);
         }
 
+        THANDLE_ASSIGN(CHANNEL)(&channel_op->channel, NULL);
```
jebrando, line 271: >THANDLE_ASSIGN(CHANNEL)(&channel_op->channel, NULL); [](http://example.com/codeflow?start=8&length=52)

This could use a comment as to why it was moved.

jebrando (dismissed):
:shipit:

src/channel.c
parth21999: Done.

dcristoloveanu (approved):
:shipit:

